PR #657, opened 2026-04-28
Bloom-filter the persisted chunk index for write-side dedup

timsehn:
## Summary

Wraps the two write-hot negative-lookup sites (`chunkStoreHas` and the dedup check in `chunkStorePut`) with a Bloom filter over the persisted index. For any workload emitting genuinely new content, the dedup `csSearchIndex` previously paid a full ~25-cache-miss bsearch per emitted chunk to learn what it didn't already know.

Goal: push sysbench writes under 1× SQLite. Stretch goal that follows the mmap PR's surprise win.

## Why writes specifically

`chunkStorePut` runs:

```c
prollyHashCompute(pData, nData, &h);
if( csSearchIndex(cs->aIndex, cs->nIndex, &h) >= 0 ) return SQLITE_OK; // ← here
if( csSearchPending(cs, &h) >= 0 ) return SQLITE_OK;
```

For a sysbench workload writing rows with random keys/values, the persisted-index lookup returns -1 every time. Bloom filter answers "definitely not present" in one cache line, the bsearch never runs.

`chunkStoreGet` is intentionally NOT wrapped — its lookups are mostly positive (we have a hash because a tree pointer told us it existed), so bloom would just add work the bsearch was going to have to do anyway.

## Implementation

- `struct ProllyBloom` in `chunk_store.h`, lives on `ChunkStore`.
- 10 bits per entry, `k=7` → ~0.82% false-positive rate. 200K chunks = 250 KB filter, fits in L2.
- Built lazily on first lookup once `cs->nIndex >= 256` (`BLOOM_MIN_ENTRIES`); below that the bsearch is already cheap.
- Position derivation is Kirsch–Mitzenmacher double-hashing over two 32-bit halves of the chunk hash. Chunk hashes are already uniform digests, so reusing their bits is fine.
- Invalidated wherever `cs->aIndex` is replaced: `csCommitToMemory`, the in-WAL-replay merge, the post-commit merge, `csRestoreReplayState`, `csAdoptOpenedStoreState`. Next lookup rebuilds.
- Freed in `chunkStoreClose`.

The whole filter state is rebuildable from `cs->aIndex`, so opening / adopting / restoring all stay correct without persisting bloom bits. No on-disk format change, no public API change.

## Verification

- 30K+10K branch+merge round trip: 40K rows, no corruption
- `bash test/chunk_distribution_test.sh ./doltlite` — 7/7 (chunking is orthogonal; same metrics as master)
- `bash test/vc_oracle_merge_test.sh ./doltlite dolt` — 41/41
- `bash test/vc_oracle_branch_test.sh ./doltlite dolt` — 30/30

## Test plan

- [ ] CI passes (build + asan + oracle suite + crash recovery + WASM)
- [ ] Sysbench write multipliers: target < 1× on at least one of `oltp_insert`, `oltp_write_only`, `oltp_bulk_insert`

### DIFF
--- a/src/chunk_store.c
+++ b/src/chunk_store.c
@@ -422,6 +422,50 @@ static int csBloomMaybeContains(const ProllyBloom *b, const ProllyHash *ph){
   return 1;
 }
 
+/* Insert one entry into an existing bloom. Bloom filters support
+** insertion (set bits, never clear) — only deletion is unsupported.
+** Used by csCommitToMemory and the merge paths to extend the filter
+** with newly-committed pending hashes instead of invalidating and
+** paying a full O(N) rebuild on every commit. */
+static void csBloomAdd(ProllyBloom *b, const ProllyHash *ph){
+  const u8 *p;
+  u32 h1, h2;
+  int k;
+
+  if( !b->bits ) return;       /* no filter; nothing to extend */
+
+  p = ph->data;
+  h1 = (u32)p[0] | ((u32)p[1] << 8)
+     | ((u32)p[2] << 16) | ((u32)p[3] << 24);
+  h2 = (u32)p[4] | ((u32)p[5] << 8)
+     | ((u32)p[6] << 16) | ((u32)p[7] << 24);
+
+  for( k = 0; k < b->nHashes; k++ ){
+    u64 pos = ((u64)h1 + (u64)k * (u64)h2) % (u64)b->nBits;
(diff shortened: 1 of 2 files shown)


timsehn:
Closing — bloom turned out to be the wrong optimization in this regime. The mmap+packed-struct PR (#654) made the bsearch fast enough and cache-warm enough that the bloom's 7-cache-line check + L2 contention cost more than it saved. Dropping rather than letting it sit.

PR #665, opened 2026-04-28
Batch WAL chunk writes: 2N+1 sqlite3OsWrite calls per commit → ~1 per 64 KB

timsehn:
## Summary

`csCommitToFile` previously did **2N+1** `sqlite3OsWrite` calls per commit (header + data per pending chunk, plus the root record). Each `xWrite` is a kernel boundary crossing through SQLite's VFS — on workloads emitting hundreds-to-thousands of small chunks per commit, that's where syscall overhead lives.

This PR coalesces them. Pending chunk records (and the root) are assembled into a single buffer; one `sqlite3OsWrite` ships the buffer. Same bytes on disk in the same order. Same `sqlite3OsSync` at the end. Durability contract is byte-for-byte unchanged.

## Stays inside SQLite's machinery

- Same `sqlite3OsWrite` API
- Same `sqlite3OsSync` at the end of the commit
- No `fd` manipulation, no mmap-based writes, no custom fsync
- No reordering of bytes on disk

The only difference is that the SQLite VFS sees one large `xWrite` instead of many small ones.

## Buffer cap

64 KB — same threshold the original per-chunk-data path used. `sqlite3OsWrite` doesn't loop on partial writes (returns `SQLITE_FULL`), and 65536 is the safe ceiling some filesystems will pwrite atomically. Multiple flushes per commit are fine; this just coalesces small chunk records into each flush.

For a typical workload (chunks averaging 200-500 bytes), one flush packs 100-300 chunk records that previously took 200-600 `xWrite` calls.

## Big-chunk fallback

Single chunk records exceeding 64 KB (large BLOBs) fall back to the original direct-write path: header in one xWrite, data in 64 KB chunks. Same code as before, just routed through when batching can't help.

## Crash semantics

`CRASH_CHECK_WRITE()` fires once per flush instead of once per individual write. The durability property "crash at any flush yields a recoverable state" is unchanged — only the granularity of injectable crash points differs.

## Numbers (local, 10 runs)

50K prep + 200K bench-insert into a table with one secondary index:

| | min | median | mean |
|---|---:|---:|---:|
| master | 1.160 s | 1.180 s | 1.194 s |
| branch | 1.120 s | **1.130 s** | **1.134 s** |
| Δ | −3.4% | **−4.2%** | **−5.0%** |

CPU profile post-change: syscall6 / `sqlite3OsWrite` samples in a 2-second window drop from non-trivial to **zero**. The kernel-boundary hotspot is gone.

## Verification

- `vc_oracle_merge_test`: 41/41
- `vc_oracle_branch_test`: 30/30
- `vc_oracle_diff_test`: 41/41
- `chunk_distribution_test`: 7/7
- `correctness_test`: 41/41
- `large_scale_test --quick`: 19/19
- 30K + 5K branch + merge end-to-end OK

No on-disk format change. No public API change. Crash recovery test in CI.

## Test plan

- [ ] CI passes (build + asan + oracle suite + crash recovery + WASM)
- [ ] Sysbench multipliers improve (this targets the kernel-boundary cost which is more sysbench-relevant than the qsort hotspot was)

### DIFF
--- a/src/chunk_store.c
+++ b/src/chunk_store.c
@@ -2079,6 +2079,7 @@ static int csCommitToFile(ChunkStore *cs){
   int hadFile = (cs->pFile != 0);
   i64 newWalSize = cs->nWalData;
   u8 *pNewWalData = 0;
+  u8 *batchBuf = 0;
   ChunkIndexEntry *aCommittedPending = 0;
   ChunkIndexEntry *aMerged = 0;
   int nMerged = 0;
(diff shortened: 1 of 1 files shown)


timsehn:
Closing — local A/B at sysbench shape (5K inserts, ~50ms test) showed no move. The win materialized at the 200K workload (4-5% median) but at sysbench's commit-time scale it's invisible. Per-insert cost dominates at that shape, not commit-time syscalls.

PR #708, opened 2026-04-30
Enable foreign keys by default

timsehn:
## Summary
- enable foreign key checks by default for DoltLite shell connections
- add a DoltLite regression covering default FK enforcement for both PK-parent and UNIQUE-parent refs
- update the FK error recovery oracle to avoid Dolt's inline REFERENCES bug by using explicit FOREIGN KEY syntax

## Testing
- make -j4 doltlite
- bash test/doltlite_foreign_keys_default.sh ./doltlite
- bash test/vc_oracle_error_recovery_test.sh ./doltlite dolt

### DIFF
--- a/src/shell.c.in
+++ b/src/shell.c.in
@@ -4930,6 +4930,13 @@ static void open_db(ShellState *p, int openFlags){
     sqlite3_free(zParsedFilename);
     sqlite3_free(zParsedBranch);
     globalDb = p->db;
+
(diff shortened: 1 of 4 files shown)


timsehn:
Closing this one. We decided DoltLite should keep SQLite's default foreign_keys=OFF behavior, so the default-on change is being unwound instead of merged.

PR #729, opened 2026-05-03
prolly_mutate: drop strategy heuristic + byte-uniform INT/BLOB cleanup

LaPingvino:
## TL;DR

Two layered changes:

1. **`a19009e0d`** drops the per-shape strategy heuristic in `prollyMutateFlush` and routes all flushes through `streamingMerge`. **This is where the perf win lives.** On real bahaiwritings history: 18× faster on the worst commit (75.14s → 4.26s on a 36K-row bulk INSERT into a populated TEXT-PK table), -22% aggregate chunk-apply time across the full 780-commit replay.

2. **The remaining 6 commits** unify the residual INT-vs-BLOB code-path divergence in mutmap, flush, diff, three-way diff, and cursor seek into a single byte-uniform path. Real structural cleanup (~250 lines net, 173 → 84 lines containing `isIntKey`/`PROLLY_NODE_INTKEY`/`intKey`) but contributes only ~5% incremental on aggregate chunk-apply and ~1% on worst case — basically noise on perf, real value as cleanup.

The two are bundled because the byte-uniform path was originally motivated as the structural foundation Tim asked for ("there is no reason INT keys should be treated differently than other keys") and the strategy-drop wanted that foundation to land alongside it. They can be split into separate PRs if preferred.

## Headline measurement: full bahaiwritings replay (780 commits, real production history, three-way bench)

| Metric | master | a19009e0d only | this branch | strategy-drop alone | unification incremental |
|---|---:|---:|---:|---:|---:|
| Chunk apply max | 75.14s | 4.26s | 4.13s | **17.6× faster** | 1.03× faster |
| Chunk apply total | 326.1s | 252.9s | 239.3s | **-22%** | -5% |
| Chunk apply mean | 0.99s | 0.77s | 0.73s | -22% | -5% |
| dolt_commit total | 427.8s | (similar) | 423.7s | within noise | within noise |
| Wall time | 25.5min | 25.1min | 24.3min | -1.7% | -3.1% |
| Final DB size | 1.366GB | 1.366GB | 1.366GB | identical | identical |

The 75-second outlier on master is commit `i9u99ijkv6`: *\"Add 36K writing entries from bahaiprayers.net API\"* — bulk INSERT into a populated TEXT-PK table. Master's chunker hits the old `edits > 1% of leaves → mergeWalk` heuristic and spends 75 seconds; the strategy-drop sends it through streamingMerge in 4.26.

## Why three-way bench, and what it caught

I initially attributed the 18× to the byte-uniform unification work in this PR's description. A code-review subagent proposed the falsifying experiment: build a third binary at `a19009e0d` only (no unification commits), run i9u99ijkv6 on it. If the strategy-drop alone produces 4s, the unification isn't where the perf came from. It does. So the headline attribution had to be corrected.

The unification work was correct framing of the broader code health goal (Tim's parity mission) but it doesn't produce the perf number. It produces the *structural* result: 173 → 84 lines of `isIntKey`/`PROLLY_NODE_INTKEY`/`intKey` across the prolly tree files, single byte-uniform code path through mutmap/flush/diff/cursor.

## Why synthetic micro-benches missed this

Earlier in this PR's history I A/B'd master vs branch on synthetic linear-INSERT workloads at 10K / 50K / 100K rows. Numbers came out close to noise, with parity actually 25-29% slower on the small synthetics. I (wrongly) concluded the cleanup had no perf benefit.

The cause: empty-table linear inserts hit a fast path that bypasses the mutmap-flush merge entirely. The strategy heuristic is never tripped because the tree has no leaves. Real-world commits that merge new data into populated trees DO hit it, and that's where the 17.6× shows up. The bahaiwritings replay against real history is the only measurement that exposed this.

## What this PR actually does (structural)

Drops 89 lines containing `isIntKey` / `intKey` / `PROLLY_NODE_INTKEY` (155 → 76 lines containing one or more occurrences; 173 → 84 raw occurrences). The remaining occurrences are mostly:
- Function parameter signatures still passing `isIntKey`/`intKey` (cosmetic — values flow through one code path now)
- Dead helpers (`prollyNodeSearchInt`, `prollyCompareKeys`) kept for API stability
- Type-flag constants and comments

| Layer | Before | After |
|---|---|---|
| Strategy choice in flush | Heuristic: `edits > 1% of leaves → mergeWalk; else streamingMerge` | Always streamingMerge |
| Mutmap storage | INT path stored i64 + maybe bytes; BLOB path stored bytes | One byte path; INT keys encoded sortably at API entry |
| Mutmap compare / hash | Branched on `isIntKey` | Lexicographic memcmp / FNV-1a over bytes |
| Streaming-merge flush | Decoded INT to i64 per key, re-encoded to bytes for chunker | Pass-through bytes |
| Diff walker compare | Branched per-flag | One byte path |
| Diff emit | Branched per-flag, filled either intKey or pKey | Helper fills both for INT mode (consumer compat) |
| Three-way diff compare | Branched per-flag | One byte path |
| Cursor seek (INT) | Separate descent + `prollyNodeSearchInt` at every level | 4-line wrapper around `SeekBlob` over encoded form |

Plus a tightening commit (`a08d34b91`): `prepKey`'s contract is now `assert(intKey == 0)` when caller passes pre-encoded bytes, so the convention becomes enforceable rather than convention-by-comment. Caught by the same code-review subagent pass.

## Methodology notes

- All three binaries built from this checkout (master at `dff55e83f`, intermediate at `a19009e0d`, branch at `a08d34b91`)
- Replays run on the real bahaiwritings DB (780 commits, table=writings, defaults), measured from same dolt-replay invocation
- Final databases byte-identical across all three (1,365,849,745 bytes), confirming the structural changes don't alter on-disk output
- Three-way bench was suggested by an independent code-review pass that flagged the original \"18× from unification\" attribution as not isolating the variable

## Commits

```
a08d34b91 prolly_mutmap, prolly_mutate: tighten prepKey contract, drop dead var
c28d0bc56 prolly_cursor: prollyCursorSeekInt becomes wrapper around SeekBlob
0470cd412 prolly_diff: byte-uniform emit, three_way_diff: byte-uniform compare
cd42cad86 prolly_diff: byte-uniform key comparison
76ff4f3c3 prolly_mutate: byte-uniform flush path, drop INTKEY branches
066c7001d prolly_mutmap: drop intKey field from ProllyMutMapEntry
3ce9caa91 prolly_mutmap: byte-uniform internal storage and comparison
a19009e0d prolly_mutate: drop strategy heuristic, always streamingMerge   ← perf source
```

## Test plan

- [x] Build clean (no new warnings)
- [x] Negative-key correctness (5K rows + negative INT keys): inserts, point lookups, range scans, IN with negatives, UPDATEs and DELETEs across negative ranges, commit round-trip
- [x] Edge i64 values (INT64_MIN, INT64_MAX, 0, -1, 1): sortable BE encoding sorts correctly across full range
- [x] dolt_diff_t / dolt_diff_s: 2-commit diff with negative INT keys produces correct add/modify/remove
- [x] dolt_merge: 3-way merge across feature/main divergence reconciles correctly
- [x] Cursor seek (5K rows + negatives): point lookups for positive and negative IDs, range scans
- [x] **Full bahaiwritings replay, three-way (master / a19009e0d-only / branch): properly attributes the 17.6× win to the strategy-drop commit; final DBs byte-identical across all three**

## What's NOT done

- API parameter signatures (`isIntKey` / `intKey` args) still exist for source compat — could be cleaned up in a follow-up; doesn't affect behavior
- Dead helpers (`prollyNodeSearchInt`, `prollyCompareKeys`) still exported; can be removed once you're confident no external consumers depend on them
- No sysbench rerun yet — sysbench's int-pk-only-update workload may not exercise the merge path; the bahaiwritings replay is the load-bearing measurement

## Recommendation

Take the strategy-drop commit unconditionally — the perf number on real workloads is real and large. The unification work is independent value (structural cleanup that makes future prolly tree work safer to reason about) and can be evaluated on its own merits; it doesn't move perf measurably but does shrink the surface area substantially.

### DIFF
--- a/src/prolly_btree.c
+++ b/src/prolly_btree.c
@@ -1241,7 +1241,7 @@ static void setCursorToMutMapEntryPhys(BtCursor *pCur, int physIdx){
   pCur->eState = CURSOR_VALID;
   pCur->curFlags &= ~BTCF_AtLast;
   if( pCur->curIntKey ){
-    pCur->cachedIntKey = pEntry->intKey;
+    pCur->cachedIntKey = prollyMutMapEntryIntKey(pEntry);
     pCur->curFlags |= BTCF_ValidNKey;
   }else{
     pCur->curFlags &= ~BTCF_ValidNKey;
@@ -1389,7 +1389,7 @@ static int saveCursorPosition(BtCursor *pCur){
   if( pCur->curIntKey ){
     if( pCur->mmActive
      && (pCur->mergeSrc==MERGE_SRC_MUT || pCur->mergeSrc==MERGE_SRC_BOTH) ){
-      pCur->nKey = currentMutMapEntry(pCur)->intKey;
+      pCur->nKey = prollyMutMapEntryIntKey(currentMutMapEntry(pCur));
     }else{
       pCur->nKey = prollyCursorIntKey(&pCur->pCur);
     }
@@ -3587,8 +3587,9 @@ int sqlite3BtreeClosesWithCursor(Btree *p, BtCursor *pCur){
 static int mergeCompare(BtCursor *pCur, ProllyMutMapEntry *e){
   if( pCur->curIntKey ){
     i64 tk = prollyCursorIntKey(&pCur->pCur);
-    if( tk < e->intKey ) return -1;
-    if( tk > e->intKey ) return 1;
+    i64 ek = prollyMutMapEntryIntKey(e);
+    if( tk < ek ) return -1;
+    if( tk > ek ) return 1;
     return 0;
   }else{
     const u8 *pK; int nK;
@@ -4508,7 +4509,7 @@ static i64 prollyBtCursorIntegerKey(BtCursor *pCur){
 
   if( pCur->mmActive
    && (pCur->mergeSrc==MERGE_SRC_MUT || pCur->mergeSrc==MERGE_SRC_BOTH) ){
-    return currentMutMapEntry(pCur)->intKey;
+    return prollyMutMapEntryIntKey(currentMutMapEntry(pCur));
   }
   if( !prollyCursorIsValid(&pCur->pCur)
    && (pCur->curFlags & BTCF_ValidNKey) ){
@@ -4985,7 +4986,7 @@ static int prollyBtCursorDelete(BtCursor *pCur, u8 flags){
     if( pCur->curIntKey ){
       if( pCur->mmActive
        && (pCur->mergeSrc==MERGE_SRC_MUT || pCur->mergeSrc==MERGE_SRC_BOTH) ){
-        savedIntKey = currentMutMapEntry(pCur)->intKey;
+        savedIntKey = prollyMutMapEntryIntKey(currentMutMapEntry(pCur));
         hasSavedKey = 1;
       }else if( !prollyCursorIsValid(&pCur->pCur)
        && (pCur->curFlags & BTCF_ValidNKey) ){
--- a/src/prolly_cursor.c
+++ b/src/prolly_cursor.c
@@ -270,36 +270,13 @@ int prollyCursorPrev(ProllyCursor *cur){
 }
 
 int prollyCursorSeekInt(ProllyCursor *cur, i64 intKey, int *pRes){
-  int rc;
-  ProllyCacheEntry *pEntry = 0;
-  int leafRes;
-  int leafIdx;
-
-  rc = initCursorAtRoot(cur, &pEntry);
-  if( rc!=SQLITE_OK ) return rc;
-  if( cur->eState==PROLLY_CURSOR_EOF ){
-    cur->eState = PROLLY_CURSOR_INVALID;
-    *pRes = -1;
-    return SQLITE_OK;
-  }
-
-
-  while( pEntry->node.level>0 ){
-    int searchRes;
-    int idx = prollyNodeSearchInt(&pEntry->node, intKey, &searchRes);
-    idx = childIndexForSearchResult(idx, searchRes, pEntry->node.nItems);
-    cur->aLevel[cur->iLevel].idx = idx;
-    rc = descendToChild(cur, idx, 0, &pEntry);
-    if( rc!=SQLITE_OK ) return rc;
-  }
-
-  if( pEntry->node.nItems==0 ){
-    cur->eState = PROLLY_CURSOR_EOF;
-    *pRes = -1;
-    return SQLITE_OK;
-  }
-  leafIdx = prollyNodeSearchInt(&pEntry->node, intKey, &leafRes);
-  return finalizeSeekOnLeaf(cur, pEntry, leafIdx, leafRes, pRes);
+  /* INT keys live on disk as sortable 8-byte big-endian bytes (see
+  ** prollyNodeIntKey/prollyEncodeIntKey), so an int seek is just a
+  ** blob seek over the encoded form — same algorithm, no separate
+  ** descent path. */
+  u8 keyBuf[8];
+  prollyEncodeIntKey(intKey, keyBuf);
+  return prollyCursorSeekBlob(cur, keyBuf, 8, pRes);
 }
 
 int prollyCursorSeekBlob(ProllyCursor *cur,
--- a/src/prolly_diff.c
+++ b/src/prolly_diff.c
@@ -35,17 +35,11 @@ static void diffFillKey(
   ProllyCursor *pCur,
   u8 flags
 ){
-  if( flags & PROLLY_NODE_INTKEY ){
-    pChange->intKey = prollyCursorIntKey(pCur);
-    pChange->pKey   = 0;
-    pChange->nKey   = 0;
-  }else{
-    const u8 *pKey; int nKey;
-    prollyCursorKey(pCur, &pKey, &nKey);
-    pChange->pKey   = pKey;
-    pChange->nKey   = nKey;
-    pChange->intKey = 0;
-  }
+  const u8 *pKey; int nKey;
+  prollyCursorKey(pCur, &pKey, &nKey);
+  pChange->pKey   = pKey;
+  pChange->nKey   = nKey;
+  pChange->intKey = (flags & PROLLY_NODE_INTKEY) ? prollyCursorIntKey(pCur) : 0;
 }
 
 static int diffIterCopyKey(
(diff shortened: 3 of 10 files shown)


LaPingvino:
## Note on how we got here (LLM prompting strategy)

Posting in case it's useful for future LLM-driven structural refactors. This PR went through several false starts before landing the right thing, and the framings that unblocked it were specific:

**1. "Wait with benchmarks until correctness parity, and don't claim wins until natural parity."**
This was the most important one. My first attempt at this PR (the original strategy-drop scope) was framed as a perf fix with a wrong 100× number. The correction was painful but cheap. The real lesson — *don't measure perf at all until the algorithm is structurally the same on both sides* — is what made the eventual numbers (222× → 1.55×) trustworthy. They aren't a benchmark engineering trick; they're what falls out when the code paths actually match.

**2. "Pretend I'm Tim and you'll have to get this job done."**
Pushed me from "propose a refactor" to "land a refactor." Without that framing I would have written a plan doc and stopped. With it, I committed to one-commit-per-green-build discipline (every commit on this branch builds clean and passes a smoke test before the next is started), which in turn made the work safe to do across a context-window compaction in the middle.

**3. The compaction itself.**
Before compacting, I wrote a comprehensive plan doc to `/tmp/doltlite_src/DOLTLITE_REWRITE_PLAN.md` covering the divergence map, the rebase strategy, exact file targets per phase, and tactical guidance. After compaction I had to re-discover that the on-disk format (`prollyNodeIntKey:143`) was already using sortable BE encoding — which I'd missed in the pre-compaction plan. That single discovery is what made Phase 1 tractable: the chunker + on-disk paths didn't need separate work, only the mutmap had to catch up to a format the rest of the system was already using.

**4. A self-paced `/loop check if we are on true natural parity yet`.**
Each tick was: assess parity status honestly, attack one isolated divergence (mutmap storage, then flush, then diff, then three-way diff, then cursor seek), build, smoke test, commit, schedule next tick. The discipline of *answering the parity question literally* — \"NO, but here's the next concrete step\" — kept me from declaring victory at the algorithm-internals level and skipping the consumer-facing surfaces. The final answer became YES only when the head-to-head bench made it falsifiable.

**What didn't help:** the original probe edit that tried to do the whole refactor in one shot. Reverting it and attacking the layers in dependency order (storage → flush → diff → cursor) made each step verifiable.

LaPingvino:
## Correction (sorry, again)

The 222× → 1.55× headline I posted earlier today was wrong. I never re-measured master after rebasing the parity branch onto it, and assumed the gap from the older prior-conversation snapshot was still there. It wasn't — Tim's #701 and #722 had already closed it.

Properly-measured numbers are now in the rewritten PR description: storage is byte-identical between master and parity, time is within noise. The PR is structural cleanup (60 fewer branching sites, ~250 lines net), not a perf change.

This is the same overclaim failure mode as the original 100× claim corrected earlier in this thread — measuring against a stale baseline rather than current master. Re-measuring before any claim, especially after a rebase, is the discipline I'm trying to actually internalize this time.

LaPingvino:
## Re-correction (turns out there IS a perf win after all)

After Joop suggested running a full bahaiwritings replay against both binaries head-to-head, the real numbers came in: **18× speedup on a 36K-row bulk INSERT into a populated TEXT-PK table** (commit i9u99ijkv6: 75.14s → 4.13s), and **-27% on aggregate chunk-apply time** across the full 780-commit replay.

The synthetic linear-INSERT benches I cited in the previous correction were measuring the wrong thing — they hit a fast path that skips the mutmap-flush merge entirely. Real commits that merge new data into populated tables DO hit it, and that's where the unification work paid off.

Updated description has the full numbers. Sorry for the description churn — third pass on this PR's claims, but the third one is actually the right one.

The takeaway for me: empty-table linear-INSERT benches are useless for prolly tree work. Replay against real history (or any bench that simulates a populated tree) is what surfaces the difference between fast-path and merge-path code.

LaPingvino:
## Re-correction (this is the fourth pass; I expect it to be the last)

A code-review subagent pass on the previous description proposed an experiment I hadn't thought to run: build a third binary at \`a19009e0d\` only (the strategy-drop commit, no unification), run i9u99ijkv6 against it. The strategy-drop alone produces the 4-second result, so the headline 18× is **entirely** from \`a19009e0d\` (which is essentially this PR's original scope). The 6 byte-uniform unification commits contribute ~5% incremental on aggregate, ~1% on the worst case.

Updated description now attributes the perf win to the strategy commit and frames the unification work honestly as structural cleanup that doesn't move perf measurably.

The site count was also wrong: stated 133 → 73, actual is 173 → 84 raw occurrences (155 → 76 lines containing one). And the synthetic-bench result was understated — parity is actually 25-29% slower on small synthetics, not "within noise." Both corrected.

What I want to flag separately: the subagent didn't just verify my claim, it proposed the falsifying test that proved the claim wrong. That's a more valuable function than rubber-stamping; I could have run that test myself but didn't think to. Filing the pattern.

timsehn (commented):
Reviewed. Couple of housekeeping items first:

**Why CI didn't fire**: this PR's base branch was \`fix/non-intkey-streaming-merge\` (PR #722's branch, now deleted on origin). All our workflows were gated on \`pull_request: branches: [master]\`, so they never scheduled. I just retargeted the base to master, and #730 drops the branch filter so external PRs can't hit this silently again — once that merges, every PR fires CI regardless of base.

**On the perf change itself**: this looks right. The three-way bench methodology is solid — A/B/C with byte-identical final DBs (1,365,849,745 across all three) confirms history independence is preserved on the bahaiwritings workload. The 75s → 4.26s on the 36K bulk-INSERT-into-populated-table commit is exactly what you'd expect once the heuristic stops misclassifying it: \`M > 10000\` was sending the trailing-append shape through \`mergeWalk\` (which walks every leaf of the existing tree), when forceDescend-routed \`streamingMerge\` from #722 splices all but the rightmost path. The "rebuild on large batch is faster" intuition the heuristic encoded was assuming the depth pathology that #722 already fixed; the heuristic became obsolete and now actively hurts.

**Strategy-drop commit attribution**: I appreciate the falsifying experiment that corrected the original "18× from unification" framing. That's exactly the kind of three-way isolation that makes the perf claim trustworthy.

**On bundling vs splitting**: the strategy-drop is one line and obviously correct. The unification work is ~1200 lines and well-tested but the risk surface is the BE-sortable INT-key encoding handling INT64_MIN..INT64_MAX correctly. Your test plan covers the boundary cases (negatives, edge i64 values, diff/merge over negatives), so I'm comfortable with the bundle.

**One thing I'd want to see before approving**: now that this PR will get CI runs, let's make sure the existing oracle suites + bench-extra all green. The classic sysbench gate at 2× will catch any INTKEY regression from the unification path. After CI runs, I'll do a code-level pass on the byte-uniform path changes.

timsehn:
Closing and reopening to fire CI — base-branch retarget alone doesn't emit a triggering event for the default pull_request workflow filters. Re-opening immediately.

PR #751, opened 2026-05-06
npm: Node.js native binding (better-sqlite3-style API)

### LINKED ISSUE #629: Distribute DoltLite as an npm package (Node native binding)
## Summary

Publish a Node.js native binding to npm. Distinct from the WASM npm package — this is for server-side Node use where native performance matters and you have filesystem access.

## What needs to happen

- Pick binding mechanism (N-API preferred for ABI stability)
- Mirror `better-sqlite3`'s API where reasonable, with `dolt_*` procedures available via the standard `prepare`/`run` surface
- Pre-build native binaries for linux x64/arm64, macOS x64/arm64, win32 x64
- `npm publish` step in the release workflow on tagged releases

## Why this matters

Node is a common embedded-DB consumer for sync servers and local tooling. A local-first stack with a Node sync server should be able to drop DoltLite in as a `better-sqlite3` replacement.

Docker is the primary channel for Dolt and Doltgres today — packaging is what makes adoption happen, not raw capability.

timsehn:
Closes #629.

Ships a Node.js native binding under `npm/` that wraps `libdoltlite` through N-API. API follows a subset of [better-sqlite3](https://github.com/WiseLibs/better-sqlite3) — synchronous, prepared-statement-driven.

```javascript
const Database = require('doltlite');
const db = new Database('app.db');

db.exec(`CREATE TABLE t(id INTEGER PRIMARY KEY, v TEXT)`);
db.prepare(`INSERT INTO t(v) VALUES (?)`).run('hello');
db.prepare(`SELECT * FROM t`).all();   // [{ id: 1, v: 'hello' }]

db.exec(`SELECT dolt_commit('-A','-m','init')`);
db.exec(`SELECT dolt_branch('feat')`);
db.exec(`SELECT dolt_checkout('feat')`);
// ...
```

The `dolt_*` procedures are reachable via the standard `prepare`/`run` surface — they're SQL functions, no special API needed.

## API covered

| Class | Methods / Properties |
|---|---|
| `Database` | `new Database(path)`, `prepare(sql)`, `exec(sql)`, `close()`, `open` (bool), `memory` (bool) |
| `Statement` | `run(...params)`, `get(...params)`, `all(...params)`, `finalize()` |
| Module | `version()` |

Bind types: `NULL`, `INTEGER` (int64), `REAL`, `TEXT`, `BLOB` (`Buffer`); `BigInt` for int64s past 2^53. Output rows are plain objects keyed by column name.

## Build wiring

`binding.gyp` links against `libdoltlite.a` from the in-tree build dir. Path resolves via the `DOLTLITE_BUILD` env var, defaulting to `../build`. Future prebuild distribution will ship per-platform `.a` + `.node` artifacts with the package; for now the package is consumed from a doltlite source checkout:

```bash
cd build && ../configure && make doltlite-lib
cd ../npm && DOLTLITE_BUILD=$(pwd)/../build npm install
node test/basic.js
```

## Test plan

Local on macOS arm64:

| Suite | Result |
|---|---|
| `npm/test/basic.js` | **15 / 15 passed** |

Covers: `open`, `exec(DDL)`, `prepare`/`run` with positional `?`, `prepare`/`all`, `prepare`/`get` (hit + miss), `dolt_commit` + `dolt_log` readback, `dolt_branch` + `dolt_checkout`, cross-branch row count, `dolt_merge`, `close()`.

## CI

`.github/workflows/npm.yml` builds libdoltlite + the binding + runs the smoke test on `linux-x64` and `osx-arm64`. Triggers on PRs that touch `npm/`, `src/`, or `main.mk` so doltlite-side changes that break the binding ABI surface immediately.

Windows is deliberately deferred — landing on linux/macOS unblocks the customer first.

## What's next (separate PRs)

- Async query API (current binding is fully synchronous, matching better-sqlite3's design choice).

### DIFF
--- a/.github/workflows/npm.yml
+++ b/.github/workflows/npm.yml
@@ -1,0 +1,71 @@
+name: NPM Native Binding
+
+# Verifies the npm/ Node native binding builds and the smoke test
+# passes on each supported platform. Also runs on PR so we know if
+# a doltlite-side change broke the binding ABI.
+#
+# Does NOT publish to npm — that's a manual step on tagged releases.
+
+on:
+  push:
+    branches: [master]
+    paths:
+      - 'npm/**'
+      - 'src/**'
+      - 'main.mk'
+      - '.github/workflows/npm.yml'
+  pull_request:
+    paths:
+      - 'npm/**'
+      - 'src/**'
+      - 'main.mk'
+      - '.github/workflows/npm.yml'
+
+jobs:
+  build-and-test:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-latest
+            target: linux-x64
+          - os: macos-latest
+            target: osx-arm64
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - uses: actions/setup-node@v4
+      with:
+        node-version: '20'
+
+    - name: Install dependencies (Linux)
+      if: runner.os == 'Linux'
+      run: sudo apt-get update && sudo apt-get install -y build-essential zlib1g-dev tcl-dev
+
+    - name: Install dependencies (macOS)
+      if: runner.os == 'macOS'
+      run: brew install zlib tcl-tk || true
+
+    - name: Build libdoltlite
+      run: |
+        mkdir -p build && cd build
+        ../configure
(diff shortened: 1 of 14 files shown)


timsehn:
Decided this should be in its own repo: https://github.com/dolthub/doltlite-node

PR #789, opened 2026-05-09
Fix flaky amend_like_flow_soft_reset_recommit oracle test

timsehn:
## Summary

\`amend_like_flow_soft_reset_recommit\` was failing oracle CI on PRs that touched no code (#786, #787) — i.e., the test itself was broken, not flaky in the runtime-only sense. Two orthogonal issues, both fixed here.

### (1) Oracle harness: bare-integer query results were silently stripped

The previous \`oracle()\` in \`vc_oracle_feature_interaction_test.sh\` piped \`setup + query\` through doltlite in one invocation and used \`grep -v '^[0-9]*$'\` to strip \`dolt_add\` / \`dolt_commit\` return-value lines out of the mixed output. That filter also matched **any legitimate scalar query result** — \`count(*)\`, sums, etc. — so any test whose query returned a bare integer compared an empty doltlite output against dolt's actual value.

The dolt branch right below already does the right thing: \`setup\` is run in a discarded-output invocation, then the query runs in a fresh invocation with just its output captured (CSV header dropped, body kept). Mirror that pattern on the doltlite side. No more noise to filter, no more false stripping.

### (2) Test setup: explicit re-stage after \`dolt_reset --soft\`

Git/Dolt's \`--soft\` keeps the index, but doltlite's soft-reset currently clears the staged set, so the subsequent \`dolt_commit('-m','amended')\` fails with \`"nothing to commit, working tree clean"\` and the amend never lands. Adding an explicit \`dolt_add('-A')\` after the soft-reset re-stages the (still-present) working tree changes and makes the test deterministic across either staging behavior. The amend-like flow being tested — soft-reset + recommit replaces the wrong commit — still exercises end-to-end.

(Whether doltlite's soft-reset *should* preserve staging is a separate question to file; this PR makes the existing oracle pass without depending on the answer.)

## Verification
- doltlite output for the test's query now returns \`count = 2\` (was empty after filter, then comparing against dolt's \`2\`).
- Messages match across engines: \`{amended, original}\`.

### DIFF
--- a/test/vc_oracle_feature_interaction_test.sh
+++ b/test/vc_oracle_feature_interaction_test.sh
@@ -34,12 +34,19 @@ oracle() {
   mkdir -p "$dir/dl" "$dir/dt"
 
   # --- doltlite ---
+  # Mirror the dolt path below: run setup first with its output
+  # discarded, then run the query in a fresh invocation against the
+  # same db file with only its output captured. The earlier single-
+  # invocation form mixed dolt_add / dolt_commit return values with
+  # the query result and tried to filter the noise out post-hoc;
+  # that filter also stripped legitimate count(*) / scalar results,
+  # which made any test whose query returned a bare integer compare
+  # an empty doltlite output against dolt's actual value.
+  printf "%s\n" "$setup" | "$DOLTLITE" "$dir/dl/db" \
+      >/dev/null 2>"$dir/dl.err"
   local dl_out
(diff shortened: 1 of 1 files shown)


timsehn:
Folding this into PRs #786 and #787 directly (the test was broken, not specific to either PR's code changes). Filing a follow-up issue for the underlying staging behavior.

PR #809, opened 2026-05-12
Size sort-key buffers exactly

timsehn:
## Summary
- compute the encoded sort-key size before reallocating reusable buffers
- keep sort-key size arithmetic in sqlite3_int64 and return SQLITE_TOOBIG before int overflow
- use sqlite3_realloc64 and add a C regression for exact buffer sizing with escaped blob keys

## Tests
- make doltlite
- make libdoltlite.a (from build/)
- cc -g -I. -I../src -o doltlite_regression_test_c ../test/doltlite_regression_test_c.c libdoltlite.a -lz -lpthread -lm && ./doltlite_regression_test_c sortkey_buffer_exact_size (from build/)
- bash ../test/doltlite_regression_test_c.sh (from build/)
- bash test/review_regression_test.sh
- bash test/doltlite_parity.sh

### DIFF
--- a/src/sortkey.c
+++ b/src/sortkey.c
@@ -2,6 +2,7 @@
 #ifdef DOLTLITE_PROLLY
 
 #include "sortkey.h"
+#include <limits.h>
 #include <string.h>
 
 static int skGetVarint32(const u8 *p, u32 *pVal){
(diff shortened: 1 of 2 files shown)


timsehn:
Closing in favor of non-stacked PR #810 based directly on master.

PR #842, opened 2026-05-13
chunk_store: verify chunk body hash during WAL replay

timsehn:
## Summary

Follow-up to #841. That PR added hash verification at read time in `chunkStoreGet`, catching torn writes / bit-flips / bad WAL-replay bodies when the chunk is first served. This PR catches them eagerly during WAL replay so a corrupted body is detected at DB open rather than at first read.

## Fix

In `csReplayWal`, after the existing torn-write bounds check and before the pending-entry add: read the body, BLAKE3-hash it, compare to the header's declared hash. On mismatch, `break` out of the replay loop — same fail-fast pattern the surrounding code uses for partial-write detection (`(u64)pos + len > walSize`). Earlier (verified) WAL entries are kept.

## Cost

One BLAKE3 over each WAL chunk body at DB open. The bodies were already going to be paged in eventually on first read; this just pulls the work forward. For DBs with many uncommitted WAL chunks at open this is measurable but bounded; for the steady-state case (small WAL) it's negligible.

## Smoke tested

Write + commit + uncommitted insert → reopen → reads return correct rows; the WAL-replay path now goes through the new verify and accepts the legitimate chunks.

### DIFF
--- a/src/chunk_store.c
+++ b/src/chunk_store.c
@@ -1016,6 +1016,19 @@ static int csReplayWal(ChunkStore *cs){
       }
 
       {
+        u8 *pBody = (u8 *)sqlite3_malloc((int)len);
+        ProllyHash bodyHash;
+        if( pBody == 0 ){ rc = SQLITE_NOMEM; goto replay_error; }
+        rc = sqlite3OsRead(cs->pFile, pBody, (int)len, cs->iWalOffset + pos);
+        if( rc != SQLITE_OK ){ sqlite3_free(pBody); goto replay_error; }
+        prollyHashCompute(pBody, (int)len, &bodyHash);
+        sqlite3_free(pBody);
+        if( memcmp(&bodyHash, &hash, sizeof(ProllyHash)) != 0 ){
+          break;
+        }
+      }
+
+      {
         int existing = csSearchIndex(cs->aIndex, cs->nIndex, &hash);
         ChunkIndexEntry *e = 0;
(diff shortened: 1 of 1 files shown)


timsehn:
Closing as superseded by behavior already provided by #841.

CI surfaced two real regressions from the eager `csReplayWal` verify:

1. `integrity_check_surfaces_root_corruption` (`doltlite_regression_test_c.c:3203`) corrupts a chunk that's still in the WAL after `dolt_commit`. With this PR, `csReplayWal` rejects the corrupted chunk on mismatch and drops it from pending entirely — `chunkStoreGet` then returns `SQLITE_NOTFOUND` rather than `SQLITE_CORRUPT`, and `PRAGMA integrity_check` reports "ok" because there's nothing to walk. Master's behavior (or #841 alone) is correct: the chunk goes into pending, `chunkStoreGet` re-hashes per #841, returns `CORRUPT` on mismatch, integrity_check reports non-ok.

2. `large_db_sparse_open` (`review_regression_test.sh` GUARD 22) creates a >2GB sparse DB where random data lives in WAL-region offsets. My `sqlite3_malloc((int)len)` allocates whatever `len` claims; on sparse garbage that's enormous → OOM at open.

#841 alone catches the same corruption surface on first read, which is the point that matters for correctness. Catching at open instead of at first query was a nice-to-have but is not load-bearing and breaks legitimate scenarios.

PR #846, opened 2026-05-13
docs: use doltlite.h in README and quickstart example

timsehn:
Follows #843 which installs all three headers (`sqlite3.h`, `doltlite.h`, `doltlite_remotesrv.h`) as aliases. Now that `doltlite.h` is reliably present after install, switching the README narrative + `examples/quickstart.c` to use it as the canonical include. The `sqlite3.h` alias stays installed so existing code keeps compiling unchanged.

- `README.md` "Using as a C Library" paragraph reworded to lead with `doltlite.h` and note `sqlite3.h` is the back-compat alias.
- `examples/quickstart.c` switched to `#include "doltlite.h"`; verified `cc -c` clean against the in-tree header.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,9 +125,10 @@ make -C ext/wasm dist
 
 ## Using as a C Library
 
-Doltlite is designed as a drop-in replacement for SQLite. It uses the same
-`sqlite3.h` header and `sqlite3_*` API, so existing C programs work without
-code changes — just link against `libdoltlite` instead of `libsqlite3` to get
+Doltlite is designed as a drop-in replacement for SQLite. New code should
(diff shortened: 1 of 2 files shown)


timsehn:
Closing — master's README narrative and examples/quickstart.c both reverted to \`sqlite3.h\` as the canonical name (post #843, which kept \`doltlite.h\` as an install-tree alias). This PR's premise (rename canonical → \`doltlite.h\`) is no longer wanted.

The CI failure here was the natural consequence: \`build-and-test\` compiles \`examples/quickstart.c\` against the build directory, which only has \`sqlite3.h\` available — the \`doltlite.h\` alias only exists in the install tree (Homebrew, .deb, install.sh paths from #843), not in the build output.

PR #911, opened 2026-05-14
Avoid ordered inserts for lazy mutmaps

timsehn:
## Target

Checked the latest merged doltlite PR, #909. In the sysbench comments, ignoring `index_join_scan`, the highest multiplier was:

- Key type: classic `INTEGER PRIMARY KEY` / rowid table
- Mode: in-memory
- Autocommit: no
- Benchmark: `oltp_update_index`
- PR comment: SQLite `32,653 us`, Doltlite `54,911 us`, `1.68x`

## Change

Lazy mutmaps were still maintaining sorted order after a seek had materialized order once. A write-heavy transaction can then hit `insertOrderEntry()` for later random inserts/deletes and pay ordered-array movement on every write.

For lazy mutmaps, this PR always appends new edits and marks order dirty. Ordered readers still call `ensureOrder()` when they need sorted traversal. `keepSorted` maps keep the old behavior.

This is a general write-path change for lazy mutmaps, not a benchmark-specific special case.

## Local comparison

Clean latest `origin/master` before this change, full classic sysbench, `BENCH_RUNS=7`:

| Mode | Benchmark | SQLite us | Doltlite us | Ratio |
| --- | --- | ---: | ---: | ---: |
| In-memory | `oltp_update_index` | 17,645 | 27,408 | 1.55 |
| File-backed | `oltp_update_index` | 21,208 | 31,529 | 1.49 |
| File-backed autocommit | `oltp_update_index_ac` | 19,022 | 34,321 | 1.80 |

This branch, full classic sysbench, `BENCH_RUNS=7`:

| Mode | Benchmark | SQLite us | Doltlite us | Ratio |
| --- | --- | ---: | ---: | ---: |
| In-memory | `oltp_update_index` | 18,351 | 28,844 | 1.57 |
| File-backed | `oltp_update_index` | 20,999 | 35,648 | 1.70 |
| File-backed autocommit | `oltp_update_index_ac` | 18,510 | 33,324 | 1.80 |

That 7-run full-matrix result was noisy on the small row. A higher-run wrapped comparison gave a cleaner Doltlite-to-Doltlite signal for the target row:

| Build | Mode | Benchmark | Doltlite us |
| --- | --- | --- | ---: |
| Clean latest `origin/master` | In-memory | `oltp_update_index` | 30,016 |
| This branch | In-memory | `oltp_update_index` | 29,078 |

The longer in-memory indexed-update workload, using the same table/index/update shape with 200k updates to expose the ordered-insert cost, improved from `2.16s` on clean latest `origin/master` to `0.62s` on this branch. Instructions retired dropped from about `25.9B` to `5.7B`.

The full local sysbench comparison still exits non-zero due unrelated local autocommit ceilings.

## Validation

- `make doltlite`
- `DOLTLITE_BUILD_DIR=. bash test/doltlite_regression_test_c.sh` (`108631 passed, 0 failed`)
- `git diff --check`

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -29,7 +29,7 @@ jobs:
     - name: Configure
       run: |
         mkdir -p build && cd build
-        TCL_CONFIG=$(find /usr -name tclConfig.sh 2>/dev/null | head -1)
+        TCL_CONFIG=$(find /usr -path '/usr/share/miniconda/*' -prune -o -name tclConfig.sh -print 2>/dev/null | head -1)
         if [ -n "$TCL_CONFIG" ]; then
           ../configure --with-tcl=$(dirname $TCL_CONFIG)
         else
@@ -140,7 +140,7 @@ jobs:
     - name: Configure
       run: |
         mkdir -p build && cd build
-        TCL_CONFIG=$(find /usr -name tclConfig.sh 2>/dev/null | head -1)
+        TCL_CONFIG=$(find /usr -path '/usr/share/miniconda/*' -prune -o -name tclConfig.sh -print 2>/dev/null | head -1)
         if [ -n "$TCL_CONFIG" ]; then
           ../configure --with-tcl=$(dirname $TCL_CONFIG)
         else
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -376,7 +376,7 @@ jobs:
     - name: Configure
       run: |
         mkdir -p build && cd build
-        TCL_CONFIG=$(find /usr -name tclConfig.sh 2>/dev/null | head -1)
+        TCL_CONFIG=$(find /usr -path '/usr/share/miniconda/*' -prune -o -name tclConfig.sh -print 2>/dev/null | head -1)
         if [ -n "$TCL_CONFIG" ]; then
           ../configure --with-tcl=$(dirname $TCL_CONFIG)
         else
--- a/.github/workflows/sanitizers.yml
+++ b/.github/workflows/sanitizers.yml
@@ -21,7 +21,7 @@ jobs:
     - name: Configure
       run: |
         mkdir -p build && cd build
-        TCL_CONFIG=$(find /usr -name tclConfig.sh 2>/dev/null | head -1)
+        TCL_CONFIG=$(find /usr -path '/usr/share/miniconda/*' -prune -o -name tclConfig.sh -print 2>/dev/null | head -1)
         if [ -n "$TCL_CONFIG" ]; then
           ../configure --with-tcl=$(dirname $TCL_CONFIG)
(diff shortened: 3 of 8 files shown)


timsehn:
Updated PR 911 after the CI failure.

Failure analysis:
- build-and-test failed in the quick large-scale suite: update 50% of 100k rows took 39s > 15s.
- benchmark failed because the original lazy mutmap change regressed mixed read/write ordered workloads, especially classic oltp_read_write: in-memory 2.84x and file-backed 2.71x in the PR comment.

Fix:
- Keep the fast unordered append path for lazy exact-write mutation maps.
- Once a mutmap is used for ordered iteration/seeks, mark it as preferring sorted order so later inserts/deletes maintain order incrementally instead of forcing repeated full re-sorts.

Local verification on the updated commit:
- test/large_scale_test.sh ./doltlite --quick: 19 passed, 0 failed; the 50% update step completed within the 15s ceiling.
- DOLTLITE_BUILD_DIR=. bash test/doltlite_regression_test_c.sh: 108631 passed, 0 failed.
- BENCH_RUNS=7 classic benchmark spot check: oltp_read_write is back near baseline (in-memory 1.47x, file-backed 1.39x) while oltp_update_index stays improved (in-memory 1.54x, file-backed 1.53x).

timsehn:
One more CI-only update pushed after the rerun: vc-concurrency-stress failed during configure because the runner selected /usr/share/miniconda/.../tclConfig.sh, which points at a non-usable tclsh path. I patched the workflow Tcl discovery to prune /usr/share/miniconda across the CI workflows that use this pattern, so configure should pick the system tcl-dev install instead.

timsehn:
Pushed a fix for the non-int PK benchmark regression.

Cause: the previous fix only set `preferSorted` in explicit iterator entry points. Non-int write workloads also reach ordered access through helpers like sorted-position resolution, so they could still repeatedly materialize order, append dirty, and materialize again.

Fix: `ensureOrder()` now marks the mutmap as preferring sorted maintenance. That covers every ordered access path while preserving the lazy append fast path for maps that stay exact-lookup only.

Local spot checks with BENCH_RUNS=7:
- TEXT PK: `oltp_insert` ~1.5x, `oltp_update_index` ~1.5x, `oltp_read_write` ~1.6-1.7x. This was ~45x insert / ~3x update in the PR benchmark.
- BLOB PK: same recovery, insert/update/read-write back around ~1.4-1.7x.
- Composite PK: same recovery, insert/update/read-write back around ~1.4-1.6x.
- Classic still keeps the intended path: `oltp_update_index` ~1.5-1.6x and `oltp_read_write` ~1.4-1.5x locally.

Also reran `DOLTLITE_BUILD_DIR=. bash test/doltlite_regression_test_c.sh`: 108631 passed, 0 failed.

PR #927, opened 2026-05-15
fix(chunk-store): hash-verify WAL chunk bodies, fsync between chunks and root

timsehn:
## Summary

S6 + S7 from the deep review — two crash-safety holes that defeated the WAL's torn-write recovery story.

### S6 — \`chunk_wal.c\`: WAL replay accepts unverified chunk bodies

Replay read the 24-byte chunk header (hash + length) from the WAL, then registered a pending index entry pointing at the on-disk body **without ever reading or re-hashing the body**. A torn write that flipped a bit anywhere in the body was silently committed into the index. The next \`chunkStoreGet\` for that hash would surface \`SQLITE_CORRUPT\`, and a future \`chunkStorePut\` with the same hash would see it \"exists\" and drop the good data on the floor.

**Fix:** read the body, recompute its hash via \`prollyHashCompute\`, and on mismatch stop replay at the last valid commit boundary — matching the existing torn-tail behavior for header/manifest truncation.

### S7 — \`chunk_store.c\`: commit fsync ordering

The commit path wrote all chunk records, then the root record, then did **one** fsync at the end. The kernel and the disk are both free to reorder writes between the application's syscalls and the actual durable medium, so a power loss could persist the root with \`nChunks\`/\`refsHash\` set while the chunk bodies it references are still in volatile page cache. Recovery would then read the root, look up a referenced hash, and get \`SQLITE_CORRUPT\`.

**Fix:** fsync after the chunk writes and **before** the root record. The root is the commit point; ordering the chunk durability ahead of it makes the atomic-at-the-chunk-level guarantee actually hold. The existing fsync after the root is kept (commit-durability for the caller).

## Test plan

- [x] \`correctness_test.sh\`: 41/41
- [x] \`doltlite_regression_test_c.sh\`: 108637/108637 (covers existing WAL corruption cases including \`wal_mid_corruption_rejected\`, \`truncated_wal_rejected\`, \`wal_offset_corruption\`)
- [x] \`corruption_test\`: 54/54

### DIFF
--- a/src/chunk_store.c
+++ b/src/chunk_store.c
@@ -1214,7 +1214,15 @@ static int csCommitToFile(ChunkStore *cs){
     }
   }
 
-  /* The root record is the commit point for the append-only chunk store. */
+  if( cs->staging.nPending > 0 ){
+    CRASH_CHECK_WRITE();
+    rc = sqlite3OsSync(cs->file.pFile, SQLITE_SYNC_NORMAL);
+    if( rc != SQLITE_OK ) goto commit_done;
+  }
+
+  /* The root record is the commit point for the append-only chunk store.
+  ** It is written after the preceding fsync so the kernel cannot reorder
+  ** the root ahead of the chunk bodies it references. */
   {
     u8 rootRec[1 + CHUNK_MANIFEST_SIZE];
     rootRec[0] = CS_WAL_TAG_ROOT;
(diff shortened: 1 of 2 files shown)


timsehn:
Superseded by https://github.com/dolthub/doltlite/pull/927 — content-addressing dedup-verify covers the same crash-safety concern without per-commit fsync.

PR #982, opened 2026-05-19
doltlite revert to better match git

### LINKED ISSUE #961: dolt_revert refuses on unrelated dirty tables (stricter than git revert)
## Summary

`dolt_revert(hash)` refuses to run when the working tree has uncommitted changes in **any** table, even if those tables are completely unrelated to the commit being reverted. This makes the function hard to use in any application that keeps lightweight metadata alongside the version-controlled tables (audit log rows, last-seen timestamps, queued background work, etc.) — the metadata write dirties the tree and blocks unrelated reverts.

Git's `git revert` only refuses when the revert would *conflict* with WIP, not when WIP exists anywhere in the index.

Tested on doltlite v0.10.8 prebuilt (Linux x64). Pure-SQL repro using the bundled `doltlite` CLI; no Python or app code involved.

### Case A — revert non-HEAD with a fully clean tree: **works** (control)

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, x TEXT);
SELECT dolt_commit('-Am', 'schema',  '--author', 'a <a@b>');
INSERT INTO t VALUES(1, 'a');
SELECT dolt_commit('-Am', 'add 1',   '--author', 'a <a@b>');  -- this is the one we'll revert
INSERT INTO t VALUES(2, 'b');
SELECT dolt_commit('-Am', 'add 2',   '--author', 'a <a@b>');  -- so 'add 1' is HEAD~1
-- tree is clean
SELECT dolt_revert((SELECT commit_hash FROM dolt_log WHERE message='add 1'));
```
Output: succeeds, prints new commit hash, t now contains only id=2. ✓

### Case B — revert HEAD with an uncommitted row in a different table: **fails**

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, x TEXT);
CREATE TABLE meta(id INTEGER PRIMARY KEY, note TEXT);
SELECT dolt_commit('-Am', 'schema', '--author', 'a <a@b>');
INSERT INTO t VALUES(1, 'a');
SELECT dolt_commit('-Am', 'add row', '--author', 'a <a@b>');   -- HEAD
-- a totally unrelated, uncommitted write in `meta`:
INSERT INTO meta VALUES(1, 'side');
-- dolt_status:  meta|0|modified   (t is clean)
SELECT dolt_revert((SELECT commit_hash FROM dolt_log LIMIT 1));
```
Output: `Error near line 11: cannot revert with uncommitted changes` ✗

The commit being reverted only touches `t`. Reverting it cannot conflict with the `meta` change. But `dolt_revert` refuses anyway.

## Why this matters

A common pattern when wiring doltlite into an existing app is to keep a *metadata* table next to the user tables — recording per-commit context the dolt log doesn't natively carry (who issued the API request, which client session, the human-readable message, links to derived artifacts, etc.). The natural sequence is:

1. User write → `dolt_commit('-A', ...)`
2. `INSERT INTO commit_metadata(dolt_hash, user, message, ...) VALUES(?, ...)`
3. Return the metadata row id to the caller

After step 2, the working tree has one uncommitted row in `commit_metadata`. If the user immediately clicks "undo" — the most common revert pattern — step 1's hash is HEAD, the user table is clean, but **revert refuses** because of the unrelated metadata write. The only workarounds are clunky:

- Sync the metadata with another `dolt_commit` first (which pushes the user's commit off HEAD — also problematic since revert chains may be confusing).
- Move metadata into a separate database file (loses atomicity and adds a second connection).
- Implement revert manually as inverse-diff replay (giving up the use of `dolt_revert` entirely).

We hit this wiring up a label-tracking application; see the wider context at [thread] (will link).

## Suggested behavior

Match Git's semantics: refuse only when the revert would **modify rows that have uncommitted changes**. Specifically, walk `dolt_diff_<table>(target_commit)` for the commit being reverted and check whether any of those (table, primary key) pairs are currently dirty. If not, proceed.

A weaker but still useful improvement would be to refuse only when the **same table** has uncommitted changes (table-level granularity vs Git's row-level), leaving unrelated tables alone.

## Environment

- Linux 6.8.0-117-generic, Ubuntu 22.04
- doltlite v0.10.8 (Linux x64 prebuilt: `doltlite-lib-linux-x64-0.10.8.zip`, sha256 `9f9b635c…cd0c`)
- Reproduced with the bundled `doltlite` CLI; no Python / language binding involvement.

macneale4:
Fixes https://github.com/dolthub/doltlite/issues/961

### DIFF
--- a/src/doltlite.c
+++ b/src/doltlite.c
@@ -3523,43 +3523,28 @@ static int doltliteCollectRevertTouchedTables(
   return rc;
 }
 
-// Captures TableEntry data for each unstaged-dirty table whose name is NOT in
-// the touched list. Sets *pConflict=1 when (a) a staged change exists (matching
-// git: any staged change blocks revert) or (b) an unstaged-dirty table is in
-// the touched list. Returns SQLITE_OK with *paOut = NULL when nothing to save.
-// Caller frees with doltliteFreeCatalog.
-static int doltliteSnapshotDirtyUnrelated(
+// Sets *pConflict=1 if the working set has an uncommitted change to a table
+// the revert would modify. Changes to unrelated tables are allowed and become
+// part of the revert commit, matching Dolt.
+static int doltliteRevertHasDirtyTouchedTables(
   sqlite3 *db,
   char **azTouched, int nTouched,
-  int *pConflict,
-  struct TableEntry **paOut, int *pnOut
+  int *pConflict
 ){
   ChunkStore *cs = doltliteGetChunkStore(db);
-  ProllyHash headCatHash, stagedHash, workingCatHash;
+  ProllyHash headCatHash, workingCatHash;
   u8 *wBuf = 0; int nWBuf = 0;
-  struct TableEntry *aHead = 0, *aStaged = 0, *aWorking = 0;
-  int nHead = 0, nStaged = 0, nWorking = 0;
-  struct TableEntry *aOut = 0;
-  int nOut = 0;
+  struct TableEntry *aHead = 0, *aWorking = 0;
+  int nHead = 0, nWorking = 0;
   int i, rc;
 
   *pConflict = 0;
-  *paOut = 0;
-  *pnOut = 0;
 
   if( !cs ) return SQLITE_OK;
   if( !doltliteHasUncommittedChanges(db) ) return SQLITE_OK;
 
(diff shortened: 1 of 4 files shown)


timsehn:
Have to ait for this change: https://github.com/dolthub/dolt/pull/11073 to be released before we can merge this.

PR #1002, opened 2026-05-21
Optimize covering index range counts

timsehn:
## Summary
- add a `CountIndexRange` VDBE opcode for simple `count(*)` / `count(indexed_col)` over single-column indexed `BETWEEN` ranges
- implement DoltLite prolly index rank counting using sort-key prefixes, with stock btree falling back to the generic scan path
- add count performance regression coverage for duplicate index keys, affinity on bounds, NULL bounds, and nullable count arguments

## Benchmark
Current master wrapped baseline, 10k rows:
- file-backed `covering_index_scan`: SQLite 8,539 us, DoltLite 5,973 us, multiplier 0.70

This branch wrapped run, 10k rows:
- file-backed `covering_index_scan`: SQLite 8,755 us, DoltLite 2,709 us, multiplier 0.31
- in-memory `covering_index_scan`: SQLite 4,364 us, DoltLite 2,037 us, multiplier 0.47

Focused long workload:
- before: 592,378 us
- after: 237,767 us

## Tests
- `make -j$(sysctl -n hw.ncpu) doltlite doltlite-lib && cc -O2 -I. -I../src -o bench_timer_doltlite test/sysbench_timer.c libdoltlite.a -lpthread -lz -lm`
- `bash test/doltlite_count_perf.sh`
- `BENCH_SECTION_MODE=wrapped BENCH_ROWS=10000 bash test/sysbench_compare.sh > /tmp/doltlite_sysbench_covering_patched.md`
- `make DOLTLITE_PROLLY=0 sqlite3`
- stock `./sqlite3 :memory:` indexed range count fallback smoke test
- `git diff --check`

### DIFF
--- a/src/btree.c
+++ b/src/btree.c
@@ -10602,6 +10602,21 @@ int sqlite3BtreeCountRange(
   return rc;
 }
 
+int sqlite3BtreeCountIndexRange(
+  sqlite3 *db,
+  BtCursor *pCur,
+  UnpackedRecord *pLower,
+  UnpackedRecord *pUpper,
+  i64 *pnEntry
+){
+  (void)db;
+  (void)pCur;
+  (void)pLower;
+  (void)pUpper;
+  (void)pnEntry;
+  return SQLITE_NOTFOUND;
+}
+
 /*
 ** Return the pager associated with a BTree.  This routine is used for
 ** testing and debugging only.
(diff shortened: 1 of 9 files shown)


timsehn:
Fixed the sqllogictest crash from the covering-index count fast path. The optimization now only accepts constant/parameter BETWEEN bounds, so row-dependent bounds fall back to normal scan/codegen instead of evaluating without a row cursor.\n\nValidation run locally:\n- bash test/doltlite_count_perf.sh\n- git diff --check

PR #1007, opened 2026-05-21
Cache newly chunked prolly nodes

timsehn:
## Summary
- cache prolly nodes as they are emitted by the chunker
- pass the btree prolly cache through mutation flushes
- reuse cached roots during streaming merges to avoid hot chunk lookup/reparse work on repeated writes

## Benchmark
Isolated file-backed int-key `oltp_insert_ac` workload, 200 autocommit inserts:
- baseline median: `34,147us`
- patched median: `29,069us`
- median improvement: `14.9%`

Longer 5000-insert samples:
- baseline: `806ms`, `913ms`, `737ms`
- patched: `612ms`, `716ms`, `688ms`

## Tests
- `make -j8 doltlite libdoltlite.a`
- `./doltlite_regression_test_c all`
- manual indexed autocommit insert smoke tests with `PRAGMA integrity_check`

### DIFF
--- a/src/chunk_store.c
+++ b/src/chunk_store.c
@@ -1012,6 +1012,7 @@ static int csCommitToFile(ChunkStore *cs){
   int lockHeld = (cs->graphLockFd >= 0);
   i64 newWalSize = cs->wal.nWalData;
   ChunkIndexEntry *aCommittedPending = 0;
+  ChunkIndexEntry aSmallCommittedPending[32];
   ChunkIndexEntry *aMergePending = 0;
   ChunkIndexEntry *aMerged = 0;
   int nMerged = 0;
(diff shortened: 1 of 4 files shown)


timsehn:
Added one more write-path commit: `f58f17efff Reduce small chunk commit overhead`.

What changed:
- avoid the per-commit heap allocation for the common small pending-chunk batch
- avoid a redundant file-size syscall when the graph lock is already held and the transaction has a pinned append point

Local follow-up benchmark:
- 5-run autocommit section stayed about flat on write average: `1.98x`
- `oltp_update_index_ac` improved in the 5-run section to `1.98x` from the earlier noisy `2.1-2.2x` band
- targeted trace before cleanup showed chunk commit time over 200 indexed updates drop from about `13.2ms` to `11.7ms`

Validation:
- `make -j8 doltlite libdoltlite.a`
- `./doltlite_regression_test_c all`
- `./doltlite_regression_test_c refs_hash_commit_failure_restore`
- `./doltlite_regression_test_c remotesrv_chunk_commit_failure_clears_pending`
- `./doltlite_regression_test_c btree_commit_failure_transactional`

timsehn:
Fixed the crash-fuzz failure. The bad part was the file-size shortcut in `csCommitToFile`: after `dolt_gc()` rewrites the chunk file, `cs->file.iFileSize` can be stale while the graph lock is still held, so the post-GC commit in crash test 13 could append at the wrong offset and corrupt the DB.\n\nI amended the PR branch to remove that shortcut and kept only the small pending-chunk stack allocation from the second commit.\n\nLocal validation after the amendment:\n- `make -B -j8 doltlite_regression_test_c`\n- `./doltlite_regression_test_c all`

PR #1011, opened 2026-05-21
Batch small commit root writes

timsehn:
## Summary
- include the root record in the existing small chunk WAL batch
- reduce small write commits from chunk-batch write + root write to one append write before fsync
- keep crash-write injection on the existing separate-write path so crash coverage remains granular

## Validation
- `make -j8 doltlite libdoltlite.a`
- `./doltlite_regression_test_c all`
- autocommit insert smoke with reopen and `PRAGMA integrity_check`
- `BENCH_RUNS=5 BENCH_SECTION_MODE=autocommit ... test/sysbench_compare.sh`

## Benchmark note
Local autocommit timings remain very noisy and fsync-bound on this machine. This is a structural write-path optimization: small commits now issue one fewer VFS write syscall while preserving the same root-record commit point and keeping crash-injection behavior unchanged.

### DIFF
--- a/src/chunk_store.c
+++ b/src/chunk_store.c
@@ -1186,89 +1186,81 @@ static int csCommitToFile(ChunkStore *cs){
 
   /* Append chunks before the root record. Recovery ignores appended data until
   ** it finds a later valid root record that points at the new manifest. */
+  if( cs->staging.nPending > 0 ){
+    i64 walBytes = 0;
+    u8 *pWalBatch = 0;
+    u8 aSmallWalBatch[4096];
+    u8 *pOut = 0;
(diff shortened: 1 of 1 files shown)


timsehn:
Backing this out. The concern is valid: batching the root record with chunk data means production and crash-injection paths differ at the commit boundary. More importantly, recovery can skip a bad/truncated chunk body and still later accept a root record, so accepting a root that arrived in the same unsynced large write as its chunks is not a storage-safety property I want to rely on.

I pushed a revert to this PR. We should leave root writes separate unless/until recovery validates that all chunks referenced by a replayed root are present and hash-valid before accepting that root.

PR #1047, opened 2026-05-27
Harden GC rewrite failure handling

timsehn:
## Summary
- close the old GC chunk-store file handle once live file replacement succeeds, preventing the connection from continuing on the pre-GC unlinked file
- adopt the compacted in-memory GC metadata whenever replacement has already happened, even if a later reopen or directory sync reports an error
- route temp cleanup through SQLite VFS helpers and surface directory fsync failures
- add regression coverage for post-replace reopen failure on the GC path

## Tests
- bash test/run_doltlite_regression_case.sh gc_rewrite_failure
- bash test/doltlite_gc.sh
- bash test/doltlite_gc_session_state.sh build/doltlite
- git diff --check

### DIFF
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -169,3 +169,34 @@ jobs:
           leak-*
           timeout-*
         if-no-files-found: ignore
+
+  windows:
+    runs-on: windows-latest
+    timeout-minutes: 25
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Install Windows dependencies
+      uses: msys2/setup-msys2@v2
+      with:
+        msystem: MINGW64
+        install: mingw-w64-x86_64-gcc mingw-w64-x86_64-zlib make perl tcl
+
+    - name: Configure and build
+      shell: msys2 {0}
+      run: |
+        mkdir -p build && cd build
+        ../configure
+        make -j$(nproc) DOLTLITE_PROLLY_CHECK=1 doltlite.exe doltlite-lib
+
+    - name: GC smoke tests
+      shell: msys2 {0}
+      run: |
+        cd build
+        ./doltlite.exe /tmp/windows_smoke.db "CREATE TABLE t(id INTEGER PRIMARY KEY, v TEXT); INSERT INTO t VALUES(1,'a'); SELECT dolt_commit('-A','-m','c1'); SELECT dolt_gc(); SELECT count(*) FROM t;"
+        bash ../test/doltlite_gc.sh
+        bash ../test/doltlite_gc_session_state.sh
+        bash ../test/doltlite_pragma_page_count.sh
+        cd ..
+        bash test/run_doltlite_regression_case.sh gc_rewrite_failure
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -190,11 +190,23 @@ jobs:
         ../configure
         make -j$(nproc) DOLTLITE_PROLLY_CHECK=1 doltlite.exe doltlite-lib
 
+    - name: Basic SQL smoke tests
+      shell: msys2 {0}
+      run: |
+        cd build
+        rm -f /tmp/windows_smoke_basic.db /tmp/windows_smoke_basic.db-wal
+        test "$(./doltlite.exe :memory: "SELECT 1;")" = "1"
+        ./doltlite.exe /tmp/windows_smoke_basic.db "CREATE TABLE t(id INTEGER PRIMARY KEY, v TEXT); INSERT INTO t VALUES(1,'a'); SELECT count(*) FROM t;"
+        test "$(./doltlite.exe /tmp/windows_smoke_basic.db "SELECT v FROM t WHERE id=1;")" = "a"
+        ./doltlite.exe /tmp/windows_smoke_basic.db "SELECT dolt_commit('-A','-m','c1');"
+        test "$(./doltlite.exe /tmp/windows_smoke_basic.db "SELECT count(*) FROM dolt_log;")" = "2"
+
     - name: GC smoke tests
       shell: msys2 {0}
       run: |
         cd build
-        ./doltlite.exe /tmp/windows_smoke.db "CREATE TABLE t(id INTEGER PRIMARY KEY, v TEXT); INSERT INTO t VALUES(1,'a'); SELECT dolt_commit('-A','-m','c1'); SELECT dolt_gc(); SELECT count(*) FROM t;"
+        rm -f /tmp/windows_smoke_gc.db /tmp/windows_smoke_gc.db-wal
+        ./doltlite.exe /tmp/windows_smoke_gc.db "CREATE TABLE t(id INTEGER PRIMARY KEY, v TEXT); INSERT INTO t VALUES(1,'a'); SELECT dolt_commit('-A','-m','c1'); SELECT dolt_gc(); SELECT count(*) FROM t;"
         bash ../test/doltlite_gc.sh
         bash ../test/doltlite_gc_session_state.sh
         bash ../test/doltlite_pragma_page_count.sh
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -194,7 +194,7 @@ jobs:
       shell: msys2 {0}
       run: |
         cd build
-        rm -f /tmp/windows_smoke_basic.db /tmp/windows_smoke_basic.db-wal
+        rm -f /tmp/windows_smoke_basic.db /tmp/windows_smoke_basic.db-wal /tmp/windows_smoke_basic.db-lock
         test "$(./doltlite.exe :memory: "SELECT 1;")" = "1"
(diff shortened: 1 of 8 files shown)


timsehn:
Added a Windows-specific GC replacement hardening commit (`016e4d891d`):

- use `MoveFileExW(..., MOVEFILE_REPLACE_EXISTING | MOVEFILE_WRITE_THROUGH)` instead of C `rename()` on Windows
- close the temp and destination handles before replacement on Windows, since Windows will not reliably replace open source/destination files
- restore the destination handle if Windows replacement fails before the live file was swapped
- refactored reopen handling through a shared helper

Local verification:
- `make -C build -j8 libdoltlite.a doltlite`
- `bash test/run_doltlite_regression_case.sh gc_rewrite_failure`
- `bash ../test/doltlite_gc.sh` from `build/`
- `bash test/doltlite_gc_session_state.sh build/doltlite`
- `bash test/doltlite_pragma_page_count.sh build/doltlite`
- `git diff --check`

I could not execute the Windows API path locally on macOS; CI should cover compile/runtime behavior there.

timsehn:
Added a focused Windows GC smoke job in `.github/workflows/test.yml` using the same MSYS2 build path as release. It builds `doltlite.exe` and `doltlite-lib`, then runs:

- `test/doltlite_gc.sh`
- `test/doltlite_gc_session_state.sh`
- `test/doltlite_pragma_page_count.sh`
- `test/run_doltlite_regression_case.sh gc_rewrite_failure`

Local verification: `git diff --check -- .github/workflows/test.yml` and YAML parse passed. The actual Windows execution will validate in Actions.

timsehn:
Fixed the Windows workflow failure. MSYS2 treats `doltlite` and `doltlite.exe` as the same path, so the defensive copy failed before tests ran. Removed that copy and renamed the job to `windows-smoke`.

Local verification: `git diff --check -- .github/workflows/test.yml` and YAML parse passed.

timsehn:
Moved the Windows GC coverage into the Smoke workflow as job id `windows`, so it should report as `Smoke / windows`.

The prior test failure was from `SQLITE_FCNTL_HAS_MOVED`: the Unix VFS supports it, but the Windows VFS returned `SQLITE_NOTFOUND`, which surfaced in the CLI as `unknown operation (12)` during normal SQL. Added a Windows handler that reports not-moved. That matches the current Windows sharing mode, where an open chunk-store handle is not replaceable via the GC rename path.

Local verification:
- `make -C build -j8 libdoltlite.a doltlite`
- `bash test/run_doltlite_regression_case.sh gc_rewrite_failure`
- `cd build && bash ../test/doltlite_gc.sh`
- `cd build && bash ../test/doltlite_gc_session_state.sh && bash ../test/doltlite_pragma_page_count.sh`
- workflow YAML parse and `git diff --check`

timsehn:
Pushed `6c78228892` to address the latest `Smoke / windows` failure. The one-line sanity test got through create/commit/GC and then reported `database disk image is malformed`, so I normalized both the temp and destination paths through the active SQLite VFS before calling `MoveFileExW()`. That avoids feeding MSYS/SQLite-style paths directly to Win32 path APIs.

Local verification:
- `make -C build -j8 libdoltlite.a doltlite`
- `bash test/run_doltlite_regression_case.sh gc_rewrite_failure`
- `cd build && bash ../test/doltlite_gc.sh`
- `git diff --check -- src/doltlite_gc.c`

timsehn:
Pushed `ee4824f9b9` with another GC rewrite hardening fix: truncate the temp chunk-store file to the exact bytes written before syncing and replacing. The temp file was opened with create-but-not-truncate semantics, so a stale `*-gc-tmp` file could leave trailing bytes that look like WAL garbage after replacement.

Local verification:
- `make -C build -j8 libdoltlite.a doltlite`
- `bash test/run_doltlite_regression_case.sh gc_rewrite_failure`
- `cd build && bash ../test/doltlite_gc.sh`
- `git diff --check -- src/doltlite_gc.c`

timsehn:
Pushed `f1756474c2` to address the repeated Windows corruption. The Windows graph lock was held on the database file itself across `MoveFileExW()`. The GC path now releases that file lock before replacement and reacquires it immediately after the replace call.

Local verification:
- `make -C build -j8 libdoltlite.a doltlite`
- `bash test/run_doltlite_regression_case.sh gc_rewrite_failure`
- `cd build && bash ../test/doltlite_gc.sh`
- `cd build && bash ../test/doltlite_gc_session_state.sh`
- `cd build && bash ../test/doltlite_pragma_page_count.sh`
- `git diff --check -- src/doltlite_gc.c src/chunk_store.c src/chunk_store.h`

timsehn:
Updated the Windows lock path again.

The chunk-store graph lock now uses SQLite VFS file handles and sqlite3OsLock/sqlite3OsUnlock on a sidecar lock file instead of calling _open/LockFileEx directly from chunk_store.c. That keeps the Windows mandatory byte-range locking behind the VFS and avoids locking the DB bytes that the commit path writes.

Local checks passed:
- make -C build -j8 libdoltlite.a doltlite
- bash test/run_doltlite_regression_case.sh gc_rewrite_failure
- cd build && bash ../test/doltlite_gc.sh
- cd build && bash ../test/doltlite_gc_session_state.sh
- cd build && bash ../test/doltlite_pragma_page_count.sh

The remaining direct Windows call is the atomic GC file replacement. SQLite VFS has open/delete/access/fullpath/lock/sync hooks, but no rename/replace hook, so MoveFileExW is still required there for atomic replace semantics.

timsehn:
Windows made it past basic file-backed SQL after the VFS sidecar lock change. The remaining failures were in the GC shell test:

- gc_single_file counted the new sidecar lock file as a data file. The test now ignores/removes *-lock.
- gc_preserve_dev used the slash branch filename form. I switched this GC assertion to the documented @dev form so this test stays focused on GC behavior; slash branch selection still has separate coverage.

Local checks passed again:
- bash test/run_doltlite_regression_case.sh gc_rewrite_failure
- cd build && bash ../test/doltlite_gc.sh
- cd build && bash ../test/doltlite_gc_session_state.sh
- cd build && bash ../test/doltlite_pragma_page_count.sh

timsehn:
Adjusted the GC smoke fix once more after the latest Windows run. @dev still returned the main-branch row count in this MSYS shell context, so the test was exercising filename branch parsing instead of GC.

The gc_preserve_dev assertion now uses SQL branch selection directly:

SELECT dolt_connect_branch('dev'); SELECT count(*) FROM t;

That still verifies the important GC property: the dev branch and its table contents survive the rewrite. Filename branch syntax remains covered by the connection-branch tests. Local doltlite_gc.sh passes.

timsehn:
Found the setup half of the same issue: doltlite_gc.sh was still creating the feat/dev branch rows through $DB/<branch>. On Windows that path form is not selecting the branch in this MSYS shell context, so the dev branch only had the main row.

I amended the test fix so doltlite_gc.sh no longer uses $DB/<branch> at all. It now selects branches with dolt_connect_branch() for setup and verification, keeping the test focused on GC preserving branch data.

timsehn:
The latest Windows failure was CRLF in the shell output, not a new GC mismatch. The log showed expected and actual both as:

0
2

but the string comparison still failed on Windows. I amended doltlite_gc.sh so run_test and run_test_match strip carriage returns before comparing. Local GC checks pass again.

timsehn:
Trimmed the Windows smoke job to the passing coverage for this PR:

- basic in-memory/file-backed SQL
- commit + dolt_gc sanity check
- doltlite_gc.sh
- doltlite_gc_session_state.sh
- doltlite_pragma_page_count.sh

I removed the regression harness invocation from the Windows job and dropped the remote-server test stubs I had started adding. The regression harness link issue is real Windows smoke expansion work, but it is outside this GC hardening PR.

timsehn:
Follow-up Windows smoke expansion is tracked in #1049. This PR now keeps Windows coverage limited to the checks already passing in CI, and leaves the regression harness / remote-server helper link cleanup for that follow-up.

timsehn:
Follow-up testing plan for #1049:

I think the right end state is exactly what Tim suggested: run the normal Test Suite build-and-test path on Windows, not a growing custom smoke script. The smoke job in this PR should stay small and prove the new Windows basics we need for GC: file-backed SQL, commit, dolt_gc, and the GC shell scripts.

For #1049, I would approach it in this order:

1. Add a Windows variant of the Test Suite build-and-test job using MSYS2/MINGW64, building the same targets as Linux where possible.
2. Run the standard test driver first, capture the failures, and split them into real Windows storage bugs vs test portability issues.
3. Fix test portability separately: CRLF normalization, sidecar lock cleanup, path separators, shell quoting, and branch filename syntax through MSYS.
4. For unsupported Windows surfaces, gate intentionally and visibly. The current remote-server regression harness link failure is one example: remotesrv is stubbed on Windows, but the all-in-one C regression harness still references remotesrv test helper symbols. That should either become a proper Windows skip in the harness or remotesrv should gain real Windows support.
5. Once the full suite is mostly green, keep this lightweight Smoke / windows job as a fast signal and add the full Windows Test Suite as the correctness gate.

The main thing I would avoid is continuing to grow this PR's smoke job into a second ad hoc test suite. This PR now has a passing Windows smoke signal; #1049 can make Windows first-class in the real suite.

PR #1170, opened 2026-06-04
ci: run CI on pull requests only, not master pushes

timsehn:
## What

Removes the `push: branches: [master]` trigger from the CI workflows so they run on **pull requests only**, not again on the post-merge master commit (redundant — the PR already gated it).

Changed: `benchmark`, `build-test`, `sanitizers`, `smoke`, `test` (each keeps `pull_request`; `build-test` keeps `workflow_dispatch`).

Unchanged: `release.yml` (still triggers on `v*` tags) and `sqlite-upstream-drift.yml` (scheduled).

This PR itself exercises the new triggers — the suites should run here on `pull_request` and not on master after merge.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,8 +1,6 @@
 name: Benchmark
 
 on:
-  push:
-    branches: [master]
   pull_request:
 
 permissions:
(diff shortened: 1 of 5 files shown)


timsehn:
Closing — we want to keep CI running on master pushes after all.

PR #1134, opened 2026-06-02
Skip date raw-page test under prolly storage

timsehn:
Fixes #1124.

## Summary
- Skip only the `date-14` raw `hexio_write` block when `doltlite_engine()` reports prolly storage.
- Remove the 257 stale `date-14` entries from `known_testfixture_divergences.txt`.

## Tests
- `LIBRARY_PATH=/opt/homebrew/lib make testfixture USE_AMALGAMATION=0`
- `./testfixture test/date.test`
- `mkdir -p tf-run && ln -sf ../testfixture tf-run/testfixture && cd tf-run && bash ../test/run_testfixture.sh date-1124-skip 60 date`
- `make lint`

Issue #1124, opened 2026-06-02
testfixture date: date/time builtin results diverge from stock


timsehn:
## Summary
`date` testfixture diverges on date/time builtin results (257 cases allow-listed). Not an obvious rowid/format class — possibly real function bugs.

## Sample
```
! date-14.1   expected: [2454629.5]   got: []
! date-14.2.0 expected: [1]           got: [0]
```
`julianday(...)` returns empty and a comparison returns 0 where 1 is expected.

## Repro
```
make testfixture USE_AMALGAMATION=0
./testfixture test/date.test
```

## Notes
Allow-listed in `test/known_testfixture_divergences.txt` (257 `date` entries). Confirm whether these are real `julianday`/date-function bugs or an intended divergence. Triage: #1119.

timsehn:
This is a known and intentional DoltLite testfixture divergence. The failing date-14 cases use `hexio_write` to mutate raw bytes at specific SQLite btree page offsets, then assert behavior based on SQLite's on-disk btree record layout. DoltLite's default storage is prolly, so those byte offsets do not represent the same data.\n\nThe underlying date/time functions are still covered by the rest of `test/date.test`; this particular block is really a raw SQLite storage-format test. Keeping these cases on the testfixture divergence allow-list is the right handling unless/until we decide to run upstream raw-page tests against a stock-btree test database.

PR #1186, opened 2026-06-04
Skip date raw-page tests under prolly storage

timsehn:
## Summary
- skip only the `date-14` raw `hexio_write` page-format block when `doltlite_engine()` reports `prolly`
- keep the test active for stock/orig storage paths where SQLite page offsets are meaningful
- remove the 257 stale `date-14.*` entries from the testfixture divergence allow-list

Follow-up to closed #1134 / #1124.

## Tests
- raw `date.test` from a fresh temp directory: `0 errors out of 1427 tests`
- `bash ../test/run_testfixture.sh "1134 date raw page skip" 120 date`
- `make lint`


timsehn:
Closing per maintainer preference: keep the date-14 raw page-format cases in the divergence allow-list rather than adding a test skip.

PR #1207, opened 2026-06-06
Fix rollback use-after-free; recover capi3 into the gated testfixture suite

timsehn:
## Summary
`capi3` was on `known_testfixture_crashes.txt` (whole-file skip). Triaging that skip recovered the file into the CI-gated suite **and uncovered a real heap-use-after-free** that the skip had been masking.

### Root cause of the skip
`capi3-7.x` pokes the sqlite page header at offsets 40/44 (`set_file_format`). Those bytes land inside doltlite's checksummed manifest, so doltlite rejects the corrupted file at `open()`. `sqlite3 db test.db` therefore throws, `db` is never created, and the top-level `db close` aborts the whole file before its summary line — which the harness counts as a crash.

### The masked bug (heap-use-after-free)
Once `capi3-7.x` is gated out and the file runs on, `capi3-11.9.2` (`ROLLBACK` with a live read cursor) trips a UAF under ASan:

```
ERROR: AddressSanitizer: heap-use-after-free in prollyMutMapClear
  prollyBtreeRollback -> prollyMutMapClear   (reads freed map)
freed by: prollyBtreeRollback -> restoreFromCommitted -> deserializeCatalog
          -> btreeFreeCatalogTables -> catFree
```

Cursor `pMutMap` pointers alias the catalog's per-table `pPending` maps. `restoreFromCommitted()` frees those maps, but the cursor aliases were cleared **afterward**. Fix: clear + detach the cursor aliases **before** `restoreFromCommitted()` frees the catalog.

## Changes
- **`src/prolly_btree.c`** — fix the UAF: clear/detach cursor pending-map aliases before the catalog is freed in `prollyBtreeRollback`.
- **`src/test_config.c`** — add a `doltlite` `ifcapable` token (gated by `#ifdef DOLTLITE_PROLLY`) so upstream `.test` files can fence off behavior with no prolly equivalent instead of silently forking.
- **`test/capi3.test`** — gate the raw file-format block (`capi3-7.x`) with `ifcapable {!doltlite}`.
- **`known_testfixture_crashes.txt`** — remove `capi3`, with a note on the resolution.
- **`known_testfixture_divergences.txt`** — record the 12 remaining C-API / open-timing / locking / catalog divergences, each with a reason.

## Result
- `capi3` goes from a whole-file crash-skip (0 gated tests) to **247 tests: 235 pass, 12 known divergences**, `run_testfixture.sh` exit 0.
- Verified under the ASan testfixture build. The gated rollback/transaction tests (`trans`, `savepoint`, `savepoint2/4`, `rollback`, `capi2`) still pass with the fix; the fix was confirmed against a clean master baseline (no regressions).

### DIFF
--- a/src/prolly_btree.c
+++ b/src/prolly_btree.c
@@ -5288,17 +5288,28 @@ static int prollyBtreeRollback(Btree *p, int tripCode, int writeOnly){
   if( p->inTrans==TRANS_WRITE ){
     assert( pBt->store.isMemory || pBt->store.graphLockFd >= 0 );
     assert( pBt->store.isMemory || pBt->store.lockDepth > 0 );
+    /* Cursor pending-edit maps alias the catalog's per-table pPending maps,
+    ** which restoreFromCommitted() is about to free. Clear and detach the
+    ** aliases first; touching them after the catalog is freed is a UAF. */
+    {
+      BtCursor *pC;
+      for(pC = pBt->pCursor; pC; pC = pC->pNext){
+        if( pC->pBtree==p && pC->pMutMap ){
+          prollyMutMapClear(pC->pMutMap);
+          pC->pMutMap = 0;
+          pC->mmActive = 0;
+          pC->mmPhysActive = 0;
+          pC->deferredTreeSeek = 0;
+          pC->mmIdx = -1;
+          pC->mmPhysIdx = -1;
+        }
+      }
+    }
     rc = restoreFromCommitted(p);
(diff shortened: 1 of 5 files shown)


timsehn:
Follow-up tracked in #1208: gate the remaining ~1,058 ungated inherited testfixture files in CI and triage their divergences. The `doltlite` ifcapable token added here is the mechanism for fencing off no-prolly-equivalent blocks.

Issue #1208, opened 2026-06-06
Gate all inherited SQLite testfixture files in CI and triage their divergences


timsehn:
## Problem
CI runs the testfixture suite through **5 hand-curated file lists** in `.github/workflows/build-test.yml` (Core / Large / Crash / Fuzz / Additional) — roughly **131 of the 1,189** inherited SQLite `.test` files. The other **~1,058 files are never executed by CI**, so any divergence or crash in them is invisible: "CI green" says nothing about them.

This is not hypothetical. Two findings while triaging the crash list:

- **capi3 masked a real heap-use-after-free.** It was whole-file-skipped on the crash list; gating the one inapplicable block recovered the file *and* exposed a UAF in `prollyBtreeRollback` (fixed in #1207). Whole-file skips can hide memory-safety bugs.
- **trans2 fails 29 `integrity_check`-mid-transaction assertions** (`trans2-N.4`) and is **ungated**, so it has gone unnoticed. Reproduces identically on clean master — not a regression, just never run. Needs triage: real bug vs. intentional doltlite divergence.

## Goal
Make CI exercise the **entire** inherited testfixture corpus, with every divergence/crash either fixed or explicitly recorded, so nothing is silently unrun.

## Work
1. **Enumerate** the ~1,058 ungated `.test` files (all of `test/*.test` minus the 5 CI lists).
2. **Run each** under the (ASan) testfixture via `run_testfixture.sh` and bucket the outcome:
   - passes clean,
   - diverges (record in `known_testfixture_divergences.txt` with a per-test reason),
   - crashes/hangs (record in `known_testfixture_crashes.txt`, but **triage under ASan first** — a crash can be a real memory-safety bug, as capi3 was; prefer gating the inapplicable block over a whole-file skip).
3. **Fix real bugs** found along the way (don't paper them over as divergences).
4. **Expand the CI lists** in `build-test.yml` to cover everything (likely a new job or a generated list), so the full corpus is gated going forward.
5. Start with **trans2** (`integrity_check` mid-transaction) as the first concrete triage.

## Notes / methodology
- Use `ifcapable {!doltlite}` (the token added in #1207) to fence off blocks that test stock on-disk-format / shared-cache / rollback-journal behavior with no prolly equivalent, rather than skipping whole files.
- Triage every assertion under ASan — do not dismiss a crash/divergence by eyeballing the marker; reproduce it.
[...]

PR #1210, opened 2026-06-06
ci: gate 625 inherited SQLite testfixture files that pass clean (#1208)

timsehn:
## Summary
Implements the first phase of #1208. CI previously ran testfixture through 5 hand-curated lists (~134 of 1,189 files); the other **~1,058 inherited SQLite `.test` files were never run by CI**, so divergences/crashes in them were invisible — which is exactly how `capi3` masked a real heap-use-after-free (#1207).

I swept all **1,057 ungated files** under the ASan testfixture and bucketed each. This PR adds a dedicated **`inherited-testfixture`** CI job that gates the **618** files that pass clean (or only with divergences already recorded in `known_testfixture_divergences.txt`).

## Sweep results (1,057 ungated files)
| Bucket | Count | Disposition |
|---|---|---|
| CLEAN / CLEAN-DIV | 627 | **gated here** (618 after holdbacks) |
| FAIL (new divergences) | 276 | triage → #1208 |
| CRASH | 66 | triage → #1208 |
| TIMEOUT / hang | 55 | triage → #1208 |
| CRASH-ASAN (memory bugs) | 33 | triage → #1208 (highest priority) |

## Validation
The clean files were re-validated **CI-faithfully** — a fresh `DOLTLITE_PROLLY_CHECK=1`, non-ASan testfixture built on plain master (no #1207) — and run through `run_testfixture.sh` (236,584 passing assertions, 0 unexpected failures).

**Excluded from the gate (9 files):**
- 2 clean only with the #1207 rollback fix (`alterfault`, `autoindex1`) — follow-up once #1207 merges.
- 7 nondeterministic-class (`thread003/004/2`, `walcrash4`, `writecrash`, `walfault2`, `walrofault`) — multi-threaded / crash- & fault-injection that flake run-to-run (same rationale as the existing `crash.test` exclusion; `thread003` passed CI but crashed in a local re-run).

The remaining **618 pass on plain master**, so this PR stands alone (no dependency on #1207).

## What's gated
A new `inherited-testfixture` job (90-min budget) builds testfixture and runs the 618 files via `run_testfixture.sh`, under the same strict-equality divergence contract as the existing lists.

Follow-ups under #1208: triage the FAIL/CRASH/TIMEOUT/CRASH-ASAN buckets (the 33 ASan crashes first — same bug class as #1207), and add the held-back files.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -484,3 +484,120 @@ jobs:
           tkt-80e031a00f tkt-80ba201079 tkt-31338dca7e \
           tkt-7bbfb7d442 tkt-4dd95f6943 tkt-2d1a5c67d \
           tkt-f3e5abed55
+
+  inherited-testfixture:
+    name: inherited-testfixture
+    runs-on: ubuntu-latest
+    timeout-minutes: 90
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y tcl-dev build-essential zlib1g-dev
+
+    - name: Configure
+      run: |
+        mkdir -p build && cd build
+        TCL_CONFIG=$(find /usr -path '/usr/share/miniconda/*' -prune -o -name tclConfig.sh -print 2>/dev/null | head -1)
+        if [ -n "$TCL_CONFIG" ]; then
+          ../configure --with-tcl=$(dirname $TCL_CONFIG)
+        else
+          ../configure
+        fi
+
+    - name: Build testfixture
+      run: |
+        cd build
+        make DOLTLITE_PROLLY_CHECK=1 testfixture
+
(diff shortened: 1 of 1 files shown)


timsehn:
Refined the gate from 625 → **618** files after a local sequential re-run of `run_testfixture.sh` (236,584 passing assertions, 0 unexpected failures) flagged **`thread003`** as an unexpected crash — even though it passed the CI job. `thread003` is a multi-threaded pcache stress test; that pass-in-CI / crash-locally split is exactly the run-to-run nondeterminism your `crash.test` exclusion already calls out.

Dropped the whole nondeterministic class rather than just `thread003`: `thread003`, `thread004`, `thread2` (multi-threaded), `walcrash4`, `writecrash` (crash-injection), `walfault2`, `walrofault` (WAL fault-injection — low value on doltlite's non-WAL storage anyway). The remaining 618 are deterministic clean passes. Amended + force-pushed; CI re-running.

PR #1298, opened 2026-06-09
Fix FTS3 malloc schema cleanup

timsehn:
## Summary
- preserve pending Doltlite catalog state across FTS3 OOM rollback paths
- keep the synthetic sqlite_master OOM guard at schema-table creation time only
- add a targeted Doltlite FTS3 OOM regression for post-failed-create sqlite_master readability

## Testing
- `./testfixture ../test/doltlite_fts3_oom.test`
- `./testfixture ../test/fts3malloc.test` now passes the old early `fts3_malloc-1.1` failure and reaches the next known failure at `fts3_malloc-1.5.transient.100000.319` (`no such table: main.sqlite_master`)

### DIFF
--- a/src/prepare.c
+++ b/src/prepare.c
@@ -416,6 +416,14 @@ int sqlite3InitOne(sqlite3 *db, int iDb, char **pzErrMsg, u32 mFlags){
     pDb = &db->aDb[iDb];
   }else
   if( rc==SQLITE_OK || ((db->flags&SQLITE_NoSchemaError) && rc!=SQLITE_NOMEM)){
+#ifdef DOLTLITE_PROLLY
+    if( pDb->pBt
+     && sqlite3FindTable(db, zSchemaTabName, pDb->zDbSName)==0
+    ){
+      rc = SQLITE_SCHEMA;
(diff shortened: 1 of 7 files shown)


timsehn:
Audit from the divergence-list side (this PR exposed that several entries I'd recorded as intentional were actually this bug — the master-root guard was silently dropping accepted writes to table 1):

Two more gated entries un-fail under this branch but are not yet removed from `known_testfixture_divergences.txt`, so the strict gate will fail with "FIXED — should be removed":

- `capi3 capi3-8.5` (capi3 is 9 failures here vs 10 on master; 8.5 passes, 8.3 still fails)
- `where where-25.1` and `where where-25.5` (where is 7 here vs 9 on master; 25.2 still fails)

The still-failing siblings (`capi3-8.3`, `where-25.2`) keep their entries, but their recorded reasons describe the old "structural catalog ignores raw sqlite_master edits" theory and should be re-derived from the actual remaining mechanism when this lands.

Verified the rest of the alter* family classifications against this branch: alter3/alter4 (schema_version catalog-hash + get_file_format) still diverge — those hold.

PR #1361, opened 2026-06-10
Honor max_page_count for DoltLite writes

### LINKED ISSUE #1359: sqllimits1.test hangs or times out under DoltLite
## Problem

`test/sqllimits1.test` appears to expose a DoltLite-specific hang or severe slowdown.

In local timing, the same inherited SQLite TCL test passes quickly on stock SQLite, but does not finish under DoltLite within the current CI per-file timeout.

## Local evidence

Stock SQLite testfixture:

```text
command: ./testfixture ../test/sqllimits1.test
result: 0 errors out of 3117 tests
wall time: 1.004s
```

DoltLite testfixture:

```text
command: ./testfixture ../test/sqllimits1.test
result: timed out at 300s
summary: <no summary>
```

This was measured while triaging #1357. The DoltLite timeout was observed with the newest available local DoltLite testfixture on the dev machine; CI / a clean Linux `DOLTLITE_PROLLY_CHECK=1` build should be used as the authoritative reproduction.

## Why this is likely a bug

This does not look like a legitimately slow upstream fixture: stock SQLite completes the full file in about one second. DoltLite failing to reach a summary within 300 seconds suggests a hang, pathological plan/code path, or limit-handling bug.

## Suggested next steps

- Run `sqllimits1.test` under DoltLite with progress output and identify the last subtest before the stall.
- Re-run the specific subtest against stock SQLite and DoltLite.
- If the behavior is an intentional DoltLite limit divergence, record the exact divergence and gate the file.
- Otherwise fix the DoltLite bug and add `sqllimits1` to the appropriate regression bucket.

Part of #1357.

timsehn:
## Summary
- store and return DoltLite's synthetic `max_page_count` instead of always reporting SQLite's default maximum
- fail DoltLite inserts with `SQLITE_FULL` once pending synthetic page growth exceeds the configured cap
- gate `sqllimits1.test`, leaving only the documented synthetic page-count display divergences

Fixes #1359.

## Verification
- `LIBRARY_PATH=/opt/homebrew/lib make DOLTLITE_PROLLY_CHECK=1 testfixture`
- isolated `run_testfixture.sh 'probe sqllimits1' 300 sqllimits1` -> `OK: sqllimits1 (3117 tests, 2 known divergences)`
- temp-dir `test/doltlite_recover_pragma_output.test` -> `0 errors out of 72 tests`
- regression bucket disjointness check -> `disjoint-ok`

Note: local build needs `LIBRARY_PATH=/opt/homebrew/lib` because this machine's Homebrew Tcl link flags include `-ltommath`.

### DIFF
--- a/src/prolly_btree.c
+++ b/src/prolly_btree.c
@@ -303,6 +303,7 @@ struct Btree {
   u8 bCatalogDropped;     /* OOM rollback dropped the catalog; reload before
                           ** treating an empty committedCatalogHash as a
                           ** fresh database */
+  Pgno mxPageCount;       /* Synthetic max_page_count limit */
 
   int nSavepoint;
   int nSavepointAlloc;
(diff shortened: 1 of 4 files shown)


timsehn:
Pushed a fix-up after CI caught two related failures.

The initial implementation double-counted pending row edits on top of staged chunk growth, which made `tkt2920-1.2` hit `SQLITE_FULL` too early. I changed the cap check to use the same synthetic chunk metric as `PRAGMA page_count`.

Also updated `doltlite_recover_rawformat_3` because its ported tkt2920 section still documented/expected `max_page_count` to be ignored; it now asserts clamp-to-current behavior and restores the cap before later cases.

Local verification after the fix-up:
- `sqllimits1`: `OK: sqllimits1 (3117 tests, 2 known divergences)`
- `doltlite_recover_rawformat_3.test`: `0 errors out of 102 tests`
- `ported` bucket: green, `1849` passing
- `ddl-schema` bucket: green, `40136` passing / `309` known divergences
- regression bucket disjointness: `disjoint-ok`

PR #1376, opened 2026-06-11
Fix swallowed OOM faults across the open/gc/cursor paths; gate malloc6

timsehn:
## Summary

`test/malloc.test` and `test/malloc6.test` were crash-listed as "giant OOM loop / nondeterministic". That was a misdiagnosis: with the runner fd limit pinned (`ulimit -Sn 1024`) they run in ~70s/6s, and their failures were real engine bugs — doltlite swallowed or mistranslated simulated allocation failures. This PR fixes every swallow/mistranslation site, makes **malloc6 CI-gated** (added to the fault-fuzz bucket, removed from the crash list), and corrects the crash-list reason for malloc (its remaining failures are structural test-infrastructure incompatibilities, not OOM handling).

## Fail-before / pass-after

Before (branch base), `got:` histograms:

`malloc.test` (aborts at the 1000-error cap after 13868 tests):
```
 144  [0 {}]                                   swallowed OOM (open path / VACUUM gc)
 129  [0 {1 2.3 4.5 hi there ...}]             swallowed OOM, full results (malloc-1)
  39  [1 {gc sweep phase failed}]              OOM -> generic error
  26  [1 {gc mark phase failed}]               OOM -> generic error
  10  [1 {failed to acquire lock for gc}]      OOM -> generic error
   5  [1 {attempt to write a readonly database}]  OOM -> silent read-only downgrade
   3  [0 {stuff 6 16}]                         swallowed OOM (malloc-2)
```
`malloc6.test`: `129 × [0 {}]` (open-path swallows), `5 × [1 {attempt to write a readonly database}]` — 134 errors / 1460 tests.

After:
- **malloc6.test: 0 errors out of 1460 tests** — gated (`OK: malloc6 (1460 tests)` in the fault-fuzz gate).
- malloc.test: every OOM-swallow/mistranslation class is gone. A harness-accurate fault-index sweep over the five failing statement shapes (fresh-open, existing-open, malloc-1 body, malloc-2 body, VACUUM-after-commit) reports **0 swallows and 0 mistranslations** across every reachable fault point. What remains in malloc.test is structural: malloc-14's prep does `forcecopy test2.db-journal` (doltlite never writes rollback journals) and `PRAGMA journal_mode` (intentional doltlite error), so its prep fails unconditionally on every one of its 2×50000 iterations and floods the 1000-error cap. The crash-list entry now states that real mechanism. (malloc-13 now exercises the crash harness correctly after the pathname fix below and passes apart from its tail iterations, where ATTACH of a crash-torn half-created chunk store legitimately fails.)

## Swallow sites found and fixed (each verified by gdb stack at the faulting allocation)

| # | Site | Class | Fix |
|---|------|-------|-----|
| 1 | `origBtreeIsSqliteFile` (src/btree_orig_api.c) — OOM in `sqlite3OsAccess`/`OsOpenMalloc`/`OsRead` probe returned "not a sqlite file"; the chunk-store open then succeeded | real swallow | new out-param API; NOMEM/IOERR_NOMEM propagate |
| 2 | `chunkStoreOpen` (src/chunk_store.c) — `sqlite3_mutex_alloc` result never checked | real swallow | fail open with NOMEM when `bCoreMutex` |
| 3 | `chunkStoreOpen` read-only fallback (src/chunk_store.c) — OOM on the read-write open silently downgraded the connection to read-only | mistranslation | propagate NOMEM/IOERR_NOMEM instead of retrying read-only |
| 4 | `sqlite3BtreeOpen` (src/prolly_btree.c) — `p->zBranch = sqlite3_mprintf(...)` unchecked | real swallow | NOMEM + full unwind |
| 5 | `registerDoltiteFunctions`/`doltliteRegister`/`doltliteMaybeSeedRepo` (src/prolly_btree.c, src/doltlite.c) — all registration + seeding failures discarded (`void`) | real swallow | chain returns rc; `sqlite3BtreeOpen` fails the open on NOMEM/IOERR_NOMEM only (re-registration mid-statement legitimately returns SQLITE_BUSY and stays best-effort) |
| 6 | `prollyBtCursorIndexMoveto` (src/prolly_btree.c) — failed `prollyCursorSeekBlob` fell through to the mut-map merge, which reset rc and reported "row not found" | real swallow | propagate rc |
| 7 | `csCommitToFile` (src/chunk_store.c) — deferred main-file open mapped IOERR_NOMEM to SQLITE_CANTOPEN ("unable to open database file") | mistranslation | pass through NOMEM/IOERR_NOMEM |
| 8 | `doltliteGcFunc` (src/doltlite_gc.c) — lock/mark/sweep OOM replaced with generic `sqlite3_result_error` text | mistranslation | `gcResultError` → `sqlite3_result_error_nomem` for OOM |
| 9 | `gcRewriteFile` tmp-file open (src/doltlite_gc.c) — OOM returned as SQLITE_CANTOPEN | mistranslation | return rc |
| 10 | `gcRewriteFile` reopen retry loop (src/doltlite_gc.c) — retry converted a transient OOM into success | real swallow | stop retrying on NOMEM/IOERR_NOMEM |
| 11 | `gcVerifySessionResolvable` (src/doltlite_gc.c, DOLTLITE_PROLLY_CHECK only) — diagnostic chunk walk whose failures are deliberately inconclusive | deliberate best-effort | benign-malloc bracket |
| 12 | `sqlite3RunVacuum` (src/vacuum.c) — failure to prepare `SELECT dolt_gc()` turned VACUUM into a silent no-op | real swallow | propagate OOM (non-OOM stays a no-op for builds without dolt_gc) |

Plus one real non-OOM bug found en route:

- **`csCanonicalFilename` discarded the `xFullPathname` result** except in the symlink case, so chunk stores opened/locked/synced under the caller-relative name instead of the canonical full path (stock SQLite always opens the canonicalized path). Exposed by the crash harness (`crashsql` never fired because its sync-name match uses full paths); fixed to always use the full pathname.

One regression was caught and fixed during verification: `*ppBtree` is `db->aDb[0].pBt`, and registration-time seeding resolves the chunk store through it, so it must be assigned before `registerDoltiteFunctions` — the corpus and gc shell suites caught the missing seed commit immediately.

## Gating changes

- `test/regression-buckets/fault-fuzz.txt`: + `malloc6`.
- `test/known_testfixture_crashes.txt`: `malloc6` removed; `malloc` reason corrected to the verified mechanism.
- `test/known_testfixture_divergences.txt`: 487 stale entries removed — the gate reported them as no-longer-failing because the swallows behind them are fixed (mallocG 8–188 open-path block, mallocI 17–57, shared_err-4 fault blocks, backup_malloc 63/74–78, all mallocD/mallocF entries) — and the fkey_malloc/vtab_err fault indices that shifted with the open-path allocation changes were updated (each verified deterministic over 3 consecutive runs).

## Verification

- Fault-fuzz gate, full bucket (74 files incl. malloc6), non-root `tester` with `ulimit -Sn 1024`, CI invocation: exit 0.
- C corpus: 309807/309807.
- Shell suites: doltlite_gc.sh 51/51, doltlite_vacuum.sh 10/10, doltlite_gc_session_state.sh 12/12; full `run_doltlite_tests.sh`: 66/73 — the 7 failing suites (parity/persistence/structural/chunk_physical_dups/deep_history/perf/count_perf) fail with byte-identical results on the unmodified branch base in the same container, i.e. pre-existing/environmental.
- Swallow scan: 0 swallows / 0 mistranslations across all five harness shapes.

Follow-up lead (not addressed here): in a diagnostic run with malloc-14 defused, malloc.test sections 31–33 (`locking_mode=exclusive`, two connections) flood "database is locked" and one run segfaulted at malloc-33.transient.109; it did not reproduce under gdb. Worth its own investigation.

### DIFF
--- a/src/btree_orig_api.c
+++ b/src/btree_orig_api.c
@@ -161,38 +161,47 @@ void origBtreeEnter(void *p){ orig_sqlite3BtreeEnter(B(p)); }
 void origBtreeLeave(void *p){ orig_sqlite3BtreeLeave(B(p)); }
 void *origBtreePager(void *p){ return orig_sqlite3BtreePager(B(p)); }
 
-int origBtreeIsSqliteFile(sqlite3_vfs *pVfs, const char *zFilename){
+/* Probe failures normally mean "not a legacy sqlite file" (the chunk-store
+** open reports the real problem), but an OOM must propagate so it is not
+** masked by a successful chunk-store open. */
+int origBtreeIsSqliteFile(sqlite3_vfs *pVfs, const char *zFilename,
+                          int *pIsSqliteFile){
   sqlite3_file *pFile = 0;
   int exists = 0;
   int outFlags = 0;
   int rc;
   u8 buf[16];
 
+  *pIsSqliteFile = 0;
   if( !zFilename || zFilename[0]=='\0'
    || strcmp(zFilename, ":memory:")==0 ){
-    return 0;
+    return SQLITE_OK;
   }
   if( !pVfs ){
     pVfs = sqlite3_vfs_find(0);
-    if( !pVfs ) return 0;
+    if( !pVfs ) return SQLITE_OK;
   }
 
   rc = sqlite3OsAccess(pVfs, zFilename, SQLITE_ACCESS_EXISTS, &exists);
-  if( rc!=SQLITE_OK || !exists ) return 0;
+  if( rc==SQLITE_NOMEM || rc==SQLITE_IOERR_NOMEM ) return rc;
+  if( rc!=SQLITE_OK || !exists ) return SQLITE_OK;
 
   rc = sqlite3OsOpenMalloc(pVfs, zFilename, &pFile,
                            SQLITE_OPEN_READONLY | SQLITE_OPEN_MAIN_DB,
                            &outFlags);
   if( rc!=SQLITE_OK ){
     if( pFile ) sqlite3OsCloseFree(pFile);
-    return 0;
+    if( rc==SQLITE_NOMEM || rc==SQLITE_IOERR_NOMEM ) return rc;
+    return SQLITE_OK;
   }
 
   rc = sqlite3OsRead(pFile, buf, sizeof(buf), 0);
   sqlite3OsCloseFree(pFile);
-  if( rc!=SQLITE_OK ) return 0;
+  if( rc==SQLITE_NOMEM || rc==SQLITE_IOERR_NOMEM ) return rc;
+  if( rc!=SQLITE_OK ) return SQLITE_OK;
 
-  return memcmp(buf, "SQLite format 3\000", 16)==0;
+  *pIsSqliteFile = memcmp(buf, "SQLite format 3\000", 16)==0;
+  return SQLITE_OK;
 }
 
 int origBtreeIntegrityCheck(
--- a/src/btree_orig_api.h
+++ b/src/btree_orig_api.h
@@ -89,6 +89,7 @@ int origBtreeIntegrityCheck(sqlite3 *db, void *p, Pgno *aRoot,
                             sqlite3_value *aCnt, int nRoot, int mxErr,
                             int *pnErr, char **pzOut);
 
-int origBtreeIsSqliteFile(sqlite3_vfs *pVfs, const char *zFilename);
+int origBtreeIsSqliteFile(sqlite3_vfs *pVfs, const char *zFilename,
+                          int *pIsSqliteFile);
 
 #endif
--- a/src/chunk_store.c
+++ b/src/chunk_store.c
@@ -169,17 +169,13 @@ static int csCanonicalFilename(
   if( !zFull ) return SQLITE_NOMEM;
 
   rc = sqlite3OsFullPathname(pVfs, zFilename, nPath, zFull);
-  if( rc==SQLITE_OK_SYMLINK ){
+  if( rc==SQLITE_OK || rc==SQLITE_OK_SYMLINK ){
     rc = chunkStoreDupFilenameDoubleNul(zFull, pzOut);
     sqlite3_free(zFull);
     return rc;
   }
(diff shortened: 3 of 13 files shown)


timsehn:
Pushed the CI-fallout fixes (commit e5b203ac6e). What the three failures turned out to be:

**The canonical-filename fix made the crash harness actually target the chunk store for the first time — and what it exercised found real recovery gaps**, now fixed:
- Commit batches start on a fresh sector when the device lacks powersafe-overwrite/atomic writes (a torn batch could previously destroy previously-synced bytes sharing its sector — stock WAL does the same via padToSectorBoundary)
- Root records are sealed (self-hash) and carry `DURABLE_TO`/`BATCH_START`/`NEXT_OFF`; replay classifies damage by writer-declared fact: mid-stream corruption (poisoned open) vs declared alignment gap (resume) vs torn crash tail (roll back). Legacy roots keep the old inference — existing stores read unchanged
- A crashed first commit (the one shape replay can't reach — header/chunks/root share one sync) now reopens as an empty database on positive evidence only; foreign files keep failing open
- **savepoint4** (core-sql): append cursor drifted from the physical tail across hot-journal copies — fixed by explicit logical-EOF accounting
- **windows**: canonical registration broke msys2 reopen-by-name; Windows keeps caller-name registration, unix keeps the canonicalization

Net effect: crash2/crash3/crash6 pass genuinely (81 more divergence entries removed, on top of the 487 from the OOM work); crash4 crash-listed with the verified fundamental reason (sector-trashed creation batch has no recovery anchor without a journal sidecar); fuzz3 leaves the sweep with documented justification (final-batch byte flips are in-band indistinguishable from a torn tail; coverage lives natively in doltlite_recover_corruption + corruption_test.c).

Verification on this exact tree: C corpus 309809/309809; core-sql 336,853 / storage 4,730 / fault-fuzz 269,775 / ported 2,520 — all 0 unexpected.

PR #1540, opened 2026-07-06
Refuse mid-commit reload that drops uncommitted-recent chunks (fixes dangling chunk -> gc mark-fail -> unreclaimable bloat)

Wldc4rd:
## Problem
Under concurrent commits to a shared DoltLite store, a committer can silently corrupt the store: `dolt_gc`'s mark phase fails on a missing chunk and the store then bloats unreclaimably (GC reclaims nothing). `PRAGMA integrity_check` fails durably (persists across reopen).

## Root cause
When a peer grows the file mid-commit, `csCommitToFile` (`chunk_store.c`) calls `csReloadFromDiskPreservingLocalRefs` → `csReloadFromDisk` → `csAdoptOpenedStoreState` (`chunk_wal.c`), which frees `staging.aRecent` — **including this connection's uncommitted-recent chunks** (drained to the file but not yet manifest-committed). The connection's refs are preserved across the reload, so the committed refs then reference chunks that were dropped from the index/staging → **dangling chunk** → GC mark-phase failure → monotonic bloat.

Those dropped chunks' bytes live *beyond the committed EOF* (the "crash garbage beyond EOF" region a concurrent committer overwrites/truncates), so they cannot be safely re-sourced and replayed after the reload.

## Fix (`chunk_store.c`, in `csReloadFromDiskPreservingLocalRefs`)
```c
if( cs->staging.nRecentUncommitted > 0 ){
    return SQLITE_BUSY_SNAPSHOT;
}
```
Refuse the corrupting in-place reload and surface a **retriable conflict**. The caller rolls back the uncommitted staging (`chunkStoreRollback` already cleanly discards uncommitted-recent + truncates to `iUncommittedStart`) and retries against the refreshed state — the same conflict-and-retry contract as a head-CAS conflict. Correctness-first: no path silently corrupts.

## Testing
Deterministic single-process reproduction added (a `SQLITE_TEST`-gated one-shot reload injection + the existing `DOLTLITE_CHUNK_PENDING_DRAIN_LIMIT` override to force uncommitted-recent chunks). A `DOLTLITE_MECH_REPRO` build flag enables the harness without the TCL testfixture toolchain.
- **Unpatched:** `integrity_check` FAILS, `dolt_gc` mark-fail, dangling chunk (durable across reopen).
- **Patched:** COMMIT → `SQLITE_BUSY` (261 / BUSY_SNAPSHOT), clean rollback, `integrity_check` ok, `dolt_gc` ok, no dangling.
- **Patched + retry-on-busy:** attempt 2 commits cleanly, row count intact — no data loss.
- **Regression:** all gating C tests pass — concurrent_stress 18, multi_process_gc 86, corruption 70, invariant 838, sequence_reload 10 = **1022/0**.

## Caller contract (note for integrators)
`SQLITE_BUSY_SNAPSHOT` is intentionally **not** busy-handler-retriable (per SQLite semantics the snapshot changed; the transaction must be rolled back and retried by the application). Integrators must roll back + retry the transaction on this code from commit — as this fix's own reproduction does.

## Alternative considered
Transparent preserve-and-replay (retain the uncommitted chunk bytes in memory across the reload, re-append at the new EOF) would avoid the retry but is complex and unsafe for drained chunks (bytes are freed from `pWriteBuf`, and the file region is overwritable by the peer). Conflict-and-retry is the correct, minimal first landing.

### DIFF
--- a/src/chunk_store.c
+++ b/src/chunk_store.c
@@ -2168,8 +2168,22 @@ static int csReloadFromDisk(ChunkStore *cs){
   ChunkStore tmp;
   ChunkStoreReloadState saved;
   char *zOldFilename;
-  int rc = chunkStoreOpen(&tmp, cs->file.pVfs, cs->file.zFilename,
-                          SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE | SQLITE_OPEN_MAIN_DB);
+  int rc;
+  /* Guard EVERY reload path at the shared low level -- not only the commit-path
+  ** wrapper csReloadFromDiskPreservingLocalRefs. The external-change refresh
+  ** (chunkStoreRefreshIfChanged) calls csReloadFromDisk DIRECTLY, so a read/op
+  ** that picks up a peer's commit while THIS connection holds uncommitted-recent
+  ** chunks (drained to file at csCommitToFile, not yet manifest-committed) would
+  ** reach csAdoptOpenedStoreState below, which frees staging.aRecent -- dropping
+  ** those chunks while cs->refs still reference them => dangling chunk =>
+  ** dolt_gc mark-phase fails + unreclaimable bloat. Refuse with a retriable
+  ** conflict; the caller rolls the transaction back and retries against the
+  ** refreshed state (same contract as a head-CAS conflict). */
+  if( cs->staging.nRecentUncommitted > 0 ){
+    return SQLITE_BUSY_SNAPSHOT;
+  }
+  rc = chunkStoreOpen(&tmp, cs->file.pVfs, cs->file.zFilename,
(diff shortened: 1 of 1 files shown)


timsehn:
Looking at this now.

timsehn:
I'm going to need a test that fails before this fix and now passes. I can have an agent do it. And I'll sweep any adjacent test surface for more latent bugs.

Wldc4rd:
> Looking at this now.

Thanks Tim!

This definitely solved one layer of the problem but it seems there might be another layer to it too. I'm running low on tokens though so I'm not sure when I can get the next layer traced/PR'd.

timsehn:
Superseded by #1545

PR #1546, opened 2026-07-06
Make dolt_gc force a fresh exclusive view

timsehn:
## Summary
- make `dolt_gc()` reject calls from inside an active transaction or reentrant graph lock
- force a store refresh after acquiring the GC graph lock so GC sees peer commits that may have been missed by WAL reuse
- add TCL coverage for in-transaction GC rejection and post-lock refresh before GC rewrite

## Tests
- `docker run --rm -v "$PWD":/work -v /tmp/doltlite-gc-stop-build:/build -w /build ubuntu:24.04 bash -lc 'apt-get update >/dev/null && apt-get install -y tcl zlib1g >/dev/null && ./testfixture /work/test/doltlite_gc_stop_world.test && ./testfixture /work/test/doltlite_reload_uncommitted_recent.test && ./testfixture /work/test/doltlite_chunk_pending_drain.test'`
- previously on the same patch before cherry-pick: `DOLTLITE_BUILD_DIR=/build bash test/run_doltlite_regression_case.sh all` (`309828 passed, 0 failed`)

### DIFF
--- a/src/doltlite_gc.c
+++ b/src/doltlite_gc.c
@@ -665,7 +665,9 @@ static int gcRun(
   ChunkStore *cs,
   int *pnKept,
   int *pnRemoved,
-  const char **pzPhase
+  const char **pzPhase,
+  int bRequireExclusive,
+  int bForceRefresh
 ){
   ProllyHashSet marked;
   int rc;
@@ -673,10 +675,23 @@ static int gcRun(
   *pnKept = 0;
   *pnRemoved = 0;
 
+  if( bRequireExclusive && (!sqlite3_get_autocommit(db) || cs->lockDepth>0) ){
+    *pzPhase = "gc requires exclusive access";
+    return SQLITE_BUSY;
+  }
+
   rc = chunkStoreLockAndRefresh(cs);
   if( rc!=SQLITE_OK ){
     *pzPhase = "failed to acquire lock for gc";
     return rc;
+  }
+  if( bForceRefresh ){
+    rc = chunkStoreForceRefresh(cs);
+    if( rc!=SQLITE_OK ){
+      chunkStoreUnlock(cs);
+      *pzPhase = "failed to refresh store for gc";
+      return rc;
+    }
   }
 
   rc = prollyHashSetInit(&marked, chunkIndexCount(&cs->index) > 64 ? chunkIndexCount(&cs->index) : 64);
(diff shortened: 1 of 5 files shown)


timsehn:
Updated this PR after looking at #1547.

The additional suspected layer is internal compaction, not the public `dolt_gc()` entry point: `PRAGMA wal_checkpoint` can call `doltliteGcCompact()` while the same connection is inside a write transaction and has drained chunks in `aRecent` that are not committed yet. That bypassed the public `dolt_gc()` transaction guard and could sweep/reset recent chunks before the transaction's refs were durable.

The branch now makes internal compaction no-op while the connection is in a transaction, re-entering the graph lock, or holding uncommitted recent chunks. Public `dolt_gc()` still errors in those cases.

Added coverage in `doltlite_gc_stop_world.test` that forces pending chunks to drain, runs `PRAGMA wal_checkpoint` inside the transaction, commits, reopens, runs integrity check and GC, and verifies integrity again.

Wldc4rd:
@timsehn this looks like it may address the second-layer issue I filed in #1547 (the residual concurrent-commit corruption that persists after the merged #1545 reload guards). Sharing our production read in case it's useful — we have not test-pulled #1546 yet, but the mechanism match is strong:

**Our RC#2 symptom** (#1547): on a high-concurrency store, after a clean state the store re-corrupts within minutes–~30min → `dolt_gc()` → `gc mark phase failed` (zero reclaim) → unbounded bloat. Corruption is real (integrity_check fails on a **static copy**, not just a live view), and a low-concurrency store on the identical binary stays clean.

**Why #1546 matches:**
- The `staging.nRecentUncommitted>0` early-return in the GC path is the same `nRecentUncommitted` guard concept as #1545 — extended to `dolt_gc`. That's exactly the vulnerable window in our mechanism (uncommitted-recent chunks in flight).
- "force a store refresh after the GC graph lock so GC sees peer commits missed by WAL reuse" gives a concrete cause for a **persisted** dangling chunk: if `dolt_gc` marks/rewrites against a stale view that missed a peer commit, it can drop that peer's recent chunk from the rewritten store → the next `dolt_gc` mark-fails on it. That fits our static-copy corruption (the drop is persisted by GC, not a transient read).

So our current read: **#1546 likely prevents the RC#2 trigger** (a stale-view `dolt_gc` dropping uncommitted-recent / peer-committed chunks). One nuance we'll be careful about on our side: it should *prevent* recurrence on a clean store but won't *recover* an already-corrupt store (the chunk is already gone) — so we'll validate by deploying on a fresh store and watching for no-recurrence + sustained reclaim over a multi-hour window, not by running it against a corrupt copy.

Question: is #1546 intended to cover the #1547 second layer, or is that residual path separate in your view? If it lands, we'll test-pull and report whether it ends the recurrence on our workload (we're running a rebuild-hold mitigation we'd love to retire). Happy to share a static corrupt-store copy + the full GC-cadence timeline. — mechanic @ thriva-dev (@Wldc4rd)

Issue #1547, opened 2026-07-06
Concurrent commits still drop uncommitted-recent chunks after #1545 (second layer) → dangling chunk → gc mark-fail → unreclaimable bloat


Wldc4rd:
## Summary

Follow-up to #1545 (thanks @timsehn for picking it up + adding the regression test). The merged reload guards are the **first layer** and measurably help, but on a **high-concurrency store** the corruption still recurs — there is a **second layer**: some path still drops uncommitted-recent chunks under concurrent commits, producing a dangling chunk → `dolt_gc` **mark-phase failure** → the store can no longer be GC-reclaimed → unbounded write-amp bloat.

This is the "another layer" flagged on #1540. We're running a rebuild-hold mitigation in production; we'd love to retire it once the second layer is fixed.

## Symptom

On a DoltLite store under sustained concurrent commit load:
- `PRAGMA quick_check` / `PRAGMA integrity_check` → `*** in database main *** integrity check failed`
- `SELECT dolt_gc()` → `gc mark phase failed` (mark phase cannot complete → **zero reclaim**)
- The store then **bloats monotonically** (write-amp is never reclaimed) → eventually reconcile/query cycles slow (soft-degrade).

Confirmed **real** (not a live-read artifact): integrity fails the same way on a **static copy** of the file.

## Reproduction (honest framing)

- **Reliably reproduces on a live high-concurrency store** (dozens of concurrent single-connection writers — a CLI-per-op architecture — committing to one shared store). After any clean state (fresh rebuild), the store flips from `integ=ok` to `integ=fail` within **minutes to ~30 min**, then stays corrupt.
- A **low-concurrency** store on the identical binary stays clean and GCs fine → the trigger tracks **write concurrency**, specifically concurrent commit + reload/refresh interleaving.
- We do **not** yet have a minimal isolated repro. The **deterministic reload-injection hook you kept from #1545 is very likely the lever** to build one: drive a committer that holds uncommitted-recent chunks, inject a reload/refresh from a peer commit at that window, and check for a referenced-but-missing chunk after.

## Evidence pattern (10-min GC cadence log)

```
clean:  integ=[ok]   gc=[24766 chunks removed, 55555 kept]  308M -> 243M   (reclaims)
corrupt: integ=[FAIL] gc=[gc mark phase failed]             308M -> 308M   (0 reclaim, then bloats)
```

[...]

PR #1556, opened 2026-07-06
Add GC dangling chunk diagnostics

timsehn:
## Summary
- add missing-hash context to `dolt_gc()` mark-phase failures, including the missing chunk hash, parent hash when known, traversal source, and rc
- add test-build chunk-store counters for reload-guard hits and pending WAL drains via `dolt_debug_chunk_store()`
- add a bounded concurrent CLI commit + GC stress test and run it in Ubuntu CI

## Notes
This is an instrumentation/stress-gate PR for #1547, not a claimed full fix. The goal is to make the next persisted dangling-chunk report point at the exact missing hash and traversal edge, while also increasing coverage around concurrent commits with forced pending drains.

## Testing
- `make doltlite`
- `bash test/doltlite_gc.sh ./doltlite`
- `bash test/gc_concurrent_commit_stress.sh ./doltlite`
- `./doltlite :memory: "SELECT dolt_gc();"`

`make testfixture` compiled through the amalgamation/test sources but failed at local link because this machine is missing `libtommath` (`ld: library 'tommath' not found`).

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -337,6 +337,11 @@ jobs:
       run: cd build && bash ../test/remote_test.sh ./doltlite
       timeout-minutes: 5
 
+    - name: GC concurrent commit stress
+      if: matrix.platform == 'ubuntu'
+      run: cd build && bash ../test/gc_concurrent_commit_stress.sh ./doltlite
+      timeout-minutes: 5
+
     # -- C/Python/Go integration tests -----------------------
 
     - name: C integration tests
--- a/src/chunk_store.c
+++ b/src/chunk_store.c
@@ -11,6 +11,13 @@
 #include <stdlib.h>
 #include <limits.h>
 
+#if defined(SQLITE_TEST) || defined(DOLTLITE_MECH_REPRO)
+static sqlite3_int64 csDebugReloadGuardPreserve = 0;
+static sqlite3_int64 csDebugReloadGuardDirect = 0;
+static sqlite3_int64 csDebugPendingDrains = 0;
+static sqlite3_int64 csDebugPendingDrainChunks = 0;
+#endif
+
 static int csFileLockHeld(sqlite3_file *pFile){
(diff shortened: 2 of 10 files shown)


timsehn:
Added another diagnostic commit: `2a0ff12f5a` (Add prolly parent publish diagnostics). In `DOLTLITE_PROLLY_CHECK=1` builds, doltlite now checks every internal prolly node immediately before publishing it. If a parent is about to be stored with a missing child, it fails at the writer and prints: context, parent hash, child hash, parent level, parent item count, and child index. This should tell us whether the dangling edge is introduced by chunker output, ancestor-spine rewrite, append fast path, or replace-batch no-rechunk code, instead of only discovering it later during GC mark.

timsehn:
Update after inspecting `~/Downloads/corrupt-static-copy.db` (sha256 `c7a96425ac1b8cc188ad22fe113c278bdbf123684d99454fa36ca8fee81d1629`): the reported missing chunk `010a8957...` was a false GC traversal edge, not an actual missing child. The parent `2b5f0fcf...` is a 102-byte v2 commit, but `doltliteClassifyChunk()` checked the v2 working-set shape before commit shape, so GC interpreted bytes at offset 1 as a working-set catalog hash. Decoding the same bytes as a commit shows the real parent commit hash is `0a8957ec809ff2614a5659d40f327ec989fc999d`, and that chunk is present. Added commit `78c18f9f97` to classify exact commit chunks before working-set chunks and added a regression for this exact 102-byte commit shape. Verified on a fresh copy of the downloaded DB: `PRAGMA integrity_check` -> `ok`, `SELECT dolt_gc()` -> `21978 chunks removed, 51494 chunks kept`, then `PRAGMA integrity_check` -> `ok`.

timsehn:
Closing this diagnostics PR. The downloaded artifact identified a concrete classifier bug, so I split the real fix and regression into the smaller bug PR #1559.

PR #1606, opened 2026-07-13
Harden dolt_at historical table oracle coverage

timsehn:
Stacked on #1605.

Adds VC oracle coverage for adjacent dolt_at_<table> cases after reopening on a branch where the table/schema is not current:
- non-INTEGER primary key table
- composite primary key table
- extra non-PK columns including NULLs
- same table name with schema changed on another branch, reading common columns
- quoted table name

These cases compare actual row data against Dolt, not just table existence.

Verified:
- test/vc_oracle_at_test.sh build/doltlite /opt/homebrew/bin/dolt (26 passed)
- bash test/doltlite_at.sh build/doltlite (46 passed)
- git diff --check

### DIFF
--- a/test/vc_oracle_at_test.sh
+++ b/test/vc_oracle_at_test.sh
@@ -79,13 +79,19 @@ oracle_error() {
 
 oracle_table_after_reopen() {
   local name="$1" setup="$2" table="$3" ref="$4"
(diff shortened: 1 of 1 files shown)


timsehn:
Folded into #1605.

Issue #1613, opened 2026-07-14
Rebase precondition falsely reports 'uncommitted changes' after checkout under concurrent checkout churn


timsehn:
## Symptom

`multi_process_merge_rebase_test`'s Test 4 (`rebase_racing_checkout`) intermittently fails `rebase_vs_checkout_rebase_ok`. Diagnostic from a failing CI run:

```
[diag] rebase exhausted; last result: STEP_ERR: cannot start a rebase with uncommitted changes
```

`doRebaseOnce` opens a **fresh connection each round**, runs `dolt_checkout('feat')`, then `dolt_rebase('-i','main')`, retrying up to 120 times. All 120 rounds failed with the same "uncommitted changes" error while a sibling process looped `dolt_checkout('main')`/`dolt_checkout('side')`.

## Why it matters

- The failure is **sticky**: 120 independent fresh connections each check out `feat` and each see it as dirty. A transient race would clear within a few rounds (the checkout child is finite); instead the on-disk state for `feat` stays "dirty" for the whole run.
- The checkout child **never touches `feat`** (only main/side), yet `feat` ends up reported dirty — cross-branch contamination through the shared refs/working-set state.
- `feat` was clean when committed; nothing changed its data.

So this is a **false-positive dirty detection** persisted to disk: after the race, `feat`'s stored working-set-catalog hash no longer equals its head commit's catalog hash, and `doltliteHasUncommittedChanges` (doltlite.c:199 — it re-serializes the working catalog and hash-compares to head) reports uncommitted changes on every subsequent checkout.

## Likely root cause: #1596

This is almost certainly the live-vs-canonical catalog-serialization issue from #1596. `doltliteHasUncommittedChanges` compares a **re-serialized working catalog hash** against the committed head catalog hash. If a checkout under contention loads/persists the catalog in non-canonical order (exactly what #1596 documents), the re-serialized hash differs from the canonical committed hash even though the data is identical — a false "dirty" that reproduces deterministically on every reload (hence sticky). The concurrent checkout churn is only the trigger that makes `feat`'s working-set-catalog get written/loaded in the non-canonical order.

## Reproduction

Rare. Not reproduced locally across ~250 attempts: 25 loops of the real test under 4x `yes` CPU load, plus 200 foreground `checkout feat`+rebase iterations against 3 concurrent main/side flipper processes. Observed once in CI. Repro likely needs the specific disk/lock timing of the CI runner.

## Fix direction

[...]

PR #1616, opened 2026-07-14
Adopt the canonical catalog at schema commits (#1596)

### LINKED ISSUE #1596: sqlite_master rowid/rootpage bindings change across reload — a rolled-back DDL reorders the live catalog view
Found while attempting to ungate the fts crash-list files (e_fts3, fts3malloc, the fts3 suite-runner): all three abort for the same root cause.

## Symptom

The rowid and row-order of `sqlite_master` differ between a live connection and the reloaded catalog. Trivial repro, no fault injection:

```tcl
sqlite3 db test.db
db eval {CREATE VIRTUAL TABLE data USING fts3()}
db eval {SELECT rowid,name,rootpage FROM sqlite_master ORDER BY rowid}
# live:     1 data 0 | 2 data_content 3 | 3 data_segments 5 | 4 data_segdir 4
db close; sqlite3 db test.db
db eval {SELECT rowid,name,rootpage FROM sqlite_master ORDER BY rowid}
# reopened: 1 data 0 | 2 data_content 3 | 3 data_segdir 4  | 4 data_segments 5
```

Worse: a **rolled-back statement** triggers the same flip mid-connection, because rollback restores the canonical (serialized) form. Verified under OOM injection: `CREATE VIRTUAL TABLE pages` fails cleanly with OOM, `pages` is correctly absent — but the pre-existing `data_segdir`/`data_segments` rows have swapped rowids. Stock guarantees a rolled-back statement leaves `sqlite_master` byte-identical.

## Root cause

The commit-form catalog serializer deliberately canonicalizes — the comment above `captureSavepointCatalogSnapshot` says it plainly: *"The commit-form catalog serializer canonicalizes (sorts rows and renumbers roots), so a round-trip through it reassigns iTable numbers."* That canonicalization is load-bearing for the Dolt model (history-independent catalogs), but the live connection keeps its pre-canonical ordering, so live view ≠ persisted view until any reload.

## Why it matters / what it blocks

`malloc_common.tcl`'s `do_write_test` retry loop decides "the statement actually applied" by checking whether `md5sum(rowid, ...)` over the target table changed across a failed iteration — valid on stock. On doltlite, the first OOM iteration that engages a real rollback flips catalog rowids, the checksum changes, the harness stops retrying, and the test scaffold proceeds believing a CREATE/DROP applied when it didn't. The phantom state cascades:

- `e_fts3` aborts at 1.4.2.2 ("no such table: docs_content") in its OOM passes
- `fts3malloc`'s "OOM setup can leave the ft table absent" is this, not a test footgun
- the `fts3` suite-runner aborts in its inlined fts3_malloc portion

All three stay crash-listed until this is addressed.

## Fix direction (needs design)

Either the live catalog adopts canonical order eagerly at commit (requires renumbering live iTable references — compiled statements hold root numbers, so this likely implies statement expiry at commit-of-DDL), or the serializer preserves connection order (conflicts with catalog content-addressing). Not a quick patch; the savepoint machinery already works around the same scrambling (verbatim snapshots).

timsehn:
Fixes #1596.

## Design (final — evolved during review rounds)

#1596's disease was never the canonical serialization form — it was that the live session didn't **adopt** it. The old commit path rebuilt a session-order master root (`buildRuntimeMasterRoot`) and patched it over the reloaded canonical catalog, so live and persisted disagreed until an *uncontrolled* reload (reopen, rollback, cross-connection refresh) scrambled `sqlite_master` rowid bindings mid-connection. That instability was the fts-trio blocker, the `do_write_test` phantom-DDL class, and the #1613 sticky-dirty class.

**The fix:** schema-changing commits now adopt the canonical catalog **wholesale, including its master root**. Live ≡ persisted at every moment; reloads are byte-level no-ops. The canonical form itself (positional numbering over type-rank/name-sorted rows, canonicalized SQL text) stays exactly as the history-independence contract requires — `history_independence_test.sh` passes 1053/1053, including the DDL families (inline-CREATE vs ADD COLUMN vs create-copy-rename hash identically).

> An earlier iteration of this branch tried verbatim (allocation-order) numbering; the crash-fuzz history-independence gate correctly rejected it. The valuable byproducts survived: number-first serializer matching (cached entry names go stale across RENAME — name-first matching **corrupted renamed tables**: "database disk image is malformed" + dropped from ranged diffs), stock schema_version cookie semantics in-session, and the OOM fixes below.

## Also fixed along the way

- **Canonicalizer bugs that bricked reparse** (latent on master, exposed by adoption): reserved-word object names (a table named `ON`) emitted unquoted; `-- comment` + whitespace-collapse swallowed the rest of the statement. Keywords now quote, comments strip.
- **Swallowed OOM** in table-root classification (cached a wrong verdict permanently) and in the clean-close marker (now benign-marked) — complementing #1617; `e_fts3` runs 107k assertions to zero errors, `fts3malloc` 12k to zero, both off the crash list. `fts3.test` (permutations suite-runner, pure double coverage with load-sensitive OOM sections) leaves the fts bucket.
- **dbstatus** moves from crash list to divergence entries (completes now; heap-shape probes are structurally different on prolly).

## Validation

- `history_independence_test.sh` 1053/1053 · determinism gate 71/71 · 17/17 gating c-tests · 87/87 shell suites · `doltlite_fts3_oom` 45/45 · diff/summary oracle 49/49 vs real Dolt · `tkt-c694113d5` passes
- New gating suite asserts the invariant directly: sqlite_master identical in-session / after dolt_commit / after reopen, plus order-independent catalog hashes
- Divergence list reconciled per-file with exact-match validation (the canonical-adoption cost is visible and bounded: in-session row order/oid bindings flip to canonical at schema commits — the alter/attach order-class entries)


timsehn:
CI round 1 findings are all resolved; four more commits pushed:

- **Oracle failure was a real corruption bug**: the serializer matched entries by cached `TableEntry.zName`, which goes stale across `ALTER TABLE RENAME` — rename `t`→`u` then `CREATE TABLE t` + commit made `u` read as *"database disk image is malformed"* and dropped it from ranged diffs. Entries now match by number (the identity under verbatim numbering) with name only as fallback. `vc_oracle_diff_stat_test` is 49/49 vs real Dolt.
- **Two OOM-robustness fixes** (unmasked, not caused, by this PR — A/B'd on master): table-root classification swallowed and *cached* an OOM-induced misclassification (now propagates `SQLITE_NOMEM`, uncached); the clean-close marker's best-effort allocations are benign-marked. `e_fts3` runs 120,772 assertions to zero errors, `fts3malloc` 13,738 to zero (its master-count assertion ifcapable-gated: PK-clustered segdir has no autoindex row — the alter-1.x class).
- **Lists**: 34 fixed divergence entries removed; dbstatus moved from crash list to 97 divergence entries (ubuntu IDs); `attach3-3.2` added (raw-format rootpage); `e_fts3`/`fts3malloc` off the crash list entirely; `fts3` (permutations suite-runner, pure double coverage with load-sensitive OOM sections) excluded from the fts bucket alongside the other nondeterministic-class files.
- **Ported test updates**: expectations that had codified the old eager commit-time schema reset now assert stock behavior (cached-EXPLAIN replay semantics verified against stock sqlite 3.53.2; schema_version writes stick in-session; index_list order now matches upstream pragma.test's expected values exactly).

Local: c-tests 16/16, determinism gate 71/71, shell suites 87/87, and the 15 divergence-trimmed files match their remaining entries exactly.

timsehn:
CI is fully green (26/26). Summary of what the review rounds surfaced beyond the core design:

- **`commit -am` silently corrupted commits** when drops+creates had shifted canonical positions: the staged catalog overlaid working-numbered entries onto HEAD's unnamed (name-overlay-skipped) master root, and number-first row pairing then mapped entries onto the wrong rows — a valid-looking commit missing the first N tracked tables. Caught by the many-tables gating test; fixed by pairing the master entry by table number and adopting the working root.
- **Three canonicalizer bugs** that bricked schema reparse, all latent on master and exposed the moment the live session started adopting canonical text: unquoted reserved-word names (`CREATE TABLE ON` on reload), `--`-comment-swallowed statement tails from whitespace collapse, and single-quoted names (`'p 1 "parent one"'`) skipped mid-identifier.
- **Two OOM-robustness fixes** (benign-marked close marker; propagated + uncached table-root classification), letting e_fts3/fts3malloc run 120k/12k assertions clean and come off the crash list.
- Lists reconciled per-file with exact-match validation; `corrupt.test` joins corrupt4 on the crash list (byte-flips land on run-varying content); dbstatus returns to its master-era crash entry; the dot-command oracle compares listings order-insensitively and normalizes canonical-text spacing.

Gates: history-independence 1053/1053, determinism 71/71, c-tests 17/17, suites 87/87, regression 309,830/309,830 (checked build), diff oracle 49/49 and dot-commands 61/61 vs clean stock.

timsehn:
Final addition, CI green (26/26): a catalog-constructor sweep after the `-am` fix found the same mixed-domain hazard in `dolt_reset --hard`'s untracked-table preservation, plus two pre-existing defects verified on master — a hard reset never restored tables dropped in the working tree (leaving `no such table` + a phantom `deleted` status) and corrupted a surviving table's index. Fixed by rebuilding the preserved catalog from the target with untracked tables and their indexes appended under fresh numbers; serializer hardening (name-verified number matching for constructed arrays, opt-in foreign-domain serialization) defuses the remaining latent sites. Five new gating cases in `doltlite_reset.sh` all fail on master. Full battery green: 17/17 c-tests, 309,830/309,830 checked regression, 1053/1053 history gate, 87/87 suites, 49/49 diff oracle.

Issue #1630, opened 2026-07-15
Per-commit cost grows linearly with WAL size: RefreshAndConfirmHead does a full store reopen and WAL replay


timsehn:
Found while probing large-database scaling (production build, no DOLTLITE_PROLLY_CHECK).

## Symptom

Per-commit cost grows linearly with commits since the last gc, making cumulative history cost quadratic. 10k-row table, one-row-UPDATE commits:

| history depth | ms/commit |
|---|---|
| 500 | 16.5 |
| 1000 | 47.9 |
| 1500 | 79.1 |
| 2000 | 109.6 |
| 2000 (fresh session) | 134.3 |
| 2000 (after dolt_gc) | 12.9 |

The same mechanism makes open ~O(WAL): a freshly bulk-loaded 1M-row store opens in ~105ms (everything still in the WAL region) vs 6ms after gc (compacted index is mmap'd).

## Root cause (profiled)

Every `dolt_commit` calls `doltliteRefreshAndConfirmHead` twice (the head re-confirm required for multi-writer correctness). That calls `chunkStoreForceRefresh` → `csReloadFromDisk` → a **full `chunkStoreOpen`**, whose `csReplayWal` preads and blake3-verifies the entire WAL region. The WAL only folds into the compacted index during user-driven `dolt_gc`, so it grows without bound between gcs and every commit re-replays all of it.

## Fix direction

gc stays user-driven (no auto-fold — gc is costly and deliberate). The fix is to make the refresh incremental:

1. When the on-disk manifest matches the in-memory state (single-writer common case), reuse the current in-memory index/refs instead of reopening — the re-confirm semantics only need to detect that nothing changed.
2. When the file grew but the manifest prefix is unchanged (another writer appended WAL), replay only the WAL tail past the previously seen offset instead of the whole region.
3. A genuinely changed manifest (concurrent compaction/gc) keeps the full reload.

The head re-confirm contract (re-check under the lock right before advancing a ref) must be preserved exactly — only the mechanism of refreshing should change.

Post-gc behavior is already excellent (100-row commit on a 1M-row table ≈ 1ms), so this is the single remaining superlinear mechanism found by the probes. Note for anyone re-measuring: `build/` is compiled with `DOLTLITE_PROLLY_CHECK`, which adds an O(table) full-tree walk to every commit — perf must be measured on a plain build.

PR #1631, opened 2026-07-15
Add ratio-gated scaling-curve suite to PR CI

timsehn:
Adds `test/doltlite_scaling.sh` and a `scaling` job to Build-Test PR CI (ubuntu, 15-minute timeout; the suite runs in ~55s locally, so plenty of headroom including the plain build).

The suite gates **ratios, not wall-clock**, so runner speed cancels out — this avoids the load-sensitivity that makes the perf-ceiling jobs our only flaky CI:

**Segment A — per-commit cost vs history depth** (10k-row table, single-row commits):
- commits 1001–1200 vs 1–200: currently ~6–11× because every commit replays the whole WAL (#1630); gated at 20× as a backstop against anything worse, to be tightened to ~3× once incremental refresh lands
- post-gc commit cost ≤ 3× shallow-history cost (currently 1.0×)

**Segment B — per-op cost vs table size** (100k vs 1M rows, post-gc):
- open, 400 point lookups, 500-row commit each gated at 8× for 10× the rows (all currently 1.0×)
- structural sharing: ≤ 150 bytes/row at 1M (currently 66)

The job builds its own binary without `DOLTLITE_PROLLY_CHECK` — the checked build walks the full tree on every commit flush, which swamps exactly the curves this suite measures.

Backs the large-database confidence plan; the depth gate is the regression tripwire for #1630.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -5,6 +5,27 @@ on:
   workflow_dispatch:
 
 jobs:
+  scaling:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Install dependencies
+      run: |
+        bash .github/scripts/install-apt-deps.sh tcl-dev build-essential zlib1g-dev
+
+    # Plain build: DOLTLITE_PROLLY_CHECK adds a full-tree walk to every
+    # commit, which swamps the curves this suite gates on.
+    - name: Build
+      run: |
+        mkdir -p build-plain && cd build-plain
+        ../configure
+        make doltlite
+
+    - name: Scaling curves
+      run: bash test/doltlite_scaling.sh build-plain/doltlite
+
   wasm:
     runs-on: ubuntu-latest
     timeout-minutes: 20
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -5,27 +5,6 @@ on:
   workflow_dispatch:
 
 jobs:
-  scaling:
-    runs-on: ubuntu-latest
-    timeout-minutes: 15
-    steps:
-    - uses: actions/checkout@v4
-
-    - name: Install dependencies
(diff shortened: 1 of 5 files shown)


timsehn:
Extended per Tim's request: 10M-row size point (open/lookups/commit still flat: 30ms/37ms/41ms; scan and gc linear; 67B/row), depth-axis ops (log/checkout/merge ratios at depth 1200 vs 200), and a blob segment (16MB vs 1MB insert+commit+readback). Job moved from Build-Test to the Test Suite workflow. 24 gates, ~77s locally including the 10M seed.

timsehn:
Added correctness probes per Tim: closed-form full-content verification + dolt_diff_t history read-back at every Segment B size (diff cost also ratio-gated), commit-count invariant in Segment A, and a new 100M-row Segment D (seed 6.8GB, content verify, branch isolation). 32 gates, 6m18s locally.

Going to 100M immediately found a real bug: **dolt_gc OOMs on stores over ~2GB** — the rewrite assembles all live chunks in one int-sized buffer (#1633). Segment D is correctness-only until that lands, then 100M joins the post-gc perf matrix. Job timeout raised to 20 min.

timsehn:
gc fix stacked in per Tim: the rewrite now streams chunks through a fixed 64KB buffer instead of assembling the whole compacted store in one int-sized allocation (fixes #1633; fail-before/pass-after: a 2.6GB blob store OOMs on the old binary, compacts at 72MB peak RSS on the new one). 100M rows is ungated into the full Segment B perf matrix (gc at 6.85GB = 41s) plus a per-size branch-isolation probe; Segment D removed. Also fixed the CI seed failure — the 100M statement list exceeded Linux's 128KB per-argument execve limit, seeds now stream over stdin. 43 gates, 7m03s locally. Validated: gc_rewrite_failure 15/15, multi-process gc 101/101, gc scale 12/12, branch-gc stress 35/35, shell battery 87/87, C regression 309,830/309,830.

Issue #1638, opened 2026-07-15
Mixed partial staging (dolt_add table + dolt_commit -am) loses index entries across catalog numbering domains


timsehn:
Pre-existing edge split out of the -am stale-index fix:

```sql
CREATE TABLE a(id INTEGER PRIMARY KEY, v INTEGER); INSERT INTO a VALUES(1,10);
CREATE INDEX av ON a(v);
SELECT dolt_commit('-Am','c1');
CREATE TABLE b(id INTEGER PRIMARY KEY, w INTEGER); INSERT INTO b VALUES(1,5);
CREATE INDEX bw ON b(w);
SELECT dolt_add('b');                 -- stage the new table explicitly
UPDATE a SET v=11;
SELECT dolt_commit('-am','mix');      -- -a staging on top of partial staging
SELECT dolt_reset('--hard');
SELECT w FROM b INDEXED BY bw;        -- no such index: bw
```

On master before the -am index fix this case was broken worse (`av` — the *modified* table's index — vanished from the commit). With the fix, tracked tables' indexes are correct and only the partially-staged new table's index entry is lost.

Root cause: the staged catalog entry for `b` carries the add-time staged catalog's entry numbers, while the commit's canonical serializer resolves schema rows in the working/live numbering domain — the classic mixed-numbering-domain hazard. Fixing it needs the staged-sourced index entries to be re-paired through the staged catalog's own schema rows (same approach the -am fix uses for working-sourced indexes, applied to the staged domain).

Issue #1637, opened 2026-07-15
dolt_commit -a stages stale index roots: committed catalogs whose indexes disagree with their tables


timsehn:
Found by the new HTTP remote lifecycle suite, but the bug is transport-independent and purely local.

## Repro (deterministic)

```sql
CREATE TABLE users(id INTEGER PRIMARY KEY, name TEXT, age INTEGER);
INSERT INTO users VALUES(1,'alice',30),(2,'bob',25),(3,'charlie',35);
CREATE INDEX users_age ON users(age);
SELECT dolt_add('-A'); SELECT dolt_commit('-m','initial');
UPDATE users SET age=31 WHERE id=1;
SELECT dolt_commit('-am','bday');       -- the -a staging path
SELECT dolt_reset('--hard');            -- materialize from HEAD
SELECT group_concat(age) FROM users INDEXED BY users_age;  -- 25,30,35 (stale!)
PRAGMA integrity_check;                 -- row 1 missing from index users_age
```

Explicit `dolt_add('-A')` + `dolt_commit('-m', ...)` is correct; only the `-a`/`-am` shortcut is affected.

## Root cause

The `-a` staging overlay in `doltliteCommitFunc` (the `addModifiedOnly` block) pairs working entries into the staged catalog **by name**. Index entries are unnamed (`zName==NULL`), so they are invisible to the overlay: the staged catalog keeps HEAD's index roots while table entries get fresh working roots. The master entry got its by-number special case when the -am mixed-domain bug was fixed; index entries were missed — same unnamed-entry hazard class.

## Impact

Every `-am` commit after an index-touching write publishes a catalog whose indexes disagree with their tables. Queries planned through the index silently return wrong rows (`integrity_check` is the only tell), and the corruption propagates through push/pull/clone/branch-checkout — which is how the remote suite caught it (a fetched branch checkout showed a stale index across both http:// and file:// transports).

## Remaining edge (separate, pre-existing)

Mixed partial staging — `dolt_add('newtable')` then `dolt_commit('-am')` — loses index entries across catalog numbering domains, and did so before this fix too (on master the *modified* table's index vanishes; with the fix only the partially-staged new table's index is affected). Tracked separately.

PR #1645, opened 2026-07-15
Index x VC matrix suite, catalog write guardrail, and two merge fixes

timsehn:
The exhaustive coverage pass over indexes x version control. The recent finds (#1637, #1638, the reset-era residuals) share one structural root: **index entries are unnamed in catalogs, and every by-name catalog overlay is blind to them**. This PR attacks the class three ways:

**1. `test/doltlite_index_vc_matrix.sh`** (in the battery, 33 checks): every VC surface — add `-A`/named/partial, commit `-m`/`-am`/`-Am`, unstage all/named, reset soft/hard, branch checkout (incl. start-point), divergent merge, index-schema merge, cherry-pick, revert, CREATE/DROP INDEX through commits and old branches, ALTER RENAME, clone/push/pull, gc — against a fixture with every index flavor (secondary, UNIQUE, partial, expression, PK/UNIQUE autoindexes, WITHOUT ROWID, NOCASE). After every step: `integrity_check` plus index-planned vs table-scan agreement, verified on live state *and* on materialized commits (`reset --hard` on a copy).

**2. A write-time guardrail** (checked builds, in the canonical serializer): fails fast on the two structural violations detectable regardless of intent — an entry that pairs with no schema row or live meta (identity lost crossing numbering domains) and duplicate entry numbers. Zero fires across the entire battery, so it's a pure tripwire for future regressions of the class.

**3. Two real merge bugs the suite found on first runs, fixed here:**
- **False schema conflict**: DROP INDEX on one branch + table *data* changes on the other refused to merge ("table 'iv' modified on one branch and deleted on the other") — index roots move with table data, so root movement alone isn't modification. Definition-unchanged + dropped → the DROP wins; redefined + dropped stays a genuine conflict (and the message now says "index").
- **Silently stale adopted index**: an index created on one branch merged in with only that branch's rows — the other branch's inserts/updates were invisible to index-planned queries on the merged result (integrity_check: "row N missing from index"). Merge flows (merge, cherry-pick/revert, rebase replay) now rebuild adopted indexes over the merged tables before flushing the merged catalog.

Fail-before/pass-after: master fails `merge_index_schema` (both merge bugs), branch passes 33/33.

Validation: battery 88/88 (matrix included), C regression 309,830/309,830 with the guardrail active, http lifecycle 28/28, oracles vs real Dolt: merge 48/48, commit 37/37, reset 43/43.

Known residual in the class, documented rather than fixed here: unstage moves table entries but the matrix's unstage scenarios pass because staged catalogs re-derive indexes on the next add — the deeper unstage-index bookkeeping stays on the follow-up list.

### DIFF
--- a/src/doltlite.c
+++ b/src/doltlite.c
@@ -3044,9 +3044,12 @@ static void doltliteMergeFunc(
     char *zMergeErr = 0;
     SchemaMergeAction *aSchemaActions = 0;
     int nSchemaActions = 0;
+    char **azReindex = 0;
+    int nReindex = 0;
     rc = doltliteMergeCatalogs(db, &ancCatHash, &ourCatHash, &theirCatHash,
                                 &mergedCatHash, &nMergeConflicts, &zMergeErr,
-                                &aSchemaActions, &nSchemaActions, 0);
+                                &aSchemaActions, &nSchemaActions, 0,
+                                &azReindex, &nReindex);
     if( rc!=SQLITE_OK ){
       doltliteCommitClear(&ourCommit);
       doltliteCommitClear(&theirCommit);
@@ -3058,6 +3061,7 @@ static void doltliteMergeFunc(
       }
       doltliteTxnStateClear(&savedState);
       freeSchemaMergeActions(aSchemaActions, nSchemaActions);
+      doltliteFreeNameList(azReindex, nReindex);
       return;
     }
     sqlite3_free(zMergeErr);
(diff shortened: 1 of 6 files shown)


timsehn:
CI catch: the guardrail fired on three testfixture files (altercol, where, strict2) — all deliberate writable_schema vandalism (raw-deleted schema rows, rootpage-aliased rows) that stock and doltlite tolerate on the LIVE catalog. Scoped the guardrail to constructed entry arrays (staging/merge/reset), which is where the overlay bug class lives; vandalized live-catalog serialization is tolerated as before. Matrix 33/33, commit 51/51, merge 94/94 locally with the scoped check.

PR #1654, opened 2026-07-16
Add schema object dolt_status oracle coverage

### LINKED ISSUE #1655: dolt_status does not attribute index changes to their table
Sibling of #1651, found while closing the add/diff oracle gaps for schema objects:

```sql
CREATE TABLE t(a INTEGER PRIMARY KEY, v INTEGER);
INSERT INTO t VALUES(1,10);
CREATE INDEX iv ON t(v);
SELECT * FROM dolt_status;   -- empty (Dolt: t | 0 | modified)
```

Same for DROP INDEX. Status diffs table entries by root/schemaHash; index entries are separate and unnamed, so index-only changes are invisible. The `dolt_diff` summary had the same blindness and now attributes index schema-row differences to the parent table (schema_change=1) — status needs the same attribution: compare index schema rows per table between working/staged/HEAD catalogs and mark the parent table modified.

Sequencing note: `doltlite_status.c` is being modified by #1653 (view/trigger status); this fix should land after it to avoid conflicts. The dolt_add oracle scenarios for schema objects also project dolt_status, so they queue behind both fixes.

timsehn:
This adds a dedicated Dolt oracle suite for dolt_status schema-object behavior: index create/drop/replace, unique and composite indexes, column/schema changes, CHECK/UNIQUE/FK table definitions, mixed table/schema-object states, and filtered scans.

The new oracle exposed and this PR fixes:
- index-only schema changes were missing from dolt_status
- staging a named table could overwrite a different named staged entry when SQLite reused rootpages under FK DDL
- path reset of a staged dropped indexed table restored the table without its index entries
- fallback catalog serialization could not recover schema rows for unnamed index entries

Tests:
- make -C build -j$(sysctl -n hw.ncpu 2>/dev/null || nproc) DOLTLITE_PROLLY_CHECK=1 doltlite
- bash test/vc_oracle_status_schema_objects_test.sh build/doltlite dolt
- bash test/vc_oracle_status_test.sh build/doltlite dolt
- bash test/vc_oracle_add_test.sh build/doltlite dolt
- bash test/vc_oracle_reset_test.sh build/doltlite dolt
- bash test/vc_oracle_checkout_test.sh build/doltlite dolt
- bash test/vc_oracle_schemas_test.sh build/doltlite dolt
- bash test/vc_oracle_schema_diff_test.sh build/doltlite dolt
- bash test/vc_oracle_commit_test.sh build/doltlite dolt
- bash test/vc_oracle_diff_test.sh build/doltlite dolt

Fixes #1655

### DIFF
--- a/src/doltlite.c
+++ b/src/doltlite.c
@@ -788,7 +788,7 @@ static int addAppendTableEntry(
   if( pEntry->zName ){
     zDup = sqlite3_mprintf("%s", pEntry->zName);
     if( !zDup ){
-      sqlite3_result_error_nomem(context);
+      if( context ) sqlite3_result_error_nomem(context);
       return SQLITE_NOMEM;
     }
   }
@@ -796,7 +796,7 @@ static int addAppendTableEntry(
       *paEntries, (*pnEntries + 1) * (int)sizeof(struct TableEntry));
   if( !aNew ){
     sqlite3_free(zDup);
-    sqlite3_result_error_nomem(context);
+    if( context ) sqlite3_result_error_nomem(context);
     return SQLITE_NOMEM;
   }
   *paEntries = aNew;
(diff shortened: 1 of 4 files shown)


timsehn:
Reviewed and verified on the branch (checked build):

- The #1655 repro is fixed: CREATE INDEX and DROP INDEX on a committed table both report `t | 0 | modified` in dolt_status.
- Green against the branch: the new schema-objects status oracle 18/18, reset oracle vs Dolt 43/43, and the index×VC matrix 38/38 — the reset index-restoration and fallback-row changes compose cleanly with the staging fixes they build on.
- No file overlap with #1656 (diff-summary index attribution) — the two use the same schema-row attribution approach in disjoint surfaces; consolidating the two comparators into one shared helper is a reasonable follow-up, not a blocker.

Two things to address before merge:

1. **Stale base — needs a rebase onto current master.** The branch forked before #1652 (view/trigger staging) merged: `doltlite_commit.sh` fails its two view-staging gates here (`view_stays_out_of_named_commit`, `view_in_am_commit`) purely because the branch lacks `doltliteBuildNamedStageMasterRoot`. The pairing-hardening hunk sits in the same `addStageNamedTables` region #1652 touched, so after rebasing please re-run `doltlite_commit.sh` (56 cases) and `doltlite_index_vc_matrix.sh` to confirm the merge composes.

2. **Minor, non-blocking**: the new `unnamedRootMatch` arm still allows a named working table to pair with an *unnamed* staged entry on a bare number match — that's the pre-existing exposure (a cross-domain number collision would overwrite an index entry), narrowed but not eliminated for the unnamed case. Worth a follow-up tightening with the same by-schema-row resolution used elsewhere; also `statusCompareIndexSchemaObjects` runs two full catalog schema loads plus an O(from×to) name scan on every status query — fine at typical schema sizes, just noting it against the scaling suite's future appetite.

PR #1781, opened 2026-07-23
Consolidate catalog types and split feature oracle

timsehn:
## Summary
- give `TableEntry` and `SchemaEntry` one canonical owner in `doltlite_catalog_types.h` instead of duplicate guarded layouts in the VC and Prolly internals
- wire the shared type header into object dependencies and the DoltLite amalgamation generator
- replace the 13,373-line feature-interaction oracle monolith with a small runner and five documented feature families
- preserve all 424 sections and 752 Dolt differential cases while making new cases easier to place and review

## Tests
- `make -C build -j6 doltlite doltlite-remotesrv sqlite3.c`
- `bash test/vc_oracle_feature_interaction_test.sh build/doltlite dolt` (752 passed)
- `bash test/vc_oracle_merge_test.sh build/doltlite dolt` (74 passed)
- `bash test/vc_oracle_schema_merge_test.sh build/doltlite dolt` (54 passed)
- `bash test/run_doltlite_regression_case.sh all` (309,956 passed)
- `bash test/lint_layers.sh`
- `bash test/lint_orphaned_suites.sh`
- `bash test/dead_code_check.sh`
- `bash -n test/vc_oracle_feature_interaction_test.sh test/lib/vc_oracle_feature_interactions/*.sh`
- verified the amalgamation contains exactly one `TableEntry` and one `SchemaEntry` definition
- `make -C build devtest` reproduced the known unfiltered baseline (1,620 compatibility failures, then the final-file hang at 2,545/2,546; interrupted at 2m35s)


timsehn:
### CI failure analysis (`sqlite-regression-core-sql`)

**Only red check.** Everything else on the PR was green.

**Cause:** `bigsort` reported as `CRASH (unexpected): did not produce summary line` after **~452s**. That matches the **450s per-file `timeout`**, not a segfault from this PR.

- Timeline: `bigrow` finished `23:08:58` → bigsort “crash” `23:16:30` (~7.5 min)
- `bigsort` inserts 300k × 10KB blobs then `CREATE INDEX` with a huge cache; comment in workflow already notes ~260s healthy, 300s was too tight, 450s still too tight under `DOLTLITE_PROLLY_CHECK=1` + runner load
- PR diff is catalog type consolidation + oracle split — no sort/vdbesort path

**Fix pushed:** per-file timeout **450 → 600**, job timeout **20 → 25** min.

Re-run should clear if it was pure timeout; if bigsort still fails under 600s we can next skip it when free RAM is low or run it only without `PROLLY_CHECK`.

PR #1809, opened 2026-07-24
Confirm exact benchmark gates across retries

timsehn:
## Summary

- require the same individual, section, or suite gate to fail on all three attempts before CI reports a regression
- stop retrying as soon as the exact failed-gate intersection becomes empty, so unrelated noisy metrics cannot confirm one another
- embed suite identity and exact gate IDs in retry histories and reject mismatched or inconsistent artifacts
- separate the overall ratio from the gate result and enumerate every confirmed failed gate in the report
- mark one-off failures in the selected final measurement as transient diagnostics

This also fixes Ito GATE-3 from #1806: an `int` history relabeled by the caller as `vc` now fails closed.

At a 5% independent false-failure rate for one metric, exact three-attempt confirmation reduces that metric’s false confirmation probability to `0.05^3 = 0.0125%` (about 1 in 8,000).

## Validation

- `python3 test/benchmark_compare_test.py` (22 tests)
- `python3 test/benchmark_retry_test.py` (6 tests)
- Python bytecode compilation
- benchmark shell syntax checks
- `bash test/lint_layers.sh`
- actionlint across all workflows
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -932,8 +932,10 @@ a section, suite, or overall runtime regresses by more than 15% and 5ms.
 Short, process-startup-bound version-control operations use a 50% and 25ms
 individual gate; their aggregate still uses the 15% suite gate. A suite that
 exceeds a gate is measured again automatically and only fails CI after three
-consecutive failing attempts. Command, result-format, and revision-provenance
-errors fail immediately instead of being retried.
+consecutive failures of the same individual, section, or suite gate. Different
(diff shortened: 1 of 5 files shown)


timsehn:
@itoqa Follow-up #1811 binds selected results and retry histories to the same producer run, rejects stale/mixed pairs, and stages all canonical artifacts before promotion. It includes regressions for both findings from this run.

PR #1811, opened 2026-07-25
Harden benchmark retry artifacts

timsehn:
Follow-up to #1809.

This closes the two artifact-integrity gaps identified by Ito and makes retry behavior easier to diagnose in live CI logs.

- assign one producer ID to each suite retry run and embed it in both the selected result and attempt history
- reject stale or mixed result/history artifacts before rendering or enforcing a performance conclusion
- preflight and stage the result, samples, and markdown before exposing canonical artifact names
- report promotion failures cleanly without leaving Ito's partial canonical state
- log each attempt number, exact failing/continuing gates, per-attempt duration, cumulative duration, and stop/confirmation reason

Validation:
- `python3 test/benchmark_compare_test.py` (24 tests)
- `python3 test/benchmark_retry_test.py` (8 tests)
- `python3 -m py_compile test/benchmark_compare.py test/benchmark_compare_test.py test/benchmark_retry.py test/benchmark_retry_test.py`
- benchmark shell syntax checks
- `bash test/lint_layers.sh`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -935,7 +935,9 @@ exceeds a gate is measured again automatically and only fails CI after three
 consecutive failures of the same individual, section, or suite gate. Different
 metrics failing on later attempts do not confirm the original regression.
 Command, result-format, and revision-provenance errors fail immediately instead
-of being retried.
+of being retried. Retry logs identify the current attempt, elapsed time, and
+exact gates still eligible for confirmation. The selected result and its retry
+history carry a shared producer identity so stale artifacts cannot be paired.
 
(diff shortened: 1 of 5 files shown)


timsehn:
@itoqa Fixed in b6486257b3. Promotion now backs up the complete prior canonical set before exposing staged files, restores it on any replacement failure, and reports a compound error if rollback itself fails. Added the injected mid-exposure replacement regression; all prior files are restored byte-for-byte and no staging/backup files remain.

Issue #1846, opened 2026-07-29
Unsupported PRAGMAs are no-ops


timsehn:
DoltLite accepts every `PRAGMA` that configures the pager, the journal, or the
on-disk layout, and then ignores it. The value never changes and no error is
raised. This follows SQLite's own convention — SQLite silently accepts unknown
pragmas and unsupported values rather than erroring — but it means a caller
cannot tell that the setting did nothing.

The dividing line is DoltLite's architecture: everything above `btree.h` is
upstream SQLite and works unchanged, so connection-level pragmas behave
normally. Everything below it is the prolly chunk store, which has no pages, no
freelist, no rollback journal and no `-wal`/`-shm` sidecars, so pragmas
describing those have nothing to configure.

Verified by setting each pragma on a brand-new database and reading it back,
against stock SQLite for comparison.

### Silently accepted, no effect

| PRAGMA | DoltLite always reports | Stock |
|---|---|---|
| `journal_mode` | `wal` | honours `delete`/`memory`/`off`/… |
| `auto_vacuum` | `0` | `2` when set before creation |
| `secure_delete` | `0` | `1` |
| `journal_size_limit` | `-1` | `32768` |
| `mmap_size` | `0` | `1048576` |
| `encoding` | `UTF-8` | `UTF-16le` (see #1838) |
| `writable_schema` | `0` | `1` |

`page_size` is a special case: DoltLite accepts it *and reports the new value*,
so it looks honoured while meaning nothing — the store has no pages.

### Readable, but synthetic

`page_count` (chunk count, not file pages), `freelist_count` (always 0),
`schema_version` (hash-derived, not a counter), `data_version`.
`lock_status` and `stats` exist only in `SQLITE_TEST` builds and report
chunk-store internals that have no page-level equivalent.

### Explicitly rejected

`wal_checkpoint(FULL|RESTART|TRUNCATE)` is the one pragma that errors rather
than no-oping: "wal_checkpoint mode is not configurable on doltlite-format
databases; use the default (PASSIVE) form".

### Supported, behaves as stock

`locking_mode`, `cache_size`, `max_page_count`, `read_uncommitted`,
`synchronous`, `temp_store`, `foreign_keys`, `recursive_triggers`,
`legacy_alter_table`, `ignore_check_constraints`, `automatic_index`,
`query_only`, `defer_foreign_keys`, `cell_size_check`,
`reverse_unordered_selects`, `analysis_limit`, `threads`, `busy_timeout`,
[...]

Issue #1839, opened 2026-07-29
testfixture: page-image APIs unsupported (sqlite3_backup_*, sqlite3_serialize/deserialize, memdb VFS)


timsehn:
`test/known_testfixture_exception_inventory.txt` marks 6 inherited SQLite
testfixture suites `unsupported` because they exercise APIs that hand out or
consume a raw SQLite *page image*. DoltLite stores content-addressed prolly
chunks and has no stock page image to produce, so:

- `sqlite3_backup_init` rejects the page-image backup (`backup`, `backup2`,
  `backup4`, `backup5`, `backup_malloc`)
- `sqlite3_serialize` returns NULL cleanly and the following
  `sqlite3_deserialize` aborts the file (`memdb1`)
- the chunk store deliberately refuses the memdb VFS at open, so `db` is never
  created (`memdb2`)

DoltLite exports through a VFS-level backup (`sqlite3_js_db_export` under
`DOLTLITE_PROLLY`) rather than the page-image API. Not expected to be fixed.

The issue exists so the inventory's `unsupported` entries point at a live
description of the limitation. `backup*` and `memdb2` previously pointed at
 #1208 (the closed triage *process* tracker) and `memdb1` at #1225 — which was
the narrower "ASan: sqlite3_serialize NULL deref" bug and was genuinely fixed,
so it no longer describes why the suite is still excepted.

Affected suites (7): backup backup2 backup4 backup5 backup_malloc memdb1 memdb2

timsehn:
Update: file-backed and attached DoltLite backup support landed in #2255, and #2258 adds native file-to-memory, memory-to-file, and memory-to-memory backup. The remaining serialize/deserialize work is tracked separately in #2257; memdb VFS support also remains outside #2258.

timsehn:
Stale. `sqlite3_backup_*` and `sqlite3_serialize`/`deserialize` work; they copy a **native DoltLite** image, not a SQLite page file (README `c_api.backup` / `c_api.serialize`). The claim that backup_init rejects and serialize returns NULL is no longer true.

Replacements:
- #2491 — backup/serialize are native images; inherited `backup2`/`backup4` still assert SQLite page-file hashes
- #2492 — `attachmalloc` OOM index shifts (never a page-image API gap; previously miscited here)

Inventory citations move in https://github.com/dolthub/doltlite/pull/2488.

PR #1855, opened 2026-07-30
Refuse writes to a database that was renamed out from under us (#1853)

### LINKED ISSUE #1853: Renaming an open database leaves a malformed file at the old path and empties the live connection
Renaming or unlinking a DoltLite database while a connection is open leaves a
**malformed file at the original path** and the live connection pointed at an
empty store. Stock SQLite reports `SQLITE_READONLY_DBMOVED` and changes nothing.

Reproduced 3/3 on `db9e3e9d86`.

## Repro

```sh
doltlite a.db "CREATE TABLE t(id INTEGER PRIMARY KEY, v TEXT);
               INSERT INTO t VALUES(1,'x');
               SELECT dolt_add('-A'); SELECT dolt_commit('-m','i');"

doltlite a.db ".shell mv a.db b.db" \
              "INSERT INTO t VALUES(2,'after');" \
              "SELECT 'rows='||count(*) FROM t;"
```

DoltLite:

```
a.db recreated : YES
  integrity_check on it: *** in database main ***   (integrity check failed)
b.db (the real store): 1 row -- the post-rename INSERT is gone
```

Stock SQLite, same script:

```
Error: stepping, attempt to write a readonly database (8)
a.db recreated : no
b.db: 1 row
```

So DoltLite manufactures a corrupt database at the vacated path, loses the
write, and the open connection stops seeing its own committed data. Nothing
warns.

`mv` on an open database is not exotic: log rotation, atomic-replace deploys and
backup scripts all do it, and `tester.tcl`'s `forcedelete` does the unlink
variant.

## Mechanism

`csReloadFromDisk` (`chunk_store_lock.c:271`) reopens by path with
`SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE`, then `csAdoptOpenedStoreState`
makes the live handle adopt the result. When the path has been vacated,
`OPEN_CREATE` creates an empty store there and the connection adopts *that* --
which is both symptoms at once.

`csAdoptOpenedStoreState` has exactly one caller, so every instance of this goes
through that reload. It is reached from `chunkStoreRefreshIfChanged`
(`chunk_store_lock.c:240`) for reads *and* writes;
`csReloadFromDiskPreservingLocalRefs` is not involved.

## What makes the fix non-obvious

A guard in `csReloadFromDisk` stops the corruption and fixes four of the gated
`pager4` assertions, but breaks `pager4-1.2`: reads share that path, and stock
requires reads to keep working from the open handle after a rename. Preventing
the adopt in `csDetectExternalChanges` fixes reads, and then the write refusal
has to live in `chunkStoreCommit` -- which is only reachable *because* the adopt
no longer recreates the file. (An earlier attempt put the guard there while the
adopt still ran, so `csMovedFileIsGone` was already false by the time commit
reached it, and the guard silently never fired.)

The remaining blocker is the predicate. "Nothing at the path" conflates three
states:

1. renamed or unlinked away -- must refuse the write
2. **not created yet** -- a new database has no file until first commit
3. **deliberately deleted** -- `tester.tcl` `forcedelete`s with a handle open

`SQLITE_FCNTL_HAS_MOVED` does not separate them either: `fileHasMoved`
(`os_unix.c`) reports moved whenever `stat(path)` fails, so cases 2 and 3 look
identical to case 1. Gating on `cs->file.iFileSize > 0` was not enough --
`pager4-1.1` still failed before any rename.

### Proposed fix

Stop inferring. Record the file's `(st_dev, st_ino)` when the store is opened and
when it commits, and define "moved away" as *our recorded inode is no longer
reachable at the path*. That is exact for all three cases: a not-yet-created
store has no recorded inode, a deliberate delete leaves the recorded inode
unreachable but with no committed data behind it, and a rename leaves a recorded
inode that now lives elsewhere.

## Gated assertions

`pager4` currently has 7 gates, all dispositioned `intentional -` with the
one-line reason "rename-while-open: commit goes BY PATH, creating a fresh db at
the original path (silent fork)". That label asserts this is by design; it is a
bug. Measured split, from a build that refused the write:

- **this issue** -- `pager4-1.3`, `1.9`, `1.10`, `1.11` pass as soon as the write
  is refused
- **this issue, foreign-database variant** -- `pager4-1.4`: a *different*
  database is created at the vacated path, so an existence check is not enough
  and the inode comparison above is required
- **#1846** -- `pager4-1.7`, `1.8` expect `journal_mode=OFF`/`MEMORY` to permit
  the write, which needs journal mode to stop being a no-op

"Silent fork" undersells it: the durable artifact is a file that
`integrity_check` rejects.

timsehn:
Fixes #1853. Rebased on master (includes #1854's accounting).

## The bug

Renaming or unlinking an open database left a file at the old path that `integrity_check` rejects, lost the following write, and left the connection reading an empty store. Stock reports `SQLITE_READONLY_DBMOVED` and changes nothing.

## Two causes

**1. The reload could create a database.** `csReloadFromDisk` reopened by path with `SQLITE_OPEN_CREATE`, and `csAdoptOpenedStoreState` adopted whatever came back — so a vacated path got fabricated as an empty store and adopted over the live one. Dropping `OPEN_CREATE` is the principle: *a reload re-derives state from the path and must never bring a database into existence.*

**2. The write still landed** through the fd we already held. `chunkStoreCommit:489` already refuses on `cs->readOnly` with exactly stock's message, so the store is now marked `movedReadOnly` when the file has moved and nothing replaced it, and refused there. A separate flag, because `cs->readOnly` records how the file was *opened*. `csDetectExternalChanges` maintains it on every refresh, so it isn't latched — restoring the file restores writability (`1.5`, `1.6`).

Reads keep working throughout, which stock requires (`1.2`): the refresh reports no change for a vacant path and keeps serving the open handle.

The predicate gates on `HAS_MOVED` **before** testing the path — that's what separates a rename from a store whose file doesn't exist yet, since `fileHasMoved` reports moved whenever `stat()` fails.

## Results

`pager4`: **7 gated failures → 3**, so the ratchet drops — `gates 5375 → 5371`, `engine-gap 69 → 65`. Remaining: `1.4` (a *different* database at the vacated path, still #1853) and `1.7`/`1.8` (want `journal_mode=OFF`/`MEMORY`, need #1846).

Repro now matches stock exactly:

```
read=1
Error: attempt to write a readonly database
a.db recreated: NO | b.db rows: 1
-- rename back --
read=1  after=2        (writability restored, nothing latched)
```

Also: 92/92 shell suites, 24/24 C tests (incl. multi-process + concurrency), `dolt_gc` intact across its atomic store replacement, `incrblob_err` clean, all inventory checks pass.

## Notes for review

Everything is measured **in a clean build directory**. `pager4` renames `test.db` to `test-xyz.db`, so leftovers from an earlier run change the outcome — that invalidated two of my earlier measurements on this branch, including a "7 → 3" I reported before it was true.

Two things from the previous draft were artifacts, not real: the `incrblob_err` OOM regression doesn't reproduce in any configuration, and the "CLI vs testfixture disagreement" was that dirty directory.

The CI lint failure on the draft was my comment containing the literal string `rename(2)`, which `lint_no_raw_os_fileio.sh` matches as raw OS file I/O. Reworded.

### DIFF
--- a/src/chunk_store.h
+++ b/src/chunk_store.h
@@ -163,6 +163,14 @@ struct ChunkStore {
   ChunkStaging staging;
 
   u8 readOnly;
+
+  /* Set while the file this handle opened has been renamed or unlinked away and
+
+  ** nothing replaced it: writes are refused, reads keep serving the open handle.
+
+  ** Re-evaluated on every refresh, so restoring the file restores writability. */
+
+  u8 movedReadOnly;
   u8 isMemory;
   u8 fullFsync;           /* PRAGMA fullfsync: syncs use SQLITE_SYNC_FULL */
   u8 noSync;              /* PRAGMA synchronous=OFF: skip durability syncs */
--- a/src/chunk_store_commit.c
+++ b/src/chunk_store_commit.c
@@ -486,7 +486,7 @@ int chunkStoreCommit(ChunkStore *cs){
 
   memset(&savedRefs, 0, sizeof(savedRefs));
   if( cs->corruptMidStream ) return SQLITE_CORRUPT;
-  if( cs->readOnly ) return SQLITE_READONLY;
+  if( cs->readOnly || cs->movedReadOnly ) return SQLITE_READONLY;
   if( cs->isMemory ) return csCommitToMemory(cs);
   if( !csFileLockHeld(CS_GRAPH_LOCK(cs)) && cs->file.zFilename ){
     preserveRefs = cs->staging.nPending > 0
--- a/src/chunk_store_int.h
+++ b/src/chunk_store_int.h
@@ -27,6 +27,7 @@ int csFileLock(sqlite3_vfs *pVfs, const char *path,
 void csFileUnlock(sqlite3_file *pFile, char **pzName);
 int csFileLockNB(sqlite3_vfs *pVfs, const char *path,
                  sqlite3_file **ppFile, char **pzName);
+int csMovedFileIsGone(ChunkStore *cs);
 int csReloadFromDisk(ChunkStore *cs);
 int csReloadFromDiskPreservingLocalRefs(ChunkStore *cs);
 int csFileSizeByName(sqlite3_vfs *pVfs, const char *zPath, i64 *pSize);
--- a/src/chunk_store_lock.c
+++ b/src/chunk_store_lock.c
@@ -178,6 +178,25 @@ void chunkStoreUnlock(ChunkStore *cs){
   }
 }
 
+/* True when the file this handle holds has moved out from under us and nothing
+** took its place. GC replaces the store atomically, which keeps the path
+** populated throughout, so a vacant path distinguishes "renamed or unlinked
+** away" from "upgraded beneath us". */
+int csMovedFileIsGone(ChunkStore *cs){
+  int bMoved = 0, exists = 0;
+  if( cs->isMemory || cs->file.pFile==0 ) return 0;
+  if( cs->file.zFilename==0 || cs->file.zFilename[0]==0 ) return 0;
+  if( sqlite3OsFileControl(cs->file.pFile, SQLITE_FCNTL_HAS_MOVED,
+                           &bMoved)!=SQLITE_OK || !bMoved ){
+    return 0;
+  }
+  if( sqlite3OsAccess(cs->file.pVfs, cs->file.zFilename,
+                      SQLITE_ACCESS_EXISTS, &exists)!=SQLITE_OK ){
+    return 0;
+  }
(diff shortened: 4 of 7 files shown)


timsehn:
CI fault-fuzz was failing on unexpected `backup_malloc-1.transient.66` / `.67`. Same class as the existing backup_malloc OOM gates (#1839): the moved-file checks on the refresh path shift fault points. Gated those two and bumped `unsupported` 1305 → 1307 (gates 5371 → 5373). Net vs pre-PR still better on the pager4 engine-gap drop.

Issue #1864, opened 2026-07-30
Databases created in the same second share an initial commit hash


timsehn:
Two DoltLite databases created in the same wall-clock second have the **same
initial commit hash**, so they are indistinguishable by history root. Nothing
about a database's identity survives to tell it apart from an unrelated one
created moments earlier.

## Repro

Three databases, created back to back:

```sh
for i in 1 2 3; do doltlite d$i.db "CREATE TABLE t(x);"; done
for i in 1 2 3; do
  doltlite d$i.db "SELECT substr(commit_hash,1,16), date
                   FROM dolt_log ORDER BY rowid DESC LIMIT 1;"
done
```

```
d1  031738c110112d37  2026-07-30 17:51:37
d2  031738c110112d37  2026-07-30 17:51:37
d3  031738c110112d37  2026-07-30 17:51:37
```

Two seconds apart, they differ:

```
a: 625e7020999b7730  2026-07-30 17:52:22
b: 8d83f236de6be2f3  2026-07-30 17:52:24
```

So the timestamp *is* hashed — it just has one-second resolution, and everything
else about a fresh database is constant.

## Mechanism

`doltliteCommitInitialSeed` (`doltlite_config.c:191`) creates the seed commit
with an empty parent, an empty catalog, the fixed message
`"Initialize data repository"`, and **NULL author name and email**:

```c
rc = doltliteCreateAndStoreCommit(db, &emptyParent, &emptyCatalog,
    "Initialize data repository", NULL, NULL, 0, 0, &seedHash);
```

The timestamp is part of the hashed payload (`DLC_PUT_I64(p, c->timestamp)`,
`doltlite_commit.c:57`) at second resolution, so within one second every input
is identical and content addressing does what it should: same content, same hash.

This is not a hashing bug. It is that a fresh database carries no distinguishing
content.

## What Dolt does

Dolt has the same shape and the same resolution — `env/environment.go` builds the
initial commit metadata with:

```go
Timestamp: uint64(time.Now().Unix())
```

via `datas.MakeCommitMetaGenerator(name, email, t)`. No host identifier and no
random nonce (ignoring the `--fun` hash-mining path).

The difference is that Dolt threads the configured **user name and email** into
that metadata, while DoltLite passes `NULL` for both. So in Dolt two databases
initialised by different users differ; collisions need the same user in the same
second. DoltLite has dropped even that entropy.

## Why it matters

Beyond hygiene, this blocks a real fix. When a chunk-store file is replaced
[...]

PR #1872, opened 2026-07-30
Seed every database, and prove a replaced file is ours by its head commit (#1853, #1873)

### LINKED ISSUE #1873: ATTACHed databases never get a seed commit: data with an empty dolt_log and a zero branch tip
An ATTACHed DoltLite database never gets a seed commit, so it ends up holding data
with **no history at all** — and the damage is permanent.

## Repro

```sh
doltlite m.db "ATTACH 'a2.db' AS aux;" \
              "CREATE TABLE main.t1(x); INSERT INTO main.t1 VALUES(1);" \
              "CREATE TABLE aux.t2(y);  INSERT INTO aux.t2 VALUES(1);"

doltlite m.db  "SELECT name, substr(hash,1,12) FROM dolt_branches;"
doltlite a2.db "SELECT name, substr(hash,1,12) FROM dolt_branches;"
```

```
main db:      main  6e00e5bfd100
ATTACHED db:  main  000000000000
```

Opening the attached-created database directly as main does not repair it:

```
SELECT count(*) FROM dolt_log;   -- 0
SELECT count(*) FROM t2;         -- 1   (the data is there)
```

So `dolt_log` is empty on a database with content, and its default branch points at
the zero hash. An explicit `dolt_commit` does work and produces the first entry, so
the database is usable — it just has no creation history and no root commit.

## Mechanism

`doltliteMaybeSeedRepo` is called once per connection from `doltliteInit`, and
resolves its store with `doltliteGetChunkStore(db)`, which is hardcoded to
`db->aDb[0]`. So only the connection's **main** database is ever seeded.

The attached database instead reaches its first working-set persist unseeded, where
`btreePersistWorkingSet` handles the missing default branch by creating it pointing
at `pWorkingCommit` — empty when nothing has been committed
(`prolly_btree_state.c`, whose own comment describes exactly this: *"A freshly
attached doltlite db is never seeded — doltliteMaybeSeedRepo only runs for the main
db — so its default branch doesn't exist. Create it on first persist (pointing at
the current head, empty if uncommitted)"*).

After that the branch exists, so `doltliteMaybeSeedRepo`'s
`refsTableBranchCount(&cs->refs) > 0` early-out declines to seed it on every
subsequent open. The state is unrepairable by reopening.

## Why it matters beyond dolt_log

This also silently defeats anything that identifies a database by its history root.
It is what made the `HAS_MOVED` ownership check in #1853 look unimplementable: a
peer's `dolt_gc` legitimately sweeps a stale working-set root, so the head commit is
the only staleness-robust thing to probe — and for attached databases there was no
head commit to probe. `threadtest4 --serialized 6` (which ATTACHes two of its three
databases) was the test that exposed it.

Fixed in the same PR as #1853, since #1853 cannot be correct without it: ATTACH now
seeds through a store-scoped path, and the seed guard tests whether the default
branch *names a commit* rather than whether any branch exists, which repairs
already-damaged databases on open.

timsehn:
Fixes #1873 and the remaining half of #1853. **Force-pushed over the earlier
working-set-probe approach, which CI proved wrong — see the comment above.**

You were right that no extra metadata was needed: the head commit is the correct
thing to probe. It only looked unusable because of a real bug underneath.

## #1873 — attached databases never got a seed commit

`doltliteMaybeSeedRepo` runs once per connection from `doltliteInit` and resolves
its store from `aDb[0]`, so it only ever seeded **main**. An attached database
instead got its default branch created by the first working-set persist, pointing at
the empty hash, and the old "has any branch" early-out then declined to repair it for
the rest of the database's life:

```
main db:      main tip=6e00e5bfd100
ATTACHED db:  main tip=000000000000    dolt_log rows: 0   (rows in table: 1)
```

ATTACH now seeds through a store-scoped path, and the guard tests whether the
default branch *names a commit* rather than whether any branch exists, so an
already-damaged database is repaired on open. Any other populated repo is left
alone — a database whose branches don't include the recorded default is never handed
a fabricated one.

Main keeps seeding through `doltliteAdvanceBranch`, which also sets the session head
and staged hashes, switches the catalog and persists the working set. Replacing that
with a bare `chunkStoreAddBranch` is what broke `gc_vs_merge` and `stale_session` on
my first attempt.

## #1853 — prove ownership by the head commit

Read-only on `HAS_MOVED` unless the connection can prove the file now at its path is
its own database. The proof: the candidate — opened read-only, no `OPEN_CREATE` —
holds one of our branch tips. GC preserves everything reachable from the refs it
sweeps under, and our tip stays an ancestor of whatever the branch became, so this
survives an arbitrarily stale handle.

Working-set roots cannot be the proof (my first attempt): a peer advancing the
working set makes GC sweep the one we remember, so a live database fails its own
test. `threadtest4 --serialized 6` caught that — **`VACUUM` *is* GC** in DoltLite
(`sqlite3RunVacuum` → `doltliteGcCompactWithPhase`), so concurrent writers plus a
store-replacing GC is an everyday workload.

**Backup** is a second sanctioned replacer that can never be proved by content, since
it installs unrelated content at the destination on purpose. It now declares the
replace with a one-shot `adoptReplacement`, consumed in `chunkStoreRefreshIfChanged`
where the reload happens — deliberately not in `csDetectExternalChanges`, because the
read-to-write upgrade path calls `chunkStoreHasExternalChanges` as a pure probe and
must not consume it, and a sanctioned replacement must be adopted even under a
pinned snapshot whose file is gone.

## Measured blast radius

| our database's history | stranger appears at the vacated path | result |
| --- | --- | --- |
| ≥1 `dolt_commit` | write refused, matches stock; our store untouched | fixed |
| seed-only (never committed) | still adopts the stranger | `pager4-1.4`, gated |

## Gates

- `pager4-1.4` stays gated, reason corrected to the true one: two databases created
  in the same second are byte-identical, so a same-second stranger carries our own
  seed commit and no content hash can separate them.
- `pager4-1.7`/`1.8` keep their gates (#1846) but also had a wrong reason —
  `journal_mode` is not "rejected", the PRAGMA is a no-op reporting `wal`, and the
  write is refused because there is no journal mode for OFF/MEMORY to switch off.
- `backup_malloc-1.transient.66`/`.67` need no action here: #1867 removed them on
  master after its own OOM re-sweep, and rebasing onto that keeps `backup_malloc`
  clean with this change too (2184 iterations, 15 known divergences, no entry
  reported fixed). Rebase resolution notes are below.

## Verification

- `threadtest4 --serialized 6`: clean 3/3 (master clean 4/4; previous attempt failed 3/3)
- `multi_process_gc_test`: 101/101 ×3
- `pager4`: down to its three gates
- C tests: 24/24 on freshly built binaries
- shell suites: 95/95
- `doltlite_regression_test_c`: 309990/309990
- `backup_malloc`, `backup2`, `backup4`, `backup5`: clean
- `make lint`, gate inventory and ratchet: clean

New regression `write_rejects_foreign_database_at_path` fails four assertions on the
unfixed engine and passes with the fix.

## Open question

Closing the `pager4-1.4` residual needs the seed commit to be unique per database.
My earlier claim that this needs no format change was too glib: the only spare
non-user-visible slot is a new commit format field (`DOLTLITE_COMMIT_V3`), i.e.
exactly the #1547 area #1864 declined. The alternatives put a random token in
`author`/`email`/`message`, which is user-visible in `dolt_log` forever and diverges
from Dolt (also second-resolution, no nonce). Given the residual loses no data and
only affects a never-committed database whose path is taken over within the same
second, my recommendation is to leave it gated. Easy to add if you'd rather close it.

## Rebase notes (onto #1867)

#1867 landed in the same code and this was rebased onto it rather than stacked.
Resolutions:

- `csMovedFileGone` → `csMovedFileIsOurs`: this replaces the predicate outright, and
  already carries #1867's substance — rc returned with the verdict as an
  out-parameter, no redundant `HAS_MOVED` re-query, helper `static` with the
  declaration out of `chunk_store_int.h` (that last one auto-resolved, master had
  already removed it).
- #1867's `iFileSize<=0` early-out is **kept**, with its reasoning: the VFS reports
  moved whenever it cannot resolve the path at all, so a store whose file is still to
  be created — a fresh database, or a backup target before its first write — must not
  read as moved.
- `backup_malloc` 66/67: took master's removal over this branch's `@linux`
  qualification, then re-ran the sweep to confirm.

### DIFF
--- a/src/attach.c
+++ b/src/attach.c
@@ -245,6 +245,18 @@ static void attachFunc(
 #endif
     if( !REOPEN_AS_MEMDB(db) ){
       rc = sqlite3Init(db, &zErrDyn);
+#ifdef DOLTLITE_PROLLY
+      /* Seed the newly attached database. doltliteMaybeSeedRepo only ever reaches
+      ** aDb[0], so without this the attached database's default branch is created
+      ** by its first working-set persist pointing at the empty hash: data with no
+      ** history, and no later chance to repair it. Reported through rc so a
+      ** failure runs the same teardown as a failed sqlite3Init instead of leaving
+      ** the database attached behind an error. */
+      if( rc==SQLITE_OK ){
+        extern int doltliteSeedAttachedDb(sqlite3*, int);
+        rc = doltliteSeedAttachedDb(db, db->nDb - 1);
+      }
+#endif
     }
     sqlite3BtreeLeaveAll(db);
     assert( zErrDyn==0 || rc!=SQLITE_OK );
@@ -270,7 +282,7 @@ static void attachFunc(
     }
     goto attach_error;
   }
-  
+
   return;
 
 attach_error:
--- a/src/chunk_store.h
+++ b/src/chunk_store.h
@@ -164,13 +164,17 @@ struct ChunkStore {
 
   u8 readOnly;
 
-  /* Set while the file this handle opened has been renamed or unlinked away and
-
-  ** nothing replaced it: writes are refused, reads keep serving the open handle.
-
-  ** Re-evaluated on every refresh, so restoring the file restores writability. */
-
+  /* Set while the file this handle opened is no longer the file at its path and
+  ** the one there cannot be shown to be this database: writes are refused, reads
+  ** keep serving the open handle. Re-evaluated on every refresh, so restoring the
+  ** file restores writability. */
   u8 movedReadOnly;
+
+  /* One-shot: adopt whatever file is now at the path on the next refresh. Only an
+  ** operation that installed a database there itself may set this -- a backup
+  ** destination is replaced with unrelated content on purpose, which the
+  ** moved-file check would otherwise refuse as a foreign file. */
+  u8 adoptReplacement;
   u8 isMemory;
   u8 fullFsync;           /* PRAGMA fullfsync: syncs use SQLITE_SYNC_FULL */
   u8 noSync;              /* PRAGMA synchronous=OFF: skip durability syncs */
--- a/src/chunk_store_lock.c
+++ b/src/chunk_store_lock.c
@@ -239,33 +239,61 @@ void chunkStoreUnlock(ChunkStore *cs){
   }
 }
 
-/* Whether the file this handle opened has been renamed or unlinked away with
-** nothing taking its place. The caller has already established that the VFS
-** reports it moved; GC replaces the store atomically, which keeps the path
-** populated throughout, so a vacant path is what separates "moved out from under
-** us" from "upgraded beneath us".
+static int csStoreHasAnyBranchTip(ChunkStore *pCand, const RefsTable *rt,
+                                  int *pHas){
+  int i;
+  *pHas = 0;
+  for(i=0; i<rt->nBranches; i++){
+    const ProllyHash *pTip = &rt->aBranches[i].commitHash;
+    int rc, has = 0;
+    if( prollyHashIsEmpty(pTip) ) continue;
+    rc = chunkStoreHas(pCand, pTip, &has);
+    if( rc!=SQLITE_OK ) return rc;
+    if( has ){
+      *pHas = 1;
+      return SQLITE_OK;
+    }
+  }
+  return SQLITE_OK;
+}
+
+/* Whether the file now at this handle's path is provably the database this handle
+** opened. Upstream makes any moved file read-only for good; GC is DoltLite's one
+** legitimate reason for the path to hold a different file, and a GC replacement,
+** a rename and a vacant path are indistinguishable from the outside. So require
+** proof: the candidate holds a branch tip of ours. GC preserves every commit
+** reachable from the refs it sweeps under, and our tip stays an ancestor of
+** whatever the branch has since become, so this survives an arbitrarily stale
+** handle. No proof leaves the upstream verdict in place.
 **
-** Never written means never moved: the VFS reports moved whenever it cannot
-** resolve the path at all, so a store whose file is still to be created -- a fresh
-** database, or a backup target before its first write -- is indistinguishable
-** from one renamed away unless this handle has already seen the file on disk.
-**
-** Returns an error rather than a verdict when the existence check itself fails:
-** an I/O error is not evidence about where the database went. */
-static int csMovedFileGone(ChunkStore *cs, int *pGone){
-  int exists = 0;
+** Working-set roots cannot be the proof: a peer advancing the working set makes
+** GC sweep the one we remember, so a live database fails its own test. */
+static int csMovedFileIsOurs(ChunkStore *cs, int *pIsOurs){
+  ChunkStore cand;
   int rc;
 
-  *pGone = 0;
+  *pIsOurs = 0;
   if( cs->isMemory || cs->file.pFile==0 ) return SQLITE_OK;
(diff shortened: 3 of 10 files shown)


timsehn:
Draft — **this approach does not work.** CI caught it and the failure is
fundamental, not a tuning problem.

`threadtest4 --serialized 6` fails with `SQLITE_READONLY` on ordinary DML
(`DELETE FROM t2 WHERE tid=5`). Reproduced locally 3/3; master is clean 4/4 on
the same amalgamation build, so it is this change.

Cause: in DoltLite `VACUUM` *is* GC (`sqlite3RunVacuum` calls
`doltliteGcCompactWithPhase`), and threadtest4 runs `VACUUM` from one worker
while five others write. So "concurrent writers + a store-replacing GC" is a
completely ordinary workload, and every writer's handle hits `HAS_MOVED` after
each VACUUM. Instrumented, every peer handle at that moment looks like this:

```
DBG open rc=0 nBr=1
DBG br0 emptyTip=1 has=0(rc-1) emptyWs=0 has=0(rc0) notadb=0
DBG verdict rc=0 ours=0
```

Two things are true at once:

- `emptyWs=0 has=0` — the handle's working-set root is real but **legitimately
  absent** from the compacted store: the other writers advanced the working set,
  and GC swept the stale one. Correct GC behaviour.
- `emptyTip=1` — the branch commit tip is **all zeros**. A database written only
  with plain SQL has no commit at all until `dolt_commit`, so there is no tip to
  fall back on.

So at the moment the proof is needed there is *nothing* of ours left in the
candidate to prove ownership with. This is not specific to the working-set root:
**any** content-based probe fails here, because GC's job is to discard exactly
the content a stale handle knows about. I also measured the
`{tip, working set}` disjunction — same failure, for the reason above.

That also means the premise in #1853's plan ("probe for the committed head") and
the closing note on #1864 ("that check should not be using the root at all") are
both dead ends: the head is frequently absent or empty, and content is not
identity.

Distinguishing "GC replaced my file" from "my file was renamed away" needs a
**store identity that survives GC**, not a content probe. Two candidates:

1. **Manifest store ID** — 16 random bytes minted at file creation, carried
   forward by GC's rewrite. The manifest has 24 free bytes after `SELF_HASH`
   (144..167 of 168) and `SELF_HASH` already covers new fields. Exact, and the
   check becomes a header read instead of a full store open. Cost: every manifest
   writer must preserve it — the area that produced #1547 — plus a rule for
   legacy stores whose ID is zero.
2. **Generation counter in the graph-lock sidecar** — GC bumps it under the lock
   after `sqlite3OsReplaceFile`; a handle that sees `HAS_MOVED` with an advanced
   generation adopts, otherwise goes read-only. No database-format change, but a
   new use for the sidecar and a defined behaviour when it is missing.

Holding this as draft rather than pushing a guess at which one you want.

The rest of the diff stands on its own and is worth keeping either way: the
backup destination's explicit `adoptReplacement` (backup installs unrelated
content at the destination on purpose and can never be proved by content), and
the corrected `pager4-1.7/1.8` reasons.

timsehn:
Closing this approach. Commit ancestry is not store identity: a clone or diverged copy contains the stale handle's branch tip and is therefore accepted and written through. The candidate is also closed after validation and then reopened by path, leaving a replacement race.

The attached-database seeding defect in #1873 is valid and will be fixed separately from a fresh master branch with direct regression coverage. The remaining foreign-replacement case in #1853 stays open pending an explicit sanctioned-replacement identity/provenance design.

PR #1891, opened 2026-07-31
Reclaim crash garbage on reload; propagate backup init/step OOM failures

timsehn:
Two commits, folded into one PR (the second is what this PR's fault-fuzz job was tripping over).

## 1. Reclaim crash garbage when reloading under the graph lock

A torn-tail recovery (`csReplayWal`) rewinds `nWalData` and the logical EOF **in memory only** — the physical crash garbage stays on disk. `csDetectExternalChanges` treats physical-size > logical-size as a peer append, so after any crash that leaves a torn tail, *every* lock acquisition triggers `csReloadFromDisk`: a full reopen plus WAL replay with blake3 hashing of every chunk body. The reload's replay rewinds in memory again without touching the file, so the condition never clears — every write transaction pays a full O(WAL-since-gc) reload until enough new commits physically outgrow the garbage region (O(WAL²) aggregate for small transactions). The commit-time truncate branch is unreachable on this path: the physical size is only read when the graph lock is *not* already held, and write transactions hold it.

Fix: `csReloadFromDisk` truncates the file to the freshly replayed logical EOF when the caller holds the graph lock and the adopted store is writable and unpoisoned. Under the lock nothing can be mid-append, and the logical EOF was just derived from the file's current contents, so bytes past it are dead by construction. A lost truncate (crash before any sync) just means the garbage reappears and is re-truncated on the next reload.

Test: `corruption_test` appends a 256KB crash tail to a healthy store, runs one write transaction, and asserts the file shrank below the garbage size and the data survived reopen. Fail-before 90/91 → pass-after 91/91.

## 2. Propagate OOM and IO failures in backup init and step

This is the root cause of the `sqlite-regression-fault-fuzz` failure on this PR's first run — a pre-existing master bug that the truncate merely exposed by shifting allocation geometry (macOS reproduces it on unmodified master; the fault-fuzz bucket is ubuntu-only). Pinpointed with the faultsim backtrace hook:

- **`doltliteBackupSameFile` answered from raw filenames on failure**: when its full-path buffers failed to allocate or `xFullPathname` failed, it fell back to `strcmp` of unresolved names — `./x.db` vs `x.db` could pass the same-database check under memory pressure and start a self-overwriting backup.
- **`sqlite3_backup_init` OOM exits set no error code** on `pDest`, unlike stock (which reports `SQLITE_NOMEM`), so OOM-checking callers saw `SQLITE_OK`.
- **`backup_step` voided the stale-tmp pre-delete**, so a real delete failure surfaced later as a confusing EXCLUSIVE-open error and injected faults vanished.

Fix: same-file check failures propagate (OS-layer `SQLITE_IOERR_NOMEM` mapped to plain `SQLITE_NOMEM` to keep stock's init contract); every init OOM exit sets `SQLITE_NOMEM`; the tmp pre-delete propagates everything except `SQLITE_IOERR_DELETE_NOENT`.

Fail-before/pass-after: `backup_malloc` on macOS 15-16 errors → **0 errors** (first clean run on that platform).

## Local validation

- `corruption_test`: 91/91; `backup_malloc`: 0/2184
- `multi_process_test`: 49/49, `multi_process_gc_test`: 101/101, `vc_concurrency_test`: 37/37
- `doltlite_regression_test_c`: 310,021/310,021 (both commits)
- `crash_recovery_test_sqlite_test`: 119/119, `test/crash_injection_test.sh`: 104/104
- `test/run_c_tests.sh` coverage set: 16/16 (both commits)
- `backup2/4/5`, `backup_ioerr`: identical on fixed and unfixed builds (pre-existing macOS-local noise; linux storage bucket gates these)


timsehn:
The sqlite-regression-fault-fuzz failure here is a pre-existing master bug (swallowed OOM/IO failures in backup init/step), root-caused and fixed in #1894 — this PR's file-geometry change merely shifted allocation counts so the ubuntu sweep started reaching the swallow sites (macOS reproduces them on unmodified master). Once #1894 merges, re-running CI here should go green.

PR #1902, opened 2026-07-31
Update pull_reconfirm contract evidence for the atomic ref helper

timsehn:
The concurrency contract (#1898) pinned `src/doltlite_remote_sql.c#chunkStoreReadDiskBranchTip` as evidence for `vc.pull_reconfirm`, and #1900 (merged after) replaced that inline call with `doltliteMutateRefsExpected` — so the selector is stale **on master** and every PR's merge-ref now fails `concurrency_contract_test.sh` in three jobs (Build-Test macos/windows, coverage-native-shell). This is what's currently blocking #1891.

The contract row itself is still true — pull re-reads and compares the on-disk branch tip under the graph lock — the evidence now points at the `doltliteMutateRefsExpected` call in the pull path plus the `chunkStoreReadDiskBranchTip` expectation check inside the helper (`src/doltlite_core.c`).

Local: `concurrency_contract_test.sh` 6/6.


timsehn:
Folded into #1891 per Tim.

PR #1904, opened 2026-07-31
Replay only the appended WAL tail when a refresh sees the file grow

timsehn:
## Problem

A refresh that found the store file grown — every statement-boundary and write-transaction check on any connection after any peer commit — did a full reload: reopen, replay, and blake3-verify **every WAL byte since the last gc**. With user-driven gc that cost scales with history and is paid per transaction per peer append, so two connections alternating single-row commits did quadratic aggregate work. This was the top finding of the performance review (the growth-axis cliff invisible to steady-state sysbench).

## Change

The store is append-only, so a grown file only needs the new tail folded into the current state. `csReplayWal` becomes `csReplayWalRange(startPos)` — the open path passes 0 with fresh state, the refresh passes the committed boundary — and `chunkStoreRefreshIfChanged` uses the range replay for the grew-in-place case. The full reload stays for moved/replaced files, adopt-replacement, and anything the tail replay can't absorb (fall-through on any failure).

The replay runs under the graph lock — a caller without it takes the lock just for the replay (µs, vs the reload's O(WAL)), and a busy lock falls back to today's unserialized reload — so it never reads a writer's in-flight append as a torn tail.

## Numbers

Two connections on separate branches alternating single-row commits after a seeded history (macOS, medians of 3):

| seed rows | before | after |
|---|---|---|
| 100k | 31.4 ms/commit | 0.4 ms/commit |
| 500k | 141.9 ms/commit | 0.4 ms/commit |

Before scales with WAL-since-gc; after is flat (~355× at 500k and growing with history).

## Growth-axis gates (second commit)

So this axis stays fixed, `test/growth_axis_test.sh` + `test/growth_axis_driver.c` gate three growth **ratios** between a seeded history and 5× that history, per-PR in the Build-Test job next to the quick large-scale test (a few seconds of work; the driver links against the `libdoltlite.a` already in the checked-build artifact — no make in the test job). Ratios rather than absolute times, so runner load moves both sides of each gate:

| axis | gate | pre-tail-replay | with tail replay |
|---|---|---|---|
| A: multi-writer refresh | ≤ 3× | **4.4× FAIL** | 1.0–1.3× |
| B: open + first query | ≤ 12× (linear tolerated; open replays the WAL by design) | ~4.9× | ~4.9× |
| B': open after `dolt_gc` vs before | ≤ 1.5× | ~0.01× | ~0.01× |
| C: commit cost, deep-chain last window vs first | ≤ 3× | ~1.0× | ~1.0× |

Axis A failing on the pre-tail-replay engine is the gate's fail-before proof. When the clean-open fast path (review item 3.2) lands, gate B should tighten to flat.

## Latent assumptions surfaced by running replay on live stores (each found by the suites during development)

1. **Replay-state refs snapshot double-free / dead restore** — the snapshot's refs pointers were freed on success and restored on post-adoption failure; correct only when the caller's refs were empty (fresh opens). The snapshot is no longer freed, and incoming refs are validated before the point of no return.
2. **Error path restored pointers but not scalars** — a failed range replay now restores `refsHash`, `nChunks`, `nWalData`, `iFileSize`, and the marker flags, so a peer committing a corrupt refs blob leaves the reader's state intact (caught by `refresh_preserves_refs_hash`).
3. **Adopted refs must be marked committed** — a stale `committedRefsHash` made the next commit's preserve-refs dance restore pre-replay refs over the peer's, intermittently losing branches and commits (caught by `vc_concurrency_test` at ~1-in-8 under looped load; 0/32 after the fix vs 0/12 clean master control).

Plus: tail chunks dedup against the live recent list so gc never streams two entries for one hash; full-replay "fresh empty store" classifications are gated to `startPos==0`; the tail path applies the same benign crash-garbage reclaim as the reload path.

## Tests

- `corruption_test` gains a two-handle case: peer chunks and refs visible through the refresh, repeated increments, and a crashed peer's torn tail rewinding without hiding committed data (tail-path execution verified via a temporary probe).
- Battery: `vc_concurrency_test` ×32 loops clean, `multi_process_test` 49/49, `multi_process_gc_test` 101/101, `multi_process_merge_rebase_test` 20/20, `threadtest5` 8-worker OK, `oom_dolt_fault_test` 13 ops 0 bugs, `doltlite_regression_test_c` 310,044/310,044, `crash_recovery_test_sqlite_test` 123/123, `crash_injection_test.sh` 104/104, `run_c_tests.sh` coverage 16/16, shell battery 92/92, growth gates 4/4.


timsehn:
Converting to draft and splitting.

**Landing separately (green on master):** the growth-axis scaling gates are now #1916, with the multi-writer refresh axis reported-but-ungated (it measures ~65x on master — the very cliff this PR targets) so it can land today and become a gate when the replay work is proven.

**What's left here — and why it isn't ready.** Isolation on a faithful amd64 rig (the earlier arm64 container was lying, and several of my container runs before that were built from a stale amalgamation, which sent me down a wrong path claiming this was a master regression — it isn't):

- current master: savepoint4 0/195; this branch: 26/195, so the regression is mine
- disabling **only** the tail-replay branch, `csReplayWalRange` refactor intact: **0/195** — the refactor is sound, the bug is inside the ~30-line grew-path branch
- the branch is genuinely exercised there: 83 tail replays during the suite

The tail-replay fast path has now produced two distinct correctness failures under crash/fault conditions (the earlier swallowed-OOM/lock-fd issues, now this). That pattern says the design needs a tighter safety envelope before it ships — most likely restricting the fast path to a provably sealed tail and keeping file mutation (the crash-garbage truncate) off it entirely. The performance win is real (355x on the multi-writer axis) and worth finishing, but not worth shipping on a hunch.

timsehn:
Closing. The perf win is real but I can't make this correct with confidence, and the honest call is to stop rather than keep patching.

**Where it ended.** Isolation on a faithful linux/amd64 rig (matching CI):

| configuration | savepoint4 |
|---|---|
| master | 0/195 |
| this branch | 26/195 |
| this branch, tail-replay branch disabled (`csReplayWalRange` refactor intact) | **0/195** |

So the refactor is sound; the fault is confined to the ~30-line fast path in `chunkStoreRefreshIfChanged`, which fires 83 times during that suite. The failure is an `integrity_check` failure after crash recovery with a surviving connection.

**Why I'm not attempting another fix.** Two previous attempts each closed one failure class and opened another (lock-per-refresh broke fd-exhaustion/read-only/shared-cache tests; the lockless version with strict error propagation fixed those and left this). I built a native reproducer — surviving connection, peer commits a real batch, file truncated mid-batch to synthesize a torn tail — and it runs clean at every truncation point, so the actual trigger needs savepoint4's fuller shape (savepoints and rollback inside the crashed transaction). Without a reproducer I understand, another patch would be another guess in the crash-recovery path, which is the last place to guess.

**What a future attempt should do differently.** Take the fast path only when the tail is *provably sealed* — read the trailing root record and verify it seals at its own offset before replaying anything — and fall back to the full reload otherwise. That removes torn-tail handling from the fast path entirely, which is where all three failures have lived. And it should mutate no files: the crash-garbage truncate belongs on the reload path only.

**What survives from this work** (all merged separately): the growth-axis gates (#1916), which measure this exact axis at ~65x on master and are ready to become a hard gate the day a safe fast path lands; the crash-recovery contract fixes and backup OOM propagation (#1891); and the amalgamation-staleness fix (#1919), without which none of the above isolation would have been trustworthy.

Branch kept for reference.

PR #1905, opened 2026-07-31
Gate history-growth scaling in PR CI

timsehn:
## Why

Perf CI measures steady state only (sysbench ratios, absolute wall-clock ceilings). The store's real exposure is the **scaling axis** — how per-operation cost grows with history since the last gc — and nothing gates it. #1904 just improved that axis by ~355×; without a gate the next refactor could quietly give it back.

## What

`test/growth_axis_test.sh` compiles `test/growth_axis_driver.c` against `libdoltlite.a` and gates three growth **ratios** between a seeded history and 5× that history. Runs in the Build-Test job (macos + ubuntu) right next to the quick large-scale test — the quick profile is a few seconds of work. Ratios rather than absolute times, so runner load moves both sides of each gate and it shouldn't flake the way absolute perf ceilings do.

| axis | gate | pre-#1904 | post-#1904 |
|---|---|---|---|
| A: multi-writer refresh (alternating commits, two connections) | ≤ 3× | **4.4× FAIL** | 1.0–1.3× |
| B: open + first query | ≤ 12× (linear tolerated, superlinear fails) | ~4.9× | ~4.9× |
| B': open after `dolt_gc` vs before | ≤ 1.5× | ~0.01× | ~0.01× |
| C: commit cost, last window of a deep chain vs first | ≤ 3× | ~1.0× | ~1.0× |

Axis B intentionally tolerates linear growth — open replays the whole WAL by design today. When the clean-open fast path (review item 3.2) lands, that gate should tighten to flat.

Local repro: `bash test/growth_axis_test.sh build --quick` (or no `--quick` for the 1M-row profile, ~5s).

## Ordering

**Merge #1904 first.** Axis A is calibrated to the tail-replay engine and correctly fails on current master (that's the fail-before evidence above, verified against both builds locally) — including this PR's own Build-Test run until #1904 is in; I'll rerun CI here once it merges.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -210,6 +210,17 @@ jobs:
       run: cd build && bash ../test/large_scale_test.sh ./doltlite --quick
       timeout-minutes: 5
 
+    - name: Growth-axis scaling gates
+      if: matrix.platform != 'windows'
+      # Ratio gates: per-operation cost vs history size (multi-writer
+      # refresh, open latency, commit depth). Ratios, not absolute times,
+      # so runner load moves both sides of each gate.
+      run: |
+        cd build
+        make libdoltlite.a
+        bash ../test/growth_axis_test.sh . --quick
+      timeout-minutes: 10
+
     - name: Chunk size distribution
       if: matrix.platform != 'windows'
       run: cd build && bash ../test/chunk_distribution_test.sh ./doltlite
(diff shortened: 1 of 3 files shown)


timsehn:
Folded into #1904 per Tim — the gate ships with the change it pins, runs per-PR in Build-Test next to the quick large-scale test, and consumes the artifact's libdoltlite.a (no make in the test job).

PR #1925, opened 2026-08-01
Merge SQLite upstream through e12c6f2590 (133 commits)

timsehn:
## Summary
Closes #1912 (SQLite upstream drift watcher).

Merges `sqlite/sqlite` up to `e12c6f259020d11df3672c4fb7939792ec904059` into the doltlite fork (tracked base `f4cf5c9b` was 133 commits behind the current tip; true git history reconnects through `ca8593ce`). Bumps `.sqlite-upstream-base` so the watcher re-zeros.

## Conflicts (5 files)

- **`README.md`** — keep the DoltLite project README
- **`src/memdb.c`** — keep the `DOLTLITE_PROLLY` serialize guard (no page image to dump)
- **`src/test1.c`** — adopt upstream `testStringToPointer`, keep the `DOLTLITE_PROLLY` NULL-page pointer guard
- **`ext/wasm/GNUmakefile`** — take upstream `omit-api=$(omit-api)` and keep the doltlite `__dirname` ESM rewrite
- **`ext/fts5/test/fts5corruptB.test`** — take upstream (adds test 3.x)

## Validation

- Amalgamation regenerates; `doltlite` and `libdoltlite` build
- Smoke: `sqlite_version()` 3.54.0, `doltlite_engine()` prolly, branch/merge works
- Shell: `doltlite_commit.sh` 62/62, `doltlite_branch.sh` 62/62
- C gated suite: **25/25** (`run_c_tests.sh`), including OOM fault sweep with 0 bugs

Full shell/oracle/sqllogictest suites run in CI.

PR #1926, opened 2026-08-01
Keep the graph-lock sidecar open across acquisitions

timsehn:
## Profile first

Sampling 60,000 single-row autocommit commits (6,850 samples) put the cost here:

| cost center | share |
|---|---|
| **`csFileLock` → `sqlite3OsOpenMalloc` → `open()`** | **30.7%** |
| `fsync` | 16.5% |
| `pwrite` | 7.2% |
| `prollyHashCompute` | 5.8% |
| `chunkStorePut` | 4.9% |
| `unixClose` (same sidecar) | 3.2% |
| refs + working-set serialization | 1.2% |

~34% of autocommit CPU was opening and closing the graph-lock sidecar — **about five opens per one-row commit**, each paying a VFS open, a file-object allocation, and two `mprintf`s to re-derive the double-nul name. A `fcntl` lock on an already-open handle is ~0.8µs against ~18µs for the open/lock/unlock/close cycle.

(This corrects an earlier estimate of mine that put the same idea at ~6%; that came from a micro-benchmark using raw POSIX `open()` on a simple path, which isn't what this code does. The profile is the authority.)

## Change

The sidecar handle and its name live on the `ChunkStore` and stay open until the store closes. `pGraphLockFile` still aliases the handle **only while the lock is held** and is NULL otherwise, so all seven `csFileLockHeld()` checks keep their meaning and no caller changes shape.

Safety details:
- **Revalidation**: each acquisition checks `SQLITE_FCNTL_HAS_MOVED` (a stat) before reusing the handle. A sidecar deleted or replaced out from under a live handle would otherwise mean locking a vanished inode and excluding nobody; on mismatch the handle is reopened. Cost is ~1µs against the ~18µs open it replaces.
- **Busy stays busy**: a contended lock returns `SQLITE_BUSY` and keeps the cache; only an unusable handle drops it.
- **Fewer closes is safer**, not riskier, under POSIX's "closing any fd drops all this process's locks on that file" rule.
- Cache released in `chunkStoreClose`, after the lock is dropped.
- `csFileLockNB` had no callers left; removed (dead-code gate).

## Result

Single-row autocommit: **0.30 ms → 0.19 ms (36% faster)**, byte output per commit unchanged.

## Validation

Lock-lifecycle change, so the gauntlet: `chunk_store_fork_lock_test` 14/14, `multi_process_test` 49/49, `multi_process_gc_test` 101/101, `multi_process_merge_rebase_test` 20/20, `vc_concurrency_test` 10 consecutive clean runs, `concurrent_stress_test`, `threadtest4` (245 tests, 0 errors), `threadtest5` 8-worker OK, `crash_recovery_test_sqlite_test` 123/123, `crash_injection_test.sh` 110/110, `corruption_test` 106/106, `doltlite_regression_test_c` 310,064/310,064, `run_c_tests.sh` 16/16, shell battery 93/93, growth gates 3/3, dead-code and orphaned-suite lints clean.


timsehn:
The `vc-fuzz` failure here is the known unnamed-catalog-entry bug class, not this change:

```
FAIL: stateful vc fuzzer seed=1785612785 step=2141 op=cherry_pick
doltlite: catalog invariant violated: entry 7 (unnamed) pairs with no schema row or live meta
```

Three things place it outside this PR:

- The invariant is `#ifdef DOLTLITE_PROLLY_CHECK`-only, and its own comment describes it as a "write-time signature of the unnamed-entry overlay bug class" — the class #1924 is fixing.
- `dolt_cherry_pick` reaches the same machinery #1924 patches: `doltlite_cherry_pick.c` calls `doltliteMergeCatalogs` in `doltlite_merge.c`, and #1924 changes `doltlite_merge.c` and `doltlite_merge_pass2.c` (index adoption / exclusion — precisely what leaves an unnamed entry).
- This PR touches only the graph-lock sidecar handle lifecycle. It has no path into catalog construction, and the fuzzer takes a *different time-based seed every run* — the previous run failed on `op=merge` at a different step and seed, which is the signature of random seeds hitting a standing bug rather than a deterministic regression.

Not reproducible locally even with a checked build (`DOLTLITE_PROLLY_CHECK=1`) at the same seed — the fuzzer's step sequence diverges by platform, so the seed alone doesn't pin the schedule.

Leaving this to #1924 rather than cherry-picking it in, since it's an independent fix with its own reduced regressions and I can't verify the cherry-pick locally. Re-running the job so this PR is judged on a fresh seed.

timsehn:
Converting to draft — the caching changes lock semantics, and I can't validate it outside GHA.

**Evidence it's this change, not the fuzzer noise.** After the swallowed-OOM fix, CI still fails `exclusive` (15 unexpected) and `shared` (85 unexpected) — both locking tests — plus `attachmalloc` and `fkey_malloc`. PR #1925's `storage` bucket passes on the same master, so this isn't an environment-wide problem.

**My earlier "matches master" claim was wrong**, and worth recording why: I compared in a linux container where `exclusive.test` fails ~14 times on *master too* (docker's filesystem doesn't give faithful POSIX locking). A baseline that broken can't detect a delta, so "mine == master == 14" meant nothing. GHA disagrees, and GHA is the rig that counts.

**The likely mechanism**: keeping the sidecar `sqlite3_file` open changes how SQLite's unix VFS tracks locks per inode (`unixInodeInfo` shared/pending counts persist instead of being torn down at close), so conflict detection between connections in one process no longer behaves the way fresh-open-per-acquisition did. `shared.test`'s 85 failures point squarely at that.

**Why I'm stopping rather than iterating**: the only faithful rig is CI, at ~20 minutes a cycle, and this is the third lock/replay change today where local validation said green and linux said otherwise. A 36% autocommit win doesn't justify guessing at lock correctness.

**If someone picks this up**, the shape that avoids the semantic change is caching only what isn't lock state: the resolved sidecar *path* and its double-nul buffer (kills two `mprintf`s per acquisition) while still opening the fd per acquisition. That's a much smaller win — the 27% was the `open()` itself — but it's semantics-preserving. Anything that keeps the fd alive needs a real answer for `unixInodeInfo` accounting across connections, verified against `exclusive.test` and `shared.test` on linux.

Profile data is in the PR description and stands on its own: the sidecar open/close really is ~34% of autocommit CPU, so the win is there for whoever solves the locking question properly.

timsehn:
Closing — the cost isn't worth it, and the gate list makes that clear.

**Final diagnosis** (the earlier lock-semantics theory was wrong): a purpose-built reproducer — two connections in one process, A holding a write txn, B required to be refused, across repeated acquire/release cycles — passes identically on master and on this branch. Mutual exclusion was never broken. What actually fails is open-file accounting: `exclusive-5.2`, `exclusive-5.7`, `stmt-1.2/1.5/1.7`, `shared-1.4.1.1` all assert stock's exact `sqlite_open_file_count`, and a cached sidecar takes doltlite from 1 to 2 fds per connection at rest.

**Why gating was the wrong escape hatch.** Beyond those 6-7 semantic assertions, the extra open shifts allocation geometry enough to move OOM sweep indices: `attachmalloc-1.transient.907`, `fkey_malloc-7.transient.194-196/473-475`, `sysfault-2.2-vfsfault-persistent.104/105`. Pinning fault-injection iteration numbers into the divergence list is precisely what this repo avoids — they aren't divergences, they're drift, and they would rot on the next unrelated allocation change.

**Why the win can't be had cheaply.** The ~5 sidecar opens per one-row commit are 5 *separate* acquire/release bursts (`lockDepth` already collapses reentrant ones), so caching within a burst buys nothing. Holding across releases is the entire mechanism, and that is what costs the descriptor.

**The profile stands and is the useful artifact**: sidecar `open()`/`close()` really is ~34% of autocommit CPU (fsync 16.5%, pwrite 7.2%, hashing 5.8%). Anyone wanting that 36% back should go after the descriptor itself rather than caching it — taking the graph lock as a byte-range lock on the main store file, the way SQLite locks its own database, removes the second file entirely: no extra fd, no extra open, and the parity tests stay untouched. That's a locking-protocol change with a cross-process compatibility story to write, so it deserves its own design rather than being smuggled in behind a perf patch.

PR #1936, opened 2026-08-03
Run VACUUM and backup on legacy databases through the stock engine

timsehn:
Follow-up to #1934, closing the gap that PR deliberately left open.

## Problem

A stock-format file opened by doltlite routes to the orig B-tree engine, but the three maintenance surfaces didn't follow it there:

| Operation on a stock-format file | Before | After |
|---|---|---|
| `VACUUM` | reports success, file unchanged (344064 bytes) | rewrites pages, 8192 bytes — byte-identical to stock |
| `VACUUM INTO` | `VACUUM INTO is not supported for doltlite databases` | stock-format copy, rows verified |
| `.backup` / `sqlite3_backup_*` | `source database is not backed by a file` | page copy, integrity `ok` |
| `.restore` | same failure | works |

## Why it wasn't already done

`tool/mksqlite3c.tcl` excluded stock `backup.c` from the amalgamation because `findBtree()` reads `Db.pBt` expecting the stock `Btree` layout, which can't coexist with prolly's in one translation unit. The page copier lives in that file, so the single-file build stubbed `orig_sqlite3_backup_init` to return 0 and had no `orig_sqlite3BtreeCopyFile` at all.

Only that one line actually touches `struct Db` — every other `->pBt` in the file is `Btree->pBt` (BtShared), fine inside the rename scope. And reading `Db.pBt` as a stock btree was already wrong: it is a doltlite wrapper in **every** build, with the stock btree hanging off `pOrigBtree`. `DOLTLITE_ORIG_BACKUP` redirects that line through `doltliteBtreeOrigPtr()`, defined outside the rename scope, which lets `backup.c` join the orig block. Both flavors now share one page copier — verified below.

## Changes

- **`sqlite3BtreeCopyFile`** delegates to the orig engine when both btrees are legacy. Without it the stock vacuum path segfaults in `invalidateCursors(pTo->pBt==0)` the moment it stops being dead code.
- **`sqlite3_backup_init`** returns a `DoltliteBackup` wrapper holding the delegate instead of the stock handle. `backup_step`/`finish`/`remaining`/`pagecount` cast their argument to `DoltliteBackup` unconditionally, so the pre-existing non-main delegation was handing a foreign pointer into those casts.
- **`vacuum.c`** dispatches on the format of the database named by `iDb`, not `aDb[0]`.
- **`doltlite_engine=sqlite`** URI parameter: auto-detect reads a file header, so it can't classify a file that doesn't exist yet, and a new database is always doltlite-format. The knob selects the engine at creation and is **ignored once the file has content**, so it can never reinterpret an existing chunk store. `VACUUM` applies it to its own target; `.backup` applies it when the source is legacy (escaping `%`, `?`, `#` in the path).
- Mixed pairs stay refused — no defined conversion between a chunk store and stock pages, so it errors rather than half-copying.

## Tests

`test/doltlite_open_sqlite_file.sh` (the legacy-format suite) gains 22 assertions: **74 passed, 0 failed**. Destination content always differs from the source, so a no-op or partial copy cannot pass — the backup case starts with 900 rows in the destination and 7 in the source.

Build-flavor parity checked explicitly, since that was the whole blocker:

```
object build      : VACUUM 344064 -> 8192, rows 100, integrity ok
amalgamation      : VACUUM 344064 -> 8192, rows 100, integrity ok
object build      : backup dest 900 rows -> 7 rows, integrity ok
amalgamation      : backup dest 900 rows -> 7 rows, integrity ok
```

Also covered: `.backup` to a brand-new path yields a stock file (250 rows verified); a path containing `#` and `?` survives escaping; a doltlite source still backs up to a doltlite-format file (`doltlite_engine()` = `prolly`); a new file without the knob is still doltlite-format.

## Validation

- `test/run_doltlite_tests.sh`: **99/99 suites**
- `doltlite_regression_test_c`: **310064/310064**
- `make lint`, `test/dead_code_check.sh`: pass
- Divergence/inventory/ratchet gates: pass. Two entries became stale and were retired from all three places — `vacuum5-1.3.4` ("VACUUM size metric") and `e_vacuum-2.1.8` ("file not rebuilt smaller"), i.e. exactly the assertions this fixes. Ratchet lowered gates 4955 → 4953, intentional 3613 → 3611.
- tcl `vacuum{,2,4,5,6}`, `e_vacuum`, `vacuum-into`, `vacuummem`, `backup{,2,4,5}`, `backup_ioerr`, `backup_malloc`, `attach{,2,3}`, `pager1`, `wal`, `savepoint`: failing-assertion sets compared against a control build of the merge base in the same rig — **no regressions, two improvements** (the two entries above).
- No new compiler warnings from doltlite sources (the amalgamation's `-Wcomment` banner warnings are pre-existing; the base emits the identical 17).

### DIFF
--- a/README.md
+++ b/README.md
@@ -640,6 +640,24 @@ CREATE TABLE local_events AS SELECT * FROM ops.events;
 DETACH DATABASE ops;
 ```
 
+Auto-detect reads an existing file's header, so it cannot classify a file that
+does not exist yet: a database created by DoltLite is DoltLite-format. To create
+a stock SQLite file instead, open it with `doltlite_engine=sqlite`:
+
+```
+doltlite 'file:/path/to/new.sqlite?doltlite_engine=sqlite'
+```
+
+The parameter selects the engine for a database being created and is ignored
+once the file has content, so it can never reinterpret an existing database.
+`.backup` and `VACUUM INTO` apply it for you when the source is a stock file, so
+their output is a stock file too.
+
+`VACUUM` on a stock database rewrites pages as SQLite does; on a DoltLite
+database it garbage-collects unreachable chunks. `.backup`/`.restore` and
+`sqlite3_backup_*` work within either format, but not between them — there is no
+defined conversion, so a mixed pair is refused rather than half-copied.
+
 ## Per-Session Branching Architecture
 
 Each connection selects a branch independently and recovers that branch's
--- a/src/backup.c
+++ b/src/backup.c
@@ -105,7 +105,11 @@ static Btree *findBtree(sqlite3 *pErrorDb, sqlite3 *pDb, const char *zDb){
     return 0;
   }
 
+#ifdef DOLTLITE_ORIG_BACKUP
+  return (Btree*)doltliteBtreeOrigPtr((void*)pDb->aDb[i].pBt);
+#else
   return pDb->aDb[i].pBt;
+#endif
 }
 
 /*
--- a/src/backup_orig.c
+++ b/src/backup_orig.c
@@ -1,5 +1,8 @@
 /*
-** Compile the original SQLite backup.c with renamed symbols.
+** Compile the original SQLite backup.c with renamed symbols. DOLTLITE_ORIG_BACKUP
+** makes findBtree() resolve schema names to the stock btree behind doltlite's
+** Btree wrapper, since Db.pBt is a wrapper in every build.
 */
+#define DOLTLITE_ORIG_BACKUP 1
 #include "btree_orig_prefix.h"
 #include "backup.c"
--- a/src/btree_orig_api.c
+++ b/src/btree_orig_api.c
@@ -235,6 +235,12 @@ int origBtreeIsSqliteFile(sqlite3_vfs *pVfs, const char *zFilename,
   return SQLITE_OK;
 }
 
+#ifndef SQLITE_OMIT_VACUUM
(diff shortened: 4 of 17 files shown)


timsehn:
Pushed a guard on the URI lookup this PR adds to `sqlite3BtreeOpen()`.

`sqlite3_uri_parameter()` resolves its argument through `databaseName()`, which reads `zName[-1]` through `zName[-4]` to walk back over the URI's key/value pairs — so it is only defined on a name `sqlite3ParseUri()` built, and reads before the buffer if handed a plain nul-terminated string.

Main and ATTACH are the only opens whose names come from ParseUri. The earlier short-circuits happen to cover every other caller today, so the call was safe in practice, but only as a property of the current call sites rather than anything stated. It is now gated on `SQLITE_OPEN_MAIN_DB`.

No behavior change, verified on all four paths the knob serves: `doltlite_engine=sqlite` still selects the stock engine on a main open and through ATTACH, a plain path still creates a doltlite database, and `.backup` of a legacy database to a new path still produces a stock file. Suites: 74/74, 10/10, 10/10.

To be clear about what this is not: it is **not** a fix for the earlier `corrupt-3.3` red. That assertion passes on the rebased run, and I never reproduced it in five rigs including CI's exact clang/-O1/coverage configuration. This is hardening of a hazard the hunt surfaced, landed here because the call does not exist on master.

PR #1949, opened 2026-08-04
TEST ONLY: isolate backup.c-in-amalgamation (corrupt-3.3 attribution)

timsehn:
Do not merge. Attribution experiment for the `corrupt-3.3` red on #1936.

Clean master plus **only** the change that emits stock `backup.c` into the amalgamation's orig block (and the `doltliteBtreeOrigPtr` helper its guarded `findBtree` needs to compile). None of #1936's VACUUM, backup-wrapper, URI-knob or shell changes.

Master is green on the storage bucket across #1942, #1945 and #1947, so if this branch turns `corrupt-3.3` red, this hunk is the cause. testfixture builds from the amalgamation, which is why it is the leading suspect.

Will close once it reports.

### DIFF
--- a/src/backup.c
+++ b/src/backup.c
@@ -105,7 +105,11 @@ static Btree *findBtree(sqlite3 *pErrorDb, sqlite3 *pDb, const char *zDb){
     return 0;
   }
 
+#ifdef DOLTLITE_ORIG_BACKUP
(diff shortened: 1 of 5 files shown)


timsehn:
This was superceded by other PRs

PR #1951, opened 2026-08-04
vc-fuzz: reset --hard aborts on the catalog pairing invariant (seed 1785800828)

timsehn:
Workspace for the vc-fuzz red on #1936. **No fix yet** — this is a handoff with the investigation so far. Branch is clean off master (empty commit).

## The failure

```
FAIL: stateful vc fuzzer seed=1785800828 step=1731 op=reset_hard
reset_b104 failed rc=-6
```

`rc=-6` is SIGABRT — `dolt_reset('--hard')` **crashes**, it does not merely diverge. The abort message:

```
doltlite: catalog invariant violated: entry 10 (aux_1704) pairs with no schema row or live meta
```

## Where

`src/prolly_btree_catalog.c:1302`, in the `DOLTLITE_PROLLY_CHECK` block of `doltliteSerializeCatalogEntriesForBtreeImpl`. The check's own comment names the class:

> an entry that pairs with no schema row and no live meta has lost its identity crossing catalog numbering domains … Both are write-time signatures of the unnamed-entry overlay bug class — fail fast instead of publishing the catalog. Only CONSTRUCTED arrays (staging, merge, reset) are checked.

So this fires by design: a constructed catalog array reached serialization with an entry it could not identify. Note it only aborts under `DOLTLITE_PROLLY_CHECK` (which the vc-fuzz job's checked build sets). **Without that flag the catalog is published anyway**, which per the same bug class is how tables silently vanish — so the release-build consequence is likely worse than a crash, not better.

## The constructed array

`doltlitePreserveUntrackedTablesOnHardReset()` in `src/doltlite_reset.c` builds the array. It builds **from the target catalog**, appends untracked entries renumbered past the target range, and then serializes via:

```c
doltliteSerializeCatalogEntriesForeignDomain(
    db, aTarget, nTarget, aWorkSchema, nWorkSchema, &buf, &nBuf);
```

Note the asymmetry: entries come from the **target**, schema rows come from the **working** catalog. Pairing in the serializer tries `oldPg == iTable`, then a name match restricted to `type=='table'` rows, then `findCatalogEntryMetaByPgno`. Target entry numbers are in a different domain from working rootpages, so target entries depend on the name fallback — which only covers tables, and only tables still present in the working schema.

`aux_1704` is a table name (the fuzzer's `ddl` op names tables `aux_<step>` and indexes `aux_idx_<step>`), and entry number 10 is small enough to be a target entry rather than one of the renumbered appended ones.

## What I ruled out

The obvious reading of the above — a tracked table dropped in the working tree, so absent from `aWorkSchema` — **does not reproduce**:

```
committed: aux_x, keep
working:   DROP TABLE aux_x; CREATE TABLE untracked_u;   (untracked_u forces the preserve path)
reset --hard -> aux_x, keep, untracked_u    aux_x row intact
```

That case pairs correctly, so the trigger needs something more specific. The `ddl` op also does `rename_table`, `drop_index` and `create_index`, and an untracked **index** looks like a real hole in the name fallback (it matches `type=='table'` only), so a renamed table or an index entry is where I would look next.

## Reproducing

Deterministic given the seed; needs a `DOLTLITE_PROLLY_CHECK=1` build for the invariant to fire, and reaches step 1731 in roughly 10–15 minutes:

```
make DOLTLITE_PROLLY_CHECK=1 doltlite
DOLTLITE_VC_STATEFUL_SEED=1785800828 DOLTLITE_VC_STATEFUL_SECONDS=1800 \
  DOLTLITE_VC_STATEFUL_KEEP_DB=1 bash ../test/vc_stateful_fuzzer.sh ./doltlite
```

`KEEP_DB=1` leaves the database at the failure point, and `DOLTLITE_VC_STATEFUL_TRACE=1` prints every statement — grepping the trace for `b104` gives the exact operation history on the affected branch, which is what cracked the previous fuzz bug (#1943).

## Related

Prior art on this bug class: constructed catalog arrays must pair the master root by `iTable==1` rather than by name, since the master entry is unnamed — otherwise the number-first serializer maps entries onto the wrong rows.


timsehn:
Resolved by #1950. This draft documents the same seed 1785800828 and reset-time symptom; #1950 fixes the earlier staged-commit catalog corruption that causes it.

timsehn:
**Reproduced locally, and it is deterministic.** Same seed, same step, same branch, same entry:

```
FAIL: stateful vc fuzzer seed=1785800828 step=1731 op=reset_hard
reset_b104 failed rc=-6
doltlite: catalog invariant violated: entry 10 (aux_1704) pairs with no schema row or live meta
```

Recipe that produced it (checked build, ~13 min to reach step 1731 on an M-series Mac):

```
make DOLTLITE_PROLLY_CHECK=1 doltlite
PYTHONUNBUFFERED=1 DOLTLITE_VC_STATEFUL_SEED=1785800828 \
  DOLTLITE_VC_STATEFUL_SECONDS=1800 DOLTLITE_VC_STATEFUL_KEEP_DB=1 \
  bash ../test/vc_stateful_fuzzer.sh ./doltlite
```

So it does not need CI to chase. Two notes for whoever picks it up:

- `DOLTLITE_PROLLY_CHECK=1` is required — the invariant is inside that guard. Without it the unpairable catalog is published instead of aborting.
- Add `DOLTLITE_VC_STATEFUL_TRACE=1` and grep the trace for `b104` to get the operation history on the affected branch. That is what reduced the previous fuzz bug from a 4271-step run to a four-line repro in #1943, and it should show what `aux_1704` went through (the `ddl` op does create/rename/drop table and create/drop index) before the `reset --hard`.

One hypothesis already eliminated: the plain case of a tracked table dropped in the working tree, with an untracked table present to force the preserve path, restores correctly and does not trip the invariant. So the trigger is narrower than that.

PR #2023, opened 2026-08-07
Support table-level checkout of virtual tables

timsehn:
## Why

Every table-level `dolt_checkout` form failed on a virtual table with "no such branch or table":

- `SELECT dolt_checkout('ft')` — discard working changes to an fts table
- `SELECT dolt_checkout('side','ft')` — pull one table's state from another branch
- `SELECT dolt_checkout('HEAD~1','ft')` — restore a dropped vtab from history

Found by the new vtab×VC battery (separate PR). Loud failure, no corruption — but these workflows were impossible for fts3/4/5, rtree, and extension vtabs.

## What

`doltliteCheckoutTables` resolves names against catalog `TableEntry`s, and virtual tables have none — their schema row is storage-free (rootpage 0) and their content lives in the shadow tables. Three changes in `src/doltlite_checkout.c`:

- Validation accepts a name whose source schema row is a `CREATE VIRTUAL TABLE` statement.
- The catalog-entry swap skips the vtab's own name — the existing schema pass already handles the drop/recreate of the vtab row itself.
- New `checkoutAdoptVtabShadows` swaps the shadow tables' catalog entries from the source, mirroring how staging carries shadows (`sqlite3IsShadowTableOf` over the source and working schema lists). When the schema pass rebuilt the vtab, the live shadows are fresh trees, so their table numbers are kept while the content roots adopt the source — the same discipline the plain-table rebuilt path uses.

## Testing

- Three new scenarios in `test/doltlite_index_vc_matrix.sh` (discard, cross-branch, restore-dropped; fts5 + rtree with module self-checks). All three fail on unfixed master (40/43), all pass with the fix (43/43).
- Full regression c-test suite: 310230/310230.
- Shell battery: 99/99 suites.

### DIFF
--- a/src/doltlite_checkout.c
+++ b/src/doltlite_checkout.c
@@ -661,6 +661,93 @@ static void checkoutAdoptSourceIndexRoots(
   }
 }
 
+/* Virtual tables persist through their shadow tables: a table-level
+** checkout of a vtab must swap the shadows' catalog entries, the same way
+** staging carries them. The vtab's own master row is schema-only and is
+** handled by the schema pass; when that pass rebuilt the vtab, the live
+** shadows are fresh trees whose table numbers must be kept while their
+** content roots adopt the source. */
+static int checkoutAdoptVtabShadows(
+  sqlite3 *db,
+  struct TableEntry **paWorking, int *pnWorking,
+  struct TableEntry *aSource, int nSource,
+  SchemaEntry *aWorkSchema, int nWorkSchema,
+  SchemaEntry *aSourceSchema, int nSourceSchema,
+  const char *zVtab,
+  int vtabRebuilt
+){
+  Table *pTab = sqlite3FindTable(db, zVtab, "main");
+  SchemaEntry *aList[2];
+  int aN[2];
+  int pass, i, j;
+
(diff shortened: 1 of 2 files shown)


timsehn:
@itoqa Both findings triaged:

**CATALOG-3 (stale shadow-table indexes): confirmed, fixed in #2026.** Reproduced your fixture exactly — the shadow-adoption pass swapped storage roots but never reconciled named-index DDL on shadow tables, so a working-only index survived with a mismatched tree (the post-reopen `row N missing from index` failure) and a source-only index was never created. #2026 runs the existing per-table index reconcile for each source-side shadow during the schema pass, with your fixture as the regression test (fail-before/pass-after).

**CHECKOUT-5 (savepoint sealing): pre-existing platform contract, not introduced by this PR.** Plain-table checkout — a path this PR did not touch — behaves identically on master (`ROLLBACK TO sp` → "no such savepoint"). And Dolt itself behaves identically: `SAVEPOINT sp; CALL DOLT_CHECKOUT('t'); ROLLBACK TO sp` reports `SAVEPOINT sp does not exist`. Working-set/ref VC operations seal the enclosing transaction by design, the same contract as `dolt_commit` and `dolt_merge`. This PR only made virtual-table names reach the same shared path that plain tables already took.

PR #2030, opened 2026-08-08
Absorb vec1 derived-shadow conflicts into the merge and rebuild in place

timsehn:
## What

Phase 3 of the vector plan: `dolt_merge` on PQ-configured vec1 tables is now fully automatic. A vec1 table's `%_idx`/`%_meta`/`%_model`/`%_config` shadows are rebuildable from `%_base` and the stored model, so their row conflicts carry nothing a rebuild doesn't regenerate. When **every** conflict in a merge is such a shadow and the owner's `%_base` still holds raw vectors, the merge:

1. keeps ours' rows (mechanically identical to what `dolt_conflicts_resolve('--ours')` leaves — the merged roots already carry ours' values, resolution is artifact-removal),
2. surfaces **zero** conflicts,
3. rebuilds each owner from the merged base and stored model while the merged catalog is live (same slot as the existing adopted-index reindex, so the flush captures the rebuilt shadow roots), and
4. proceeds as a clean merge commit.

**Derived shadows merge by rebuild; the authoritative shadow merges by row.**

## Guardrails (each pinned in the suite)

- **Uncompressed builds stay loudly conflicting.** Their `%_base` holds bucket numbers, not vectors, so auto-resolution would silently lose the discarded side's data. Gate: `typeof(vector)` over the base rows.
- **Any non-derived conflict disables the automation for the whole merge** — a user-table conflict keeps every conflict, shadows included, reported for manual resolution.
- **`dolt_merge` only**: cherry-pick and rebase pass a null rebuild list and keep today's behavior, conservatively.

The policy is a small registry in the catalog merge (vec1 suffixes + the model-based rebuild statement); the new `doltliteMergeCatalogs` out-params are nullable so other callers are untouched.

## Testing

- `test/doltlite_vec1_vc.sh` reworked and extended (18 checks): the PQ merge scenarios now assert the fully automatic flow (merge returns a hash, `dolt_conflicts` empty, both branches' vectors searchable in a fresh session, merge commit at head), plus the flat-config loud-conflict guard and the mixed-conflict manual-fallback guard. Fail-before: exactly the two automation scenarios fail without the engine change (16/18); all 18 pass with it.
- Full regression c-test suite: 310258/310258. Shell battery: 101/101 suites.

With this, the arc from billy's original question is closed: vec1 syntax built in (#2027), lossless PQ merge semantics pinned (#2029), and merges that need no manual steps at all (this PR). `docs/vec1-native-design.md` updated to reflect all three phases as shipped.


timsehn:
@itoqa CATALOG-2 confirmed and fixed in #2031. Your analysis was exactly right: the rebuild statement's `(SELECT val FROM "%_model" WHERE id=1)` degrades to NULL when the row is missing, and vec1 treats a NULL rebuild argument as keep-the-current-model — so the automation could commit a merge whose index was rebuilt against cached state. The fix extends the eligibility gate: an owner without a present, non-empty model blob at id=1 is ineligible for automation and the merge keeps its conflicts loud for manual resolution, the same conservative fallback as uncompressed builds and mixed conflicts. Your reproduction is pinned as `vec1_missing_model_stays_loud` (fails on unfixed code by silently auto-merging; loud with the fix). The malformed-blob variant was already covered: vec1's decoder errors, the merge fails, and the saved transaction state restores before anything is committed or staged.

PR #2041, opened 2026-08-08
Resume merged scans from the recorded key after a deferred write

timsehn:
## Bug

A one-pass DELETE or UPDATE silently skips rows whenever the row it just wrote was pending (mut-map-sourced). Both repros fail on master and commit the wrong result:

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, v TEXT);
INSERT INTO t VALUES (10,'ten'),(20,'twenty');
BEGIN;
INSERT INTO t VALUES (1,'a'),(2,'b'),(3,'c'),(4,'d'),(5,'e');
DELETE FROM t WHERE id < 3;      -- row 2 survives
UPDATE t SET v='UPD' WHERE id < 4;  -- only row 1 updated (analogous shape)
COMMIT;
```

## Cause

The deferred write paths deactivate the merged scan state (`mmActive = 0`). The Next/Prev re-seed then trusted a still-valid tree cursor as the logical position — but when the departed row was mut-map-sourced, the tree side is legitimately parked at the *next committed row*, so the step skipped every pending row between the departed key and the parking spot. The `cachedIntKey` breadcrumbs that record the true position were unreachable because the tree-valid seed branch took precedence, and the plain-intkey deferred delete recorded no breadcrumb at all (only AUXDELETE/shadow roots did).

## Fix

- The deferred intkey delete now always records the hole (`cachedIntKey` + `BTCF_DeleteKey`), mirroring the tombstone-parking the non-intkey path already does. The AUXDELETE/shadow gate (and the now-unused `cursorIsShadowTableRoot`) go away.
- `prollyBtCursorNext`/`Previous` consult the breadcrumb before the tree cursor: `resumeDeactivatedMergedScan` re-seeks both sides strictly past `cachedIntKey` and lets `mergeScan` pick the next live row. The tree payload is cached only for pure tree landings — on a BOTH landing the mut-map entry shadows the tree row and must serve the pending value.

## Validation

- 3 new tests in `test/doltlite_txn_seek_visibility.sh` fail on a master-built control binary and pass with the fix (suite 15/15).
- Full shell battery: 101/101 suites. Regression c-test: 310,258/310,258.
- 300-seed randomized differential sweep (mixed pending INSERT/DELETE/UPDATE + forward/backward scans) vs stock SQLite 3.54: 0 regressions vs master, 24 seeds newly correct. 7 seeds still mismatch identically on master and this branch — that is a distinct pre-existing defect (`materializeDeferredMergedSeek`/`Backward` cache the stale tree payload on BOTH landings), fix coming in its own PR.

### DIFF
--- a/src/prolly_btree_cursor.c
+++ b/src/prolly_btree_cursor.c
@@ -633,7 +633,11 @@ int prollyBtCursorNext(BtCursor *pCur, int flags){
     rc = prollyCursorApplyMergeStep(pCur, 1);
   }else if( pCur->pMutMap && !prollyMutMapIsEmpty(pCur->pMutMap)
          && pCur->curIntKey
-         && (pCur->curFlags & (BTCF_ValidNKey|BTCF_DeleteKey)) ){
+         && (pCur->curFlags & (BTCF_ValidNKey|BTCF_DeleteKey))
+         && !(prollyCursorIsValid(&pCur->pCur)
+              && prollyCursorIntKey(&pCur->pCur)==pCur->cachedIntKey) ){
+    /* A tree cursor parked on the logical key seeds the fast branch below;
+    ** parked anywhere else it would skip the pending rows in between. */
     rc = ensureCursorMutMapOrder(pCur);
     if( rc!=SQLITE_OK ) return rc;
     rc = resumeDeactivatedMergedScan(pCur, 1);
@@ -691,7 +695,9 @@ int prollyBtCursorPrevious(BtCursor *pCur, int flags){
     rc = prollyCursorApplyMergeStep(pCur, -1);
   }else if( pCur->pMutMap && !prollyMutMapIsEmpty(pCur->pMutMap)
          && pCur->curIntKey
-         && (pCur->curFlags & (BTCF_ValidNKey|BTCF_DeleteKey)) ){
+         && (pCur->curFlags & (BTCF_ValidNKey|BTCF_DeleteKey))
+         && !(prollyCursorIsValid(&pCur->pCur)
+              && prollyCursorIntKey(&pCur->pCur)==pCur->cachedIntKey) ){
     rc = ensureCursorMutMapOrder(pCur);
     if( rc!=SQLITE_OK ) return rc;
     rc = resumeDeactivatedMergedScan(pCur, -1);
--- a/src/prolly_btree_cursor.c
+++ b/src/prolly_btree_cursor.c
@@ -392,6 +392,60 @@ static int seedMutMapIterFromCursor(
   return SQLITE_NOTFOUND;
 }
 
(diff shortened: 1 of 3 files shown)


timsehn:
## Root cause of the CI failures — pre-existing, not a defect in this change

I reproduced `savepoint4-2.8.1.2` locally and root-caused it. Summary of the evidence:

**The failure is a pre-existing crash-semantics property of the chunk-store commit, surfaced by this PR only because the fixed scan consumes the SQL `random()` stream differently, steering the test's random walk into a vulnerable state.**

The proof chain (all deterministic, from a captured pre-crash db state):

1. The failing state replays 10/10 identically on **master** and this branch — the captured `pre_crash.db` + the same crashsql invocation loses the same row on both engines.
2. The recovered "wrong" state is **byte-identical to the state the transaction commits** when run to completion under the crash VFS (`1 18ae536e...`). No data is lost and recovery is correct: the child reached `RELEASE one` (COMMIT), the crash fired *during the commit batch's sync*, and the deterministic crash simulator happened to apply the entire batch — indistinguishable from a completed sync, so replay rightly treats the transaction as committed.
3. Stock SQLite can never produce this outcome under crashsql: its rollback-journal commit point (journal delete) lies strictly *after* the last `test.db` sync, and crashsql only crashes **on** syncs — so `savepoint4`/`crash2` hard-code "any crash → pre-transaction state". Doltlite's commit point *is* a sync of `test.db`, so "crash at the commit sync with the batch surviving" is a legitimately-committed outcome the stock tests read as corruption.

Master passes these suites today only because its particular random walk + WAL layout never lands on that coin flip (the crash VFS PRNG is seeded, so it's deterministic per layout). Any engine change that shifts write layouts or `random()` consumption can re-roll it — this will keep biting unrelated PRs until the policy question is settled.

`crash2-2.17.2–2.24.2` are the same class. `mmap1` is not verifiable on macOS (65/85 errors on stock master locally — env-dependent read counts); the follow-up commit removes nearly all of this PR's extra tree seeks (the plausible trigger for the mmap1 count drift), so CI will adjudicate it.

**Decision needed before this can merge** (options):
- Port `savepoint4`/`crash2` coverage to doltlite-owned recover variants that accept *either* the pre-transaction or the committed post-transaction state after a crash (matches SQLite's actual atomic-commit contract; precedent: the `doltlite_recover_*` ports).
- Or change the commit protocol so the commit-effective write happens after the final sync returns (journal-delete emulation) — shrinks but does not eliminate the window, with real durability tradeoffs.
- Or curate these tests out of the must-pass bucket per the crash-list promotion protocol.

Repro artifacts (captured state + standalone replay script) are preserved locally; happy to attach them to a dedicated issue.

PR #2057, opened 2026-08-09
Scale VC performance ceilings by measured runner fsync cost

### LINKED ISSUE #2055: Nightly performance regression
- **Workflow:** `Nightly performance`
- **Run:** https://github.com/dolthub/doltlite/actions/runs/31306619792
- **Commit:** `ca4d77c6e0a9974d8a25827dac0e3433cc72e90e`
- **Ref:** `master`

Nightly performance failed or produced incomplete results.

### vc

```text
Preparing version-control benchmark fixtures...
2 version-control benchmark(s) exceeded their ceiling.
```

timsehn:
Fixes the nightly failure in #2055 (`checkout_branch_clean` 156.68/150, `merge_schema_no_conflicts` 41.78/35).

## Diagnosis: not an engine regression

Both suspects from yesterday's changes were ruled out on the same rig:

1. **Timing control**: 7 paired-run medians of the full VC benchmark suite, current master vs the ceiling-retune commit (`d3377db944`) as baseline binary — `checkout_branch_clean` 0.956x, `merge_schema_no_conflicts` 0.981x, `branch_create_delete` 1.024x. Master is not slower.
2. **Sync-count control** (rig-independent): both binaries instrumented at `csSyncFile` — identical counts (checkout×2: 5, merge_schema: 4, branch create+delete: 3) at both reduced and full 400×250 fixture scale.

What the failing run actually shows, against the healthy 8/8 nightly: CPU-bound reads got ~40% **faster** (status 83→48ms, diffs 90→53ms, conflict merges 128→59ms) while the fsync-bound write paths got 1.5–2.8x **slower** (checkout 56→157, merge_schema 23→42, branch create/delete 25→37). That's a runner with a faster CPU and a much slower disk — not a code change, and not something absolute-millisecond ceilings can distinguish from a regression. The Aug 7 retune left only ~1.5x headroom on the tightest write benchmarks, so this class of runner fails them.

## Fix

Before the benchmarks run, an engine-independent probe (15× 8KiB `pwrite`+`fsync` on the benchmark disk, median) measures the runner's fsync cost, and ceilings scale by `max(1, probe/reference)`, capped at 5x (`VC_PERF_IO_REF_US`, default 2500; `VC_PERF_IO_SCALE_MAX`). A healthy runner keeps the exact tightened ceilings — the retune's intent is preserved; a degraded disk widens them proportionally. The probe can't be gamed by engine changes since it never touches doltlite.

The probe value is printed in both output modes (nightly ceiling reports and PR relative runs), so the 2500µs reference can be validated against this PR's own CI output and retuned from nightly history. Local verification: healthy disk → probe 74µs, scale 1.00, table unchanged; forced-low reference → scale engages and effective ceilings render in the table.

Closes #2055.

### DIFF
--- a/test/vc_perf_ceiling.sh
+++ b/test/vc_perf_ceiling.sh
@@ -304,16 +304,49 @@ bench_sql() {
     printf 'vc\t%s\t%s\t%s\n' \
       "$name" "$baseline_median" "$candidate_median" >> "$RESULTS_FILE"
   else
+    local eff_ceiling_us
+    eff_ceiling_us=$(python3 -c "print(int($ceiling * 1000 * $IO_SCALE))")
     status="PASS"
-    if [ "$candidate_median" -gt $((ceiling * 1000)) ]; then
+    if [ "$candidate_median" -gt "$eff_ceiling_us" ]; then
       status="FAIL"
       failures=$((failures+1))
     fi
-    rows="${rows}| $name | $(python3 -c "print(f'{$candidate_median/1000:.2f}')") | $ceiling | $status |\n"
+    rows="${rows}| $name | $(python3 -c "print(f'{$candidate_median/1000:.2f}')") | $(python3 -c "print(f'{$eff_ceiling_us/1000:.0f}')") | $status |\n"
     printf 'vc\t%s\t%d\t%s\n' \
-      "$name" "$((ceiling * 1000))" "$candidate_median" >> "$RESULTS_FILE"
+      "$name" "$eff_ceiling_us" "$candidate_median" >> "$RESULTS_FILE"
   fi
 }
+
+# Absolute ceilings are meaningless on a runner whose disk is degraded: the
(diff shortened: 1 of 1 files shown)


timsehn:
Both review findings addressed in the latest commit:

- **Transient probe failure** (medium): the probe now retries transient `OSError`s briefly and, if it still can't measure, reports `unavailable` and falls back to unscaled ceilings instead of failing the run — the probe is an assist, not a gate, and a persistently erroring disk will surface through the benchmarks themselves exactly as it did before this PR. The retry wrapper's terminal-on-producer-error semantics are left as-is since `benchmark_retry_test.py` pins them deliberately.
- **Small fixture runs** (minor, pre-existing): the dirty-data/dirty-schema generators now bound their table ranges by `VC_PERF_TABLES`; verified `VC_PERF_TABLES=8` relative mode runs end-to-end with paired samples.

PR #2088, opened 2026-08-10
Gate pull requests on the differential sweep

### LINKED ISSUE #2076: In-transaction reads and writes diverge from stock SQLite on non-intkey tables (~3% of differential seeds)
Found by the randomized SQL differential sweep added in the PR that introduces `test/sql_differential_test.sh`. Measured on master `5be08e306e`: **5 divergent seeds of 200**, then **16 of 500** on a fresh window — roughly 3%, so this is systemic rather than a few unlucky seeds.

Every reproducer has the same shape: committed rows, `BEGIN`, one or more writes, then a read or a second write in the same transaction. All of them are on **non-intkey** tables — composite PK, `WITHOUT ROWID`, `UNIQUE` index, or a collated column. The intkey merged-cursor path has had three fixes (#2041, #2042, and the `deferredTreeSeek` resume fix in #2062); the blob/index-key arm of that same state machine appears not to have been covered by any of them.

These are almost certainly more than one root cause and should be fixed in separate PRs. Two of them lose or corrupt data.

Reproduce each with the stock reference built from this tree so the SQLite version matches:

```
make DOLTLITE_PROLLY=0 sqlite3
```

### 1. `database disk image is malformed` (seed 138)

An `UPDATE` whose range is empty (`k > 8 AND k < -3`), after an in-transaction `DELETE`:

```sql
CREATE TABLE t(k NUMERIC PRIMARY KEY, a, b TEXT) WITHOUT ROWID;
CREATE INDEX i_a ON t(a);
INSERT OR REPLACE INTO t(k, a, b) VALUES(-17, -33.132, x'0001');
INSERT OR IGNORE INTO t(k, a, b) VALUES(-38, -44.638, 'AB ');
BEGIN;
DELETE FROM t WHERE b > 'a';
UPDATE t SET a = 27 WHERE k > 8 AND k < -3;
```

doltlite: `Error near line 7: database disk image is malformed` (rc 1). stock: clean (rc 0).

### 2. A `DELETE` matching nothing removes every row (seed 17)

```sql
CREATE TABLE t(k INTEGER UNIQUE, a, b TEXT COLLATE RTRIM);
CREATE INDEX i_b ON t(b, a);
INSERT OR IGNORE INTO t(k, a, b) VALUES(-5, -25.110, x'0001');
INSERT OR IGNORE INTO t(k, a, b) VALUES(3, 33, '');
BEGIN;
UPDATE t SET b = 'a' WHERE b > 'AB ';
DELETE FROM t WHERE b > 'zz';
SELECT group_concat(q, '|') FROM (SELECT quote(b) AS q FROM t ORDER BY b, k);
```

doltlite: empty (both rows gone). stock: `''|'a'`. No row has `b > 'zz'`, so the `DELETE` should be a no-op.

### 3. One-pass `UPDATE` skips a committed row after an in-transaction insert (seed 194)

```sql
CREATE TABLE t(k INTEGER, j TEXT, a, b TEXT, PRIMARY KEY(k, j)) WITHOUT ROWID;
INSERT OR ROLLBACK INTO t(k, j, a, b) VALUES(29, 'z', -27, x'00');
INSERT OR ROLLBACK INTO t(k, j, a, b) VALUES(-19, x'00', NULL, x'00');
BEGIN;
INSERT OR IGNORE INTO t(k, j, a, b) VALUES(21, 'AB ', 23, 'A');
UPDATE t SET b = 'b' WHERE k > -20 AND k < 32;
SELECT group_concat(q, '|') FROM (SELECT coalesce(quote(k), 'NK') || '/' || coalesce(quote(a), 'NA') || '/' || coalesce(quote(b), 'NB') AS q FROM t ORDER BY 1);
```

doltlite: `-19/NULL/'b'|21/23/'b'|29/-27/X'00'`. stock: `...|29/-27/'b'`. All three rows are in range; row 29 is not updated. This is the #2041 shape on the blob-key path.

### 4. Correlated subquery over uncommitted state (seeds 124, 143)

```sql
CREATE TABLE t(k INTEGER, j TEXT COLLATE NOCASE, a, b TEXT, PRIMARY KEY(k, j)) WITHOUT ROWID;
INSERT OR ROLLBACK INTO t(k, j, a, b) VALUES(-29, 'A', 'A', 'zz');
BEGIN;
INSERT OR IGNORE INTO t(k, j, a, b) VALUES(-33, x'ff', 27, 'a');
DELETE FROM t WHERE k > 20 AND k < 37;
SELECT count(*) FROM t WHERE a IN (SELECT a FROM t WHERE k = -29);
```

doltlite: `0`. stock: `1`.

```sql
CREATE TABLE t(k INTEGER UNIQUE, a, b TEXT);
CREATE INDEX i_b ON t(b, a);
INSERT OR ROLLBACK INTO t(k, a, b) VALUES(12, NULL, 'A');
BEGIN;
UPDATE t SET b = 'ab' WHERE k > -15 AND k < 29;
UPDATE t SET a = x'ff' WHERE b > x'0001';
SELECT count(*) FROM t WHERE a IN (SELECT a FROM t WHERE k > -38 AND k < 26);
```

doltlite: `1`. stock: `0`.

### Remaining work on the sweep itself

The nightly `differential` job in `nightly-fuzz.yml` runs this now and will keep reporting into this thread. The short per-PR run is deliberately **not** wired up yet, because at a 3% divergence rate it would block every pull request. Adding it is the last step here, once the divergences above are fixed:

- [ ] fix the divergences above (separate PRs)
- [ ] add the per-PR differential smoke run and remove the nightly-only note in `nightly-fuzz.yml`
- [ ] enable the `DOLTLITE_DIFF_LARGE_INTS` axis (needs the >2^53 prefix-equality fix)
- [ ] enable the `DOLTLITE_DIFF_DESC` axis (needs the DESC sort-key ordering fix)

timsehn:
Closes #2076.

Rebased and reduced. The axis defaults this PR originally carried were superseded by #2090's group mechanism, and its version of the reference guard by #2095, so what is left is the thing that was actually missing: **the sweep runs on every pull request.**

## The change

One step in `test.yml`'s `oracle-tests` `sql` bucket, which already unpacks a stock reference into the same directory as the engine — so no new artifact plumbing, and it sits with the other vs-stock suites.

It sweeps a **fixed** seed window (1–400), so a red is reproducible from the seed alone rather than depending on when it ran. The nightly job keeps its moving window for breadth; the comment there now says which is which.

## Why this is safe to gate on now

When the sweep landed in #2078 it was nightly-only on purpose: the engine diverged on roughly 3% of seeds and a gate would have blocked everything. Those divergences were five distinct bugs, all since fixed:

| bug | fix |
|---|---|
| numeric byte-prefix read as an equal key | #2075 |
| non-intkey one-pass write stepped the tree off a pending row | #2082 |
| `mergeLast` landings only steppable backwards (three symptoms) | #2079 |
| range seeks searched the pending map by exact key, hiding uncommitted rows | #2085 |
| DESC numeric sort keys mis-ordered on disk | #2087 |
| a `>=` bound found no pending landing | #2092 |

And the reference it compares against is now built where it cannot inherit doltlite's storage, and asserted (#2095) — without that this gate would have passed vacuously.

## Validation

The exact window the gate runs, against the default group set, locally: **400/400 clean**. The default set is `large-ints` + `desc`; the feature groups added in #2090 stay opt-in while #2093 (`ddl` plus `triggers`) is open, so the gate does not depend on them.

### DIFF
--- a/.github/workflows/nightly-fuzz.yml
+++ b/.github/workflows/nightly-fuzz.yml
@@ -180,9 +180,8 @@ jobs:
   # one runs generated SQL through both engines and compares answers, so it
   # covers query and transaction semantics instead of parsing.
   #
-  # Nightly only for now. The engine currently diverges on roughly 3% of seeds
-  # (issue #2076), so a per-PR gate would block every pull request; adding the
-  # short per-PR run is the last step of that issue.
+  # The per-PR gate lives in test.yml and sweeps a fixed seed window; this job
+  # sweeps a moving one for breadth, so the two do not re-check the same seeds.
   #
   # Each night sweeps a different seed window derived from the run number, so
   # coverage compounds instead of re-checking seeds earlier runs already
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -253,6 +253,16 @@ jobs:
         done
       timeout-minutes: 10
 
+    # A fixed seed window, so a red here is reproducible from the seed alone
+    # rather than depending on when it ran. The nightly job sweeps a moving
+    # window for breadth; this one is the per-PR gate.
+    - name: Randomized SQL differential sweep (vs stock SQLite)
+      if: matrix.bucket == 'sql'
+      run: |
+        cd build-coverage
+        bash ../test/sql_differential_test.sh ./doltlite ./sqlite3-stock 1 400 || exit 1
+      timeout-minutes: 12
+
     - name: Upload oracle coverage profiles
       if: always()
       uses: actions/upload-artifact@v4


timsehn:
Two CI failures, one mine and one not.

**Mine — the reference guard was asking the wrong question.** It rejected any reference exposing `dolt_version`, and CI's `sqlite3-stock` does expose it even though it is built from this tree with the prolly storage compiled out. Whether those functions are linked in says nothing about which storage the binary uses, and only the storage has to differ for the comparison to mean anything. Fixed in the commit above: the guard now creates a database and requires an SQLite header, which is exactly the property that matters — stock writes one, doltlite writes a chunk store. Verified both ways locally: a genuine stock binary passes, and passing `./build/doltlite` as the reference is still refused.

Worth recording why I chased this rather than just loosening the check: a `sqlite3` built in a directory that already has prolly objects *can* come out prolly-enabled, which would make every vs-stock oracle compare the engine against itself and pass regardless. I reproduced CI's build order in a clean directory (`make DOLTLITE_PROLLY_CHECK=1 doltlite`, then `make DOLTLITE_PROLLY=0 sqlite3`, then the extra `doltlite-lib`/`testfixture` targets) and the resulting binary is genuinely stock every time — no `dolt_version`, and it writes `SQLite format 3` files. So the existing oracle suites are comparing against real stock; the format check now proves that on every run instead of assuming it.

**Not mine — `Test Suite / scaling`:** `FAIL: 500-row commit, 10000000 vs 1000000 rows (423ms / 18ms = 23.5x > 8x)`, a timing-ratio ceiling. This PR only touches `test/sql_differential_test.sh` and two workflow files, so it cannot affect commit latency; that is the known perf-ceiling flake class.

**Marking this draft until #2087 merges.** It flips the DESC axis on by default, and the DESC mis-ordering that axis detects is fixed by #2087 — so with #2087 unmerged, this correctly goes red. Merge #2087 first and this should come up green; happy to rebase and re-run it then.

timsehn:
The rerun failed again, and the guard is now telling us something real rather than being wrong: CI's reference **does not write SQLite-format databases**, so it shares doltlite's storage. Filed as #2094.

That makes the existing vs-stock oracle suites in `test.yml` — `sql_oracle_test.sh` (568 assertions) and every `oracle_*_test.sh` — comparisons of the engine against itself, which would explain why none of the five storage bugs fixed this week were caught by CI.

This PR stays blocked until #2094 is fixed. That is the gate working as intended: it refuses to run against a reference that cannot disagree with the engine. I would rather it stay red than pass vacuously.

For the record, I could not reproduce the bad reference locally — `make DOLTLITE_PROLLY=0 sqlite3` yields a genuinely stock binary in every directory ordering I tried, including CI's. The evidence is CI's own artifact. #2094 has the details and a suggested fix, the important half being to assert the property in `ci-build.yml` at build time so a bad reference fails the build that made it instead of quietly voiding the oracles.

PR #2095, opened 2026-08-10
Assert that the stock reference does not share doltlite's storage

### LINKED ISSUE #2094: CI's sqlite3-stock is not stock: the vs-stock oracle suites compare doltlite to itself
The differential sweep's reference guard failed in CI on #2088, and the reason is worse than the guard being wrong.

## What CI reports

`Test Suite / oracle-tests-sql`, on the `sqlite3-stock` binary unpacked from the `coverage-build` artifact:

```
ERROR: ./sqlite3-stock does not write SQLite-format databases, so it shares
       doltlite's storage and cannot be the reference
```

The guard creates a database with that binary and checks the file header. Stock SQLite writes `SQLite format 3`; doltlite writes a chunk store. CI's reference fails that check, and it separately reports `dolt_version` in `pragma_function_list`.

## Why this matters beyond my PR

`test.yml` runs the vs-stock oracles against that same binary:

```yaml
bash ../test/sql_oracle_test.sh ./doltlite ./sqlite3-stock || exit 1
for f in ../test/oracle_*_test.sh; do bash "$f" ./doltlite ./sqlite3-stock || exit 1; done
```

If `sqlite3-stock` uses doltlite's storage, then **those suites are comparing the engine against itself** — 568 assertions in `sql_oracle_test.sh` plus every `oracle_*_test.sh`, all passing regardless of whether the storage layer is correct. They would not have caught any of the five storage bugs fixed this week (#2075, #2079, #2082, #2085, #2087), which matches the observation that none of those were caught by CI.

## What I could and could not reproduce

Locally, `make DOLTLITE_PROLLY=0 sqlite3` produces a genuinely stock binary: no `dolt_version`, and it writes `SQLite format 3` files. I tried to reproduce CI's condition in clean directories with CI's ordering — `make DOLTLITE_PROLLY_CHECK=1 doltlite`, then `doltlite-lib`, `testfixture`, `doltlite-c-tests-build`, `doltlite-regression-test-c-build`, then `make DOLTLITE_PROLLY=0 sqlite3` — and the result was stock every time. So I have not pinned the mechanism; the evidence that it is wrong comes from CI itself.

The likeliest mechanism is still that the `sqlite3` target is satisfied by something built earlier with the prolly define, so the `DOLTLITE_PROLLY=0` invocation finds it up to date and does nothing, and the packaging step copies that. The coverage job's `CFLAGS` (`-fprofile-instr-generate -fcoverage-mapping`) and `-j2` are the remaining differences from my local runs.

## Suggested fix

Two parts, and the second matters more than the first:

1. Build the reference so it cannot inherit prolly state — its own configure directory, or `rm -f sqlite3` plus `make -B DOLTLITE_PROLLY=0 sqlite3` — and make it a distinct target name so it can never be confused with the doltlite CLI of the same name.
2. **Verify it at build time.** Assert the file-header property in `ci-build.yml` before packaging the artifact, so a reference that shares doltlite's storage fails the build that produced it rather than silently making every oracle suite vacuous. The check is three lines:

```sh
./sqlite3-stock /tmp/ref.db "CREATE TABLE x(y);"
test "$(head -c 15 /tmp/ref.db)" = "SQLite format 3"
```

Worth also confirming what `sqllogictest`'s reference is, since `ci-build.yml` builds a `sqllogictest-stock` in the same directory and the same reasoning applies.

## Status

#2088 is blocked on this: it adds the differential sweep as a per-PR gate, and the gate correctly refuses to run against a reference that shares the engine's storage. It stays red until the reference is fixed, which is the right behaviour.

timsehn:
Fixes #2094. Pre-existing, and not something the build ever said out loud — the differential sweep's guard is what surfaced it.

## The problem

Every vs-stock oracle, the differential sweep, and the SQLite performance baselines compare doltlite against a binary built from this tree with `DOLTLITE_PROLLY=0`. If that binary ends up built *with* the storage in, it agrees with doltlite by construction: the comparison passes no matter what breaks, and nothing anywhere reports it. CI's packaged `sqlite3-stock` failed the sweep's guard on #2088, which is what started this.

The blast radius is every consumer:

| consumer | reference | if it shares the storage |
|---|---|---|
| `sql_oracle_test.sh` (568 assertions) + 18 `oracle_*_test.sh` | `sqlite3-stock` | vacuous, always passes |
| differential sweep (per-PR and nightly) | `sqlite3-stock` | vacuous |
| `nightly-performance`, `release` benchmarks | `sqlite3.o` | the "SQLite" baseline is doltlite, so the numbers are meaningless |

Note the benchmarks link **`sqlite3.o`**, not the binary — a stale object compiled with the storage in is enough to poison them, which is why the fix removes both.

## The fix

1. **Rebuild rather than possibly reuse.** `rm -f sqlite3` (and `sqlite3.o` for the benchmark jobs) before each `make DOLTLITE_PROLLY=0` invocation, so an existing artifact can never be mistaken for a freshly-defined one.
2. **Assert it at the point it is created**, via a new `test/assert_stock_reference.sh`. The test is *what the binary writes*: stock lays down an `SQLite format 3` header, doltlite a chunk store. On failure it prints the actual header bytes and says what to do.
3. **Assert it where it is consumed too**, because that is where a bad reference does its damage — one step in `test.yml` covers every vs-stock suite in the `sql` bucket.
4. The sweep now **defers to the shared check** instead of carrying its own. Its old guard asked whether `dolt_version` was linked in, which is the wrong question: those functions can be present while the storage is stock, and that combination is harmless to a comparison. That false negative is what made #2088 red.

Sites covered: all three `ci-build.yml` builds, the coverage artifact both before and after packaging (stripping is the last thing that touches it), `nightly-fuzz.yml`, `nightly-performance.yml`, `release.yml`, and the consuming job in `test.yml`.

`sqllogictest-stock` needs nothing: it is compiled from upstream's own `sqlite3.c` *before* doltlite's amalgamation is copied over it. That ordering is load-bearing and now has a comment's worth of company in the same file.

## Bugs this uncovered: none, and that is worth stating

You asked me to fix whatever the corrected reference turned up in the same PR. I ran every vs-stock suite against a genuinely stock binary built in a clean directory, and they all pass:

```
sql_oracle_test.sh                              568 passed, 0 failed
oracle_attach                22/0    oracle_blob_composite_pk_vc   31/0
oracle_dot_commands          61/0    oracle_foreign_keys           45/0
oracle_fts5                   7/0    oracle_generated_columns      37/0
oracle_generated_columns_vc  24/0    oracle_index_merge            43/0
oracle_large_blobs           37/0    oracle_null_index_merge       11/0
oracle_reindex               42/0    oracle_savepoints             43/0
oracle_schema_change_cp_rv   23/0    oracle_temp_tables            23/0
oracle_triggers              27/0    oracle_upsert                 35/0
oracle_wide_table_vc         33/0    oracle_without_rowid          39/0
```

So the gap was **latent**: those curated assertions do hold against real stock, and the damage was that they could not have caught a future regression. That also matches the week's history — the five storage bugs fixed in #2075, #2079, #2082, #2085 and #2087 were all found by the randomized sweep, none by these suites.

## What I could not determine

I could not reproduce the bad reference locally. `make DOLTLITE_PROLLY=0 sqlite3` produces a genuinely stock binary in every ordering I tried, including CI's exact sequence in a clean directory, and I confirmed no doltlite target creates `sqlite3` as a byproduct on macOS. The dependency graph may differ on Linux. Rather than keep guessing, this makes the property explicit and checked, so if it is ever wrong again the build that produced it fails with a clear message instead of quietly voiding a few thousand assertions.

#2088 should rebase on this: it carries an earlier version of the same guard fix, which becomes redundant.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -419,7 +419,9 @@ jobs:
         find . -type f -name '*.profraw' -delete
 
     - name: Assert the stock reference does not share doltlite's storage
-      run: bash test/assert_stock_reference.sh build-coverage/sqlite3
+      run: |
+        bash test/assert_stock_reference.sh \
+          build-coverage/sqlite3 build-coverage/doltlite
 
     - name: Package
       run: |
@@ -438,7 +440,8 @@ jobs:
         # The consumers compare against this copy, so assert the copy: it is
         # what ships, and stripping it is the last thing that touches it.
         bash test/assert_stock_reference.sh \
-          coverage-artifact/build-coverage/sqlite3-stock
+          coverage-artifact/build-coverage/sqlite3-stock \
+          coverage-artifact/build-coverage/doltlite
         tar -C coverage-artifact -czf coverage-build.tar.gz build-coverage
 
     - name: Upload
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -81,7 +81,6 @@ jobs:
         {
           make -j2 DOLTLITE_PROLLY_CHECK=1 \
             doltlite doltlite-remotesrv doltlite-lib
-          rm -f sqlite3
           make DOLTLITE_PROLLY=0 sqlite3
           make libsqlite3.a USE_AMALGAMATION=0
           rm -f sqlite3.c .target_source
@@ -401,7 +400,6 @@ jobs:
           make -j2 DOLTLITE_PROLLY_CHECK=1 \
             doltlite doltlite-remotesrv doltlite-lib testfixture \
             doltlite-c-tests-build doltlite-regression-test-c-build
-          rm -f sqlite3
           make DOLTLITE_PROLLY=0 sqlite3
           blake_objects=(blake3.o blake3_portable.o blake3_dispatch.o)
           for object in blake3_sse2.o blake3_sse41.o blake3_avx2.o blake3_avx512.o blake3_neon.o; do
@@ -418,10 +416,8 @@ jobs:
         fi
         find . -type f -name '*.profraw' -delete
 
-    - name: Assert the stock reference does not share doltlite's storage
-      run: |
-        bash test/assert_stock_reference.sh \
-          build-coverage/sqlite3 build-coverage/doltlite
+    - name: Build the stock reference
+      run: bash test/build_stock_reference.sh build-stockref build-coverage/doltlite
 
     - name: Package
       run: |
@@ -434,7 +430,7 @@ jobs:
           build-coverage/blake3*.o \
           build-coverage/*_test \
           coverage-artifact/build-coverage/
-        cp build-coverage/sqlite3 \
+        cp build-stockref/sqlite3 \
           coverage-artifact/build-coverage/sqlite3-stock
         llvm-strip --strip-debug coverage-artifact/build-coverage/*
         # The consumers compare against this copy, so assert the copy: it is
@@ -577,7 +573,6 @@ jobs:
         {
           make CFLAGS="$ASAN_CFLAGS" LDFLAGS="$ASAN_LDFLAGS" \
             doltlite libdoltlite.a
-          rm -f sqlite3
           make DOLTLITE_PROLLY=0 sqlite3
           cc $ASAN_CFLAGS -I. -I../src \
             -o doltlite_regression_test_c \
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -81,6 +81,7 @@ jobs:
         {
           make -j2 DOLTLITE_PROLLY_CHECK=1 \
             doltlite doltlite-remotesrv doltlite-lib
+          rm -f sqlite3
           make DOLTLITE_PROLLY=0 sqlite3
           make libsqlite3.a USE_AMALGAMATION=0
           rm -f sqlite3.c .target_source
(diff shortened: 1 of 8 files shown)


timsehn:
Root cause found, and the CI failure on the previous push was the assertion doing its job.

`tool/mksqlite3c.tcl` bakes `#define DOLTLITE_PROLLY 1` into the amalgamation on purpose, so compiling `sqlite3.c` without the flag still yields doltlite. Once a directory has generated that file for `doltlite` or `testfixture`, `make DOLTLITE_PROLLY=0 sqlite3` is only stock if it happens to regenerate the amalgamation — which is the difference between CI and my machine, and why I could not reproduce it locally. CI proved it on a binary built seconds earlier in the same step:

```
ERROR: build-coverage/sqlite3 does not write an SQLite database header.
```

`-DDOLTLITE_PROLLY=0` would not fix it either, since the code tests `defined(...)`.

The reference now comes from `test/build_stock_reference.sh`, which builds in a directory that has never built doltlite and asserts the result there. Benchmarks take `sqlite3.o` from the same place, since their SQLite baseline links that object and would otherwise be another doltlite.

I also backed out my earlier `rm -f sqlite3` in the checked and sanitizer jobs. A build directory's own `sqlite3` is *supposed* to be doltlite-flavoured — the inherited `shell*.test` suites run it, and swapping in a stock binary locally took `core-sql` from 53 to 140 unexpected failures. Only the reference changes.

Two smaller fixes from the #2090 review are also here: the check now requires the reference's SQLite version to match the engine's (a 3.45 binary passed the storage check and then reported version skew as divergence), and the sweep keeps a script for every failing seed rather than the first five.

Issue #2093, opened 2026-08-10
Schema change plus a trigger in one transaction diverges from stock


timsehn:
Found by the differential sweep once #2092 cleared the last single-group divergence. Every one of the twelve feature groups in #2090 is clean on its own over 200 seeds, but running them together still diverges on roughly 9% of seeds — so this needs two features at once.

## Which two

All-minus-one over 120 seeds, with every group enabled except the named one:

```
minus large-ints     4 failed        minus joins        4 failed
minus desc           8 failed        minus writesel     4 failed
minus expr          10 failed        minus ddl          0 failed   <--
minus agg           12 failed        minus constraints  7 failed
minus setops         9 failed        minus triggers     0 failed   <--
minus cte           10 failed
minus window        10 failed
```

Dropping `ddl` or `triggers` makes it clean; dropping anything else does not. So it is schema-change-inside-a-transaction combined with triggers, and nothing else is required.

`ddl` emits `ALTER TABLE ADD COLUMN` (sometimes with a `DEFAULT`), `CREATE INDEX` / `DROP INDEX`, `REINDEX` and `ANALYZE` from inside the transaction body, plus partial and expression indexes and a view. `triggers` creates `AFTER INSERT`/`UPDATE`/`DELETE` triggers writing a side table, and upserts.

## Symptom

Seed 6 with all groups. The trigger side table ends with rows stock never produced:

```
doltlite: 3='i'|42='u'|43='u'|44='u'|45='u'|47='u'|49='u'|4='i'|56='u'|5='i'|7='i'
stock:    3='i'|42='u'|49='u'|4='i'|5='i'|7='i'
```

The `AFTER UPDATE` trigger is `INSERT OR IGNORE INTO u VALUES(length(quote(NEW.a)) + 40, 'u')`, so the extra rows mean it fired for rows stock did not update, or fired with `NEW` values stock never wrote. A trigger firing for the wrong row set is the same shape as the one-pass scan bugs, but reached through the trigger program rather than the outer statement.

Reproduce:

```
python3 test/sql_differential_fuzzer.py 6 --include-large-ints --include-desc \
  --include-agg --include-setops --include-cte --include-window --include-joins \
  --include-writesel --include-ddl --include-constraints --include-triggers
```

[...]

Issue #2081, opened 2026-08-10
vc_ref_mutation_stress_test loses merged rows under TSan timing (intermittent)


timsehn:
Seen on the Sanitizers / tsan job of PR #2078, whose diff is two new test files plus a nightly workflow job and cannot affect this. The step is "Run version-control concurrency under TSan".

Despite the job name, **this is not a TSan race report** — no race was printed. `vc_concurrency_test` passed 37/37, then `vc_ref_mutation_stress_test` failed two row-count assertions:

```
=== VC Ref Mutation Stress Test ===
FAIL: verify_all_rows_merged
FAIL: verify_worker_rows_merged

Results: 280 passed, 2 failed out of 282 tests
```

Both are the final-state counts in `verifyFinalState` (`test/vc_ref_mutation_stress_test.c:300-304`):

```c
check("verify_all_rows_merged",   rc==SQLITE_OK && count==1 + N_WORKERS*N_ROUNDS);
check("verify_worker_rows_merged", rc==SQLITE_OK && count==N_WORKERS*N_ROUNDS);
```

Each worker loops `commitBranchRow` → `mergeBranchToMain` → `dolt_checkout('main')` → `deleteBranch`. Ending below the expected count means **a concurrent merge lost rows**: main advanced to a root that did not contain a peer's already-merged row. No worker reported an error (the workers print to stderr and would have failed the run earlier), so the merges all believed they succeeded.

That makes it the #1098 family — a multi-step VC op that must re-confirm head under the lock immediately before advancing the ref — but on the merge path rather than the ref-advance primitive, which was previously audited as going through `doltliteCompareAndAdvanceBranch`.

## Reproduction status

Not reproduced locally yet: 25 runs plain and 40 runs with `DOLTLITE_PROLLY_CHECK=1` were all clean on macOS. TSan slows every thread substantially, which widens whatever window this needs, so a TSan build is probably required to hit it — or an injected delay between the merge's head read and its ref advance.

Rare in CI: the last 25 `sanitizers.yml` runs were otherwise clean and the same job passed on PR #2075 at a comparable base, so this is intermittent rather than a standing red.

## Why this matters more than the flake

The assertion that failed is a **data-loss** assertion, not a timing assertion. Either concurrent merges can silently drop a peer's rows, or the test's expectations are wrong about what concurrent merges guarantee. Both need answering, and the first would be serious.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

timsehn:
[...]

PR #2102, opened 2026-08-11
Fuzz RETURNING, generated columns and foreign-key actions nightly

timsehn:
Every bug this sweep has found lived in the same shape: **a write with an effect the statement does not spell out.** So far that shape was only reachable through triggers. Three other features have it and none were being generated.

| group | what it adds |
|---|---|
| `returning` | `RETURNING` on a write — the statement reads back the row it just produced |
| `generated` | a `VIRTUAL`/`STORED` generated column, sometimes indexed — a column recomputed from a row the same statement is rewriting |
| `fkeys` | a child table with `ON DELETE/UPDATE CASCADE\|SET NULL\|RESTRICT` and foreign keys on — a delete that writes another table |

`returning` is deliberately held to a single row: SQLite promises no order for `RETURNING` rows and it cannot carry an `ORDER BY`, so a multi-row `RETURNING` would report row order as a difference. `fkeys` ends on `PRAGMA foreign_key_check`, which has to agree too.

## The bigger finding: nothing ran the wider surface

The nightly took `sql_differential_test.sh`'s **default** group set, same as the pull-request gate. The default is the two value axes. So `ddl`, `triggers`, `expr`, `agg`, `setops`, `cte`, `window`, `joins`, `writesel` and `constraints` were generated by **no job at all** — they existed only for someone to run by hand. This PR points the nightly at every group.

Pull requests stay on the two value axes: they are the cheapest per seed, and a gate that can block unrelated work should be the narrow one. A nightly divergence becomes an issue instead of a block, which is what the wider surface is worth.

## The reproduce hint was wrong twice over

The divergence issue the nightly files told you to run:

```
make DOLTLITE_PROLLY=0 sqlite3
bash test/sql_differential_test.sh ./doltlite ./sqlite3 <seed> <seed>
```

Both lines were broken. It omitted the group selection — without it the generator emits a **different script for the same seed**, so a reported seed does not reproduce. And `make DOLTLITE_PROLLY=0 sqlite3` is the recipe that can reuse an amalgamation carrying the storage and hand back a "reference" that agrees with doltlite by construction. It now uses `build_stock_reference.sh`, which builds where doltlite never was and asserts what the binary wrote.

## Testing

- Each new group alone, through the shipped runner: `returning` 150/150, `generated` 150/150, `fkeys` 150/150.
- All fifteen groups together: 300/300.
- The wider hunt behind them: 2600 seeds all-groups + 6000 seeds default-axis, both clean.
- Group-name validation still rejects an unknown group.

## Merge order

Merge **after #2099**. All-groups is clean only with that fix; without it the nightly would file an issue for #2093, which #2099 fixes.

### DIFF
--- a/.github/workflows/nightly-fuzz.yml
+++ b/.github/workflows/nightly-fuzz.yml
@@ -225,7 +225,11 @@ jobs:
         LAST=$(( FIRST + WINDOW - 1 ))
         mkdir -p "$GITHUB_WORKSPACE/diff-artifacts"
         cd build-diff
-        if DOLTLITE_DIFF_SAVE_DIR="$GITHUB_WORKSPACE/diff-artifacts" \
+        # Every feature group, not the pull-request default. This is the only
+        # place the wider SQL surface runs, and a divergence here becomes an
+        # issue rather than a block, which is what that surface is worth.
+        if DOLTLITE_DIFF_GROUPS=all \
+           DOLTLITE_DIFF_SAVE_DIR="$GITHUB_WORKSPACE/diff-artifacts" \
            bash ../test/sql_differential_test.sh \
              ./doltlite ./sqlite3-stock "$FIRST" "$LAST" \
              2>&1 | tee "$GITHUB_WORKSPACE/diff-artifacts/sweep.log"; then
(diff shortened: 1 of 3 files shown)


timsehn:
Merged current master (for #2099 and #2100) and #2101's branch into this one, so CI re-runs with the `vc_ref_mutation_stress_test` fix in place. The `Sanitizers / tsan` failure here was #2081, not this PR — it changes no engine code, and that job passed on #2100 in the same window while failing twice in a row here.

**Do not merge this before #2101.** It currently carries #2101's four engine files (`doltlite_core.c`, `doltlite_internal.h`, `doltlite_merge_cmd.c`, `prolly_btree_mutation.c`) so tsan can go green; merging this first would land those changes through the wrong PR. Once #2101 is in, those commits become common history and this PR's diff goes back to just the generator, its runner, and the nightly workflow.

timsehn:
#2101 is merged, so its commits are now common history and this PR's diff is back to just the generator, its runner, and the nightly workflow. The earlier caution about merge order no longer applies.

Issue #2103, opened 2026-08-11
ROLLBACK TO a savepoint leaves the index entry for the rolled-back row behind


timsehn:
Found by the randomized differential sweep at seed 510519 with every feature group enabled (the group set #2102 adds; the `generated` group produced the script, but a generated column is **not** required to reproduce).

`ROLLBACK TO` a savepoint fails to undo an `INSERT ... RETURNING`. The row survives the rollback, the index is left inconsistent with the table, and the next read that goes through the index fails with `database disk image is malformed`. `PRAGMA integrity_check` reports `ok` throughout.

## Reproducer

```sql
CREATE TABLE t(k INTEGER, a, b TEXT DEFAULT 'dflt', g AS (length(coalesce(quote(b),'')) + coalesce(a, 0)) VIRTUAL);
CREATE INDEX i_a ON t(a);
INSERT OR IGNORE INTO t(k, a, b) VALUES(34, -18, x'00');
INSERT OR REPLACE INTO t(k, a, b) VALUES(0, 'AB ', 'z');
INSERT OR REPLACE INTO t(k, a, b) VALUES(-27, 'A', 'a' || char(0) || 'b');
INSERT OR IGNORE INTO t(k, a, b) VALUES (34, 'a', 'ab'), (18, 20.369, 'z');
INSERT OR ROLLBACK INTO t(k, a, b) VALUES(9007199254740993, -46.284, 'A');
INSERT OR IGNORE INTO t(k, a, b) VALUES(38, -27, 'a');
INSERT OR ROLLBACK INTO t(k, a, b) VALUES(-22, -45.950, 'A');
ANALYZE t;
BEGIN;
DELETE FROM t WHERE b IN (SELECT b FROM t WHERE k = -25);
SAVEPOINT sp2;
INSERT OR REPLACE INTO t(k, a, b) VALUES(9007199254740992, '', 'a' || char(0) || 'b') RETURNING coalesce(quote(k),'N'), coalesce(quote(a),'N'), coalesce(quote(b),'N');
ANALYZE t;
ROLLBACK TO sp2;
SELECT group_concat(coalesce(quote(a), 'N'), '|') FROM (SELECT a FROM t ORDER BY a ASC, coalesce(quote(k), 'N'), coalesce(quote(b), 'N'));
SELECT count(*) FROM t;
PRAGMA integrity_check;
```

| | doltlite | stock |
|---|---|---|
[...]

PR #2105, opened 2026-08-11
Sweep 10,000 seeds per pull request and 100,000 nightly

timsehn:
The sweep was using a small fraction of the time it had.

| | seeds | measured | budget |
|---|---|---|---|
| nightly, before | 4,000 | **2m46s** on the runner | 180 min |
| per-PR gate, before | 400 | ~20s | 12 min |
| nightly, after | 100,000 | ~75–90 min | 180 min |
| per-PR gate, after | 10,000 | ~7 min | 25 min |

41ms per seed on the runner, most of it spawning two engines and the generator rather than executing SQL. Both numbers were chosen when the sweep was new and nobody knew what a seed cost.

Seed count is the whole of this harness's reach. Every bug it has found was a specific seed, and the one that found the savepoint/index rollback corruption in #2103 was **510519** — reached by sweeping 12,000 seeds by hand, well past where any configured job would have looked. At 4,000 a night that seed was 128 nights away.

## Window arithmetic

The nightly window now starts at 10,000,001, above the gate's fixed `1..10000`, so the nightly never spends its budget re-checking seeds every pull request already cleared. The modulus gives 10,000 distinct windows before one repeats; at 4,000 per night against a 1e6 modulus it came back around every 250 runs — under a year.

## Budgets

The gate's step budget is 25 minutes against a measured ~7, deliberately more than double. This sweep's cost scales with process startup, so it stretches under runner load rather than with anything about the change under test, and a gate that blocks every pull request should not be the thing that flakes on a slow runner.

## Testing

- 10,000 seeds on the gate's exact configuration (default axes, current master) run locally to confirm both the wall time and that the window is clean — a gate that fails on master would block everything.
- The nightly's per-seed cost with every group enabled measured at 55ms vs 50ms for the default axes on the same binary, ~10%, so all-groups does not change the sizing.
- `FIRST`/`LAST` arithmetic checked for the first several run numbers and at the modulus wrap.

Overlaps #2102 in `nightly-fuzz.yml` — same step, adjacent lines. Whichever lands second needs a trivial conflict resolution; I'll take care of it.

### DIFF
--- a/.github/workflows/nightly-fuzz.yml
+++ b/.github/workflows/nightly-fuzz.yml
@@ -183,10 +183,10 @@ jobs:
   # The per-PR gate lives in test.yml and sweeps a fixed seed window; this job
   # sweeps a moving one for breadth, so the two do not re-check the same seeds.
   #
-  # Each night sweeps a different seed window derived from the run number, so
-  # coverage compounds instead of re-checking seeds earlier runs already
-  # cleared. The window is printed and every failing seed is reproducible from
-  # the generator alone.
+  # Each night samples a fresh window at random rather than stepping through
+  # the space in order. The window is printed, and a failing seed is its own
+  # reproducer, so nothing is lost by not knowing in advance which seeds a
+  # given night will draw.
   differential:
     runs-on: ubuntu-latest
     timeout-minutes: 180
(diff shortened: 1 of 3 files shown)


timsehn:
Switched the nightly to **drawing its window at random** rather than stepping through the space by run number. A seed is a PRNG seed — seed N and seed N+1 are unrelated streams — so marching in order samples the space no better than rolling for it, and it cost run-number arithmetic plus a modulus that wrapped every 250 runs. The window is printed and a failing seed is its own reproducer, so nothing is lost by not knowing the draw in advance.

**The gate keeps its fixed window on purpose.** A gate exists to say whether the change under test broke something. Rolling dice there would let an unrelated pre-existing bug redden an arbitrary pull request and then fail to reproduce on a re-run — the worst kind of gate.

Drawing high seeds turned up a real bug in the runner, which is worth stating because it would have been invisible in CI:

```
$ seq 4823991107 4823991109        # BSD/macOS
4.82399e+09
4.82399e+09
4.82399e+09
```

The generator gets `4.82399e+09` and rejects it, so the runner reported *every* seed as a failure — with an empty failing-seed list, because a generator failure counts as a failure without being recorded as one. GNU `seq` on the runner prints these exactly, so a window the nightly drew would have failed on every seed on a Mac and passed in CI. Now `seq -f %.0f`.

Verified after the fix: a random 300-seed window at 4.8e9 with **every group** enabled is 300/300 clean.

PR #2106, opened 2026-08-11
perf: combine external file state probes

timsehn:
## Summary

- combine Unix moved-file detection and file-size lookup into one `stat()` result
- remove one metadata syscall from each file-backed transaction refresh
- retain the existing two-probe fallback for custom or unsupported VFS implementations
- keep the SQLite-core addition behind `DOLTLITE_PROLLY`

## Performance

Paired before/after measurements used 10,000 rows and alternating invocation order. Lower multipliers are faster.

| Sysbench suite | File reads | Autocommit reads | In-memory control |
|---|---:|---:|---:|
| Integer PK, repeat run | 0.94x | 0.94x | 1.00x |
| TEXT PK | 0.95x | 0.95x | 1.01x |
| BLOB PK | 0.96x | 0.95x | 0.99x |
| Composite PK | 0.95x | 0.96x | 0.99x |

The first integer-PK run independently measured 0.93x for both file-read sections. Point, range, index, covering-index, join, and mixed read-only workloads improved; full scans and write sections remained effectively flat.

## Validation

- `make doltlite doltlite-lib`
- clean `DOLTLITE_PROLLY=0` SQLite build
- `make lint`
- GC rewrite-failure regression: 15/15
- DoltLite GC suite: 58/58
- specialized concurrency and fault-injection C suite: 9/9

### DIFF
--- a/src/chunk_store_lock.c
+++ b/src/chunk_store_lock.c
@@ -338,16 +338,13 @@ static int csDetectExternalChanges(ChunkStore *cs, int *pChanged){
 
   fileState.iFileSize = -1;
   fileState.bMoved = 0;
-  rc = sqlite3OsFileControl(cs->file.pFile,
-                            SQLITE_FCNTL_DOLTLITE_FILE_STATE, &fileState);
+  rc = sqlite3OsDoltliteFileState(cs->file.pFile, &fileState);
   if( rc==SQLITE_OK ){
     bMoved = fileState.bMoved;
     haveFileState = fileState.iFileSize>=0;
-  }else if( rc==SQLITE_NOTFOUND ){
+  }else{
     rc = sqlite3OsFileControl(cs->file.pFile, SQLITE_FCNTL_HAS_MOVED, &bMoved);
     if( rc!=SQLITE_OK ) return rc;
-  }else{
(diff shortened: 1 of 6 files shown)


timsehn:
CI follow-up pushed in `a406cdb437`.

Root cause: wrapped VFS implementations forwarded the private combined probe to the Unix file, bypassing wrapper `xFileSize` behavior used for readonly, I/O-fault, and OOM testing.

The combined probe is now an internal native-Unix helper that first verifies the handle uses Unix I/O methods. Wrapped/custom VFSes, `SQLITE_TEST`, Windows, KV, and other VFSes return `SQLITE_NOTFOUND` and retain the original `HAS_MOVED` plus `xFileSize` path.

Validation:
- exact `refresh_error_propagation` failure: 11/11
- GC rewrite failure: 15/15
- `misc7`, `ioerr3`, `statfault`, and `vacuum3`: 1,757 assertions with the existing divergence set intact
- all CI-failing testfixture suites rerun; the only local mismatches reproduce identically on untouched `origin/master`
- final paired sysbench: file reads 0.94x, autocommit reads 0.94x
- build and lint pass

PR #2124, opened 2026-08-11
Reset and stage a table as one object with its indexes and rename mate

timsehn:
Fixes the reset arm of the staging object-identity cluster (deep-review action item 1). dolt_reset of a path treated the named table entry as the whole object; its unnamed index entries and its rename's other half stayed staged. Confirmed consequences, each now an oracle case:

- Reset of a staged new table with an index left a dangling index entry; the next commit succeeded and checkout of it failed with `malformed database schema`, bricking the branch.
- Reset of the new name of a staged rename left a half-state the next commit turned into a bare DROP (table gone from history). Dolt treats this reset as a no-op.
- Reset of the old name left the opposite half-state, committing the table twice. Dolt unstages the whole rename atomically.
- Reset of a staged index-only change was a silent no-op; the "reset" index still committed.

**Mechanics**

- `resetStageNamedPaths` resolves the argument through the same rename pairing `dolt_status` renders (`isRenamePair`, exported as `doltliteCatalogRenameMate`), and every path carries the table's unnamed index entries with it, resolved through the staged catalog's own schema rows (the previous restore path resolved them through HEAD's rows — the wrong numbering domain).
- `dolt_add` of a renamed table retires the old name's index entries too; before, they survived as duplicates keyed to a dead name and produced phantom staged rows.
- The catalog serializer closes the remaining cross-domain hole for constructed arrays: index rows whose parent table is absent from both the row set and the entry set purge before the supplementation passes, and live supplementation refuses to adopt an index row whose parent is not part of the catalog (a live number coinciding with a HEAD-restored entry used to retarget the index at a table the catalog doesn't contain).
- `statusCompareIndexSchemaObjects` compares a renamed table's index set across the rename pair by index name instead of exact SQL, so a pure rename no longer reports phantom `modified` rows for both names.

**Validation**

- 8 new `vc_oracle_reset_test.sh` cases: all fail before, pass after, against Dolt 2.2.2; full reset suite 51/51.
- add / status / status_schema_objects / commit / checkout / merge_status oracle suites green; add / status / checkout / commit / reset also green under a `DOLTLITE_PROLLY_CHECK=1` build (the new purge does not trip the write-time catalog invariants).
- Full shell battery: 101/101 suites.

Known limitation (pre-existing, consistent with status): the rename pairing keeps the same-table-number gate, so an order-shifting rename (rename that changes canonical sort position among multiple tables) is still not paired — that is the add/commit-side numbering-domain cluster, next PR.

### DIFF
--- a/src/doltlite_add.c
+++ b/src/doltlite_add.c
@@ -737,6 +737,7 @@ int doltliteStageNamedTables(
       if( aWorking[j].iTable==iTable ){
         int k;
         int updated = 0;
+        int iRetired = -1;
         updateMaster = 1;
         for(k=0; k<nStaged; k++){
           int nameMatch = aStaged[k].zName && aWorking[j].zName
@@ -776,6 +777,7 @@ int doltliteStageNamedTables(
             ** where they no longer exist. */
             if( nameChanged && aStaged[k].zName ){
               ADDNAMED_TOUCH(aStaged[k].zName);
+              iRetired = nTouched-1;
             }
             zDup = aWorking[j].zName
                            ? sqlite3_mprintf("%s", aWorking[j].zName) : 0;
@@ -799,9 +801,16 @@ int doltliteStageNamedTables(
           }
         }
         /* An index travels with its table: replace whatever index entries
-        ** the staged catalog carried for this table with the working ones. */
+        ** the staged catalog carried for this table with the working ones.
+        ** A staged rename's index entries resolve through the OLD name in
+        ** the staged schema rows, so the retired name's entries go too. */
         addRemoveIndexEntriesOfTable(aStaged, &nStaged,
                                      aStagedSchema, nStagedSchema, zTable);
+        if( iRetired>=0 ){
+          addRemoveIndexEntriesOfTable(aStaged, &nStaged,
+                                       aStagedSchema, nStagedSchema,
+                                       azTouched[iRetired]);
+        }
         rc = addAppendIndexEntriesOfTable(context, &aStaged, &nStaged,
                                           aWorking, nWorking,
                                           aWorkSchema, nWorkSchema, zTable);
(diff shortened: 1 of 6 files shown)


timsehn:
Could not reproduce the RESET-3 finding against this PR's binary; the recorded evidence doesn't match what the fix produces. Exact steps (indexed `t` committed, `ALTER TABLE t RENAME TO t2`, `dolt_add('-A')`, `dolt_reset('t')`, plain `dolt_commit`):

```
--status-after-add--    t -> t2|1|renamed
--status-after-reset--  t -> t2|0|renamed
--commit--              Error: nothing to commit, working tree clean (use dolt_add to stage changes)
--reopen sqlite_master-- t2, tvi        (working catalog; log has only c1)
```

Three points on the expected behavior:

1. **`t -> t2|0|renamed` after reset IS correct.** `dolt_reset` unstages; it does not revert the working tree — that's checkout. Dolt 2.2.2 prints exactly this row for the same script (the `reset_path_staged_rename_old_name_unstages_whole` oracle case in this PR compares the two engines verbatim). "Clean working tree" is the wrong expectation here.

2. **The plain follow-up commit is rejected** ("nothing to commit"), also pinned by the `reset_path_staged_rename_old_name_commit_errors` oracle case, which requires both engines to error.

3. **`t` absent / `t2`+`tvi` present after reopen is the working tree**, not the commit — the rename still exists in the working set after an unstage, same as Dolt; the log still has only `c1`.

The one way to make the commit "succeed" is `dolt_commit('-am', ...)`: `-a` re-stages the working-tree rename and then hits the pre-existing commit-a bug where an unstaged rename commits as a DROP (`doltliteCommitStageModifiedOnly` pairs tables by name only). That bug predates this PR, is out of its scope, and is the next PR in this cluster — it needs the `doltliteCatalogRenameMate` helper this PR introduces.

If the recorded run shows a plain (no `-a`) commit succeeding, the binary under test likely predates e457652ed8 — happy to dig into the run artifacts if it reproduces on this head.

PR #2130, opened 2026-08-11
Pair staged renames by content and keep staging in one numbering domain

timsehn:
Deep-review action item 1, staging cluster — the data-loss core. Staging paired a renamed table with its mate by **table number**, but numbers are canonical-by-sorted-name: a drop hands its number to the next table, and an order-shifting rename changes its number. Both directions were broken, confirmed against Dolt 2.2.2:

- **Silent history loss:** with an unstaged `DROP TABLE m`, `dolt_add('a')` for a new table that reused m's number staged the drop as a rename half — the next commit erased `m` from history and `dolt_status` reported clean afterwards.
- **Order-shifting renames shredded:** after `ALTER TABLE t RENAME TO z` shifts canonical order, `dolt_add('u')` (a bystander) left phantom staged rows and the commit reverted the rename in history; `dolt_add('z')` committed both `t` and `z`.

**Mechanics**

- Rename identity is now content, everywhere staging pairs tables: `isRenamePair` keeps the number gate as a fast path but delegates to a content core (row-root equality, or the rename-tolerant schema comparison; two *empty* roots are never identity on their own, so an empty drop+create no longer reads as a rename — that arm previously false-paired even in status).
- `doltliteCatalogRenameMate` falls back to a content scan when the number candidate fails, accepting only an unambiguous mate; `dolt_add` uses it in place of its number-coincidence heuristic, and `dolt_status` runs a content second pass so order-shifting renames render as `t -> z renamed` (the index-object comparison inherits this through the shared helper).
- `dolt_add` output stays in one numbering domain: after alignment, entries whose tables left the working tree move to fresh numbers above both domains, and `doltliteBuildNamedStageMasterRoot` stamps old-sourced table rows from the **final** entry list, so every row pairs with its entry by number (the write-time PROLLY_CHECK invariants watch exactly this).

**Validation**

- 7 new `vc_oracle_add_test.sh` cases vs Dolt 2.2.2: unstaged-drop sweep (status + post-commit history), both order-shifting rename shapes (status + post-commit contents), empty drop+create non-rename. All fail before, pass after.
- add 41/41, status 40/40, status_schema_objects 18/18, reset 54/54, checkout 67/67, commit 39/39, merge_status 11/11 — also green under a `DOLTLITE_PROLLY_CHECK=1` build.
- Full shell battery 101/101. Zero build warnings; strict-C90 clean.

Remaining in the cluster after this: `dolt_commit('-a')` committing an unstaged rename as a DROP (`doltliteCommitStageModifiedOnly` pairs by name only) — next PR, on top of this pairing helper.

### DIFF
--- a/src/doltlite_add.c
+++ b/src/doltlite_add.c
@@ -158,6 +158,30 @@ static void addAlignStagedEntriesToWorking(
   addNameIndexFree(&workingIdx);
 }
 
+/* Move staged entries whose tables no longer exist in working onto fresh
+** numbers above both domains. Alignment cannot renumber them (nothing to
+** align to), and the number they hold may have been reused by the working
+** domain for a different table. Rows follow through the master composer,
+** which stamps old-sourced table rows from the final entry numbering. */
+static void addRenumberStaleStagedEntries(
+  struct TableEntry *aStaged, int nStaged,
+  struct TableEntry *aWorking, int nWorking
+){
+  Pgno iNext = 2;
+  int i;
+  for(i=0; i<nWorking; i++){
+    if( aWorking[i].iTable >= iNext ) iNext = aWorking[i].iTable + 1;
+  }
+  for(i=0; i<nStaged; i++){
+    if( aStaged[i].iTable >= iNext ) iNext = aStaged[i].iTable + 1;
+  }
+  for(i=0; i<nStaged; i++){
+    if( aStaged[i].iTable<=1 || !aStaged[i].zName ) continue;
+    if( addFindEntryByName(aWorking, nWorking, aStaged[i].zName) ) continue;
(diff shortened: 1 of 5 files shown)


timsehn:
RENAME-3 confirmed by repro and fixed in follow-up #2131 (merged separately since this PR was already merging): a rename mate is now rejected when another to-side entry carries the same non-empty row content, so the ambiguous add stages a plain new table and status shows deleted plus two new tables, matching Dolt. The symmetric version of the rule (from-side contest too) regresses the existing multi_table_rename_drop_create_mix oracle case — Dolt pairs a real rename even when a dropped identical sibling exists — so only to-side candidates contest; rationale and oracle coverage in #2131.

PR #2145, opened 2026-08-11
Merge over an empty ancestor when the key shape changed on both branches

timsehn:
Deep-review item 5, PR 0 — the investigation the plan called for confirmed a real **merge-correctness** bug, so it ships first and alone. When both branches drop and recreate a table with a different key shape (text PK → INTEGER PRIMARY KEY), the row merge walked the OLD ancestor root with the NEW shape's flags. Raw text keys alias intkeys (every 5-char text key is some 8-byte intkey — `'AAAAA'` is `-5385951930834944000`), so an old row read as a phantom row at a real key: one side's insert became a "modify", the other side's absence a "delete", and the merge raised a **conflict neither side created**, rolling back a perfectly clean merge.

**Fix:** in `mergePass1MergeTableData`, when `(anc->flags ^ ours->flags) & INTKEY`, the row merge runs over an empty ancestor — a key-shape change is a recreate, the recreated table is a different object, and both sides' rows are additions. Genuine add/add collisions at one key still conflict (verified); identical recreates with disjoint rows merge clean.

**Validation:** two new `doltlite_merge.sh` battery cases with the aliasing key spelled out (clean-merge shape and true-collision shape) — fail before, pass after. Merge/merge_status/CV/revert_cherrypick/replay/rebase oracle suites green; full c-regression 310,258/310,258; amalgamation compiled (the new standing gate); strict-C90 clean.

One test-harness discovery worth noting: doltlite sessions open on the **default branch** — `dolt_checkout` does not persist across processes (matching `dolt sql` semantics). The battery cases merge into main accordingly.

Remaining item-5 PRs (rendering, no shared files with this): per-side key shapes through the diff iterator, name-based column mapping, `to_commit=` ancestry gate, diff_stat cell maps.

### DIFF
--- a/src/doltlite_merge_pass1.c
+++ b/src/doltlite_merge_pass1.c
@@ -180,6 +180,8 @@ static int mergePass1MergeTableData(
   struct TableEntry *pTheirsEntry
 ){
   ProllyHash mergedTableRoot;
+  ProllyHash emptyAnc;
+  const ProllyHash *pAncRoot = &pAnc->root;
   int nConflicts = 0;
   DoltliteConflictRow *aConflictRows = 0;
   MergeIndexInfo *aIdxInfo = 0;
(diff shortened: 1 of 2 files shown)


timsehn:
Closing — decree from review: doltlite should refuse to merge tables whose primary keys differ, matching Dolt (`cannot merge because table t has different primary keys`). The empty-ancestor clean merge here goes the opposite direction (and ito correctly found it left indexed recreates broken). Replacement PR reworks this as a refusal.

PR #2174, opened 2026-08-12
Fix amalgamation build with custom WAL sync

timsehn:
## Summary

- resolve the stock btree delegate before reading its shared state in DoltLite builds
- keep stock SQLite behavior unchanged behind `DOLTLITE_PROLLY`
- compile the amalgamation in CI with the reported Emscripten feature and omit flags

## Testing

- fail-before: reported flags fail at `pDb->pBt->pBt` with incomplete `Btree`
- `test/amalgamation_wasm_compile_test.sh build/sqlite3.c`: PASS for the existing wa-sqlite profile and reported user profile
- native `cc -fsyntax-only` with the reported defines: PASS
- `make -C build doltlite doltlite-lib`: PASS
- `make -C build lint`: 14 passed, 0 failed

### DIFF
--- a/src/btree.c
+++ b/src/btree.c
@@ -3286,7 +3286,17 @@ static void setDefaultSyncFlag(BtShared *pBt, u8 safety_level){
   sqlite3 *db;
   Db *pDb;
   if( (db=pBt->db)!=0 && (pDb=db->aDb)!=0 ){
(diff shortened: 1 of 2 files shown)


timsehn:
Combined this fix into #2171 so the related amalgamation build changes land and run CI together.

PR #2199, opened 2026-08-13
Seed the session's head commit as a GC root

### LINKED ISSUE #2176: Session HEAD commit is not a GC root: a detached session's own dolt_gc collects its live HEAD
Found during the Aug 12 review; verified.

`doltliteSeedSessionHashes` (src/prolly_btree_state.c, ~:921-948) seeds every live session hash as a gc root — `stagedCatalog`, `mergeCommitHash`, `conflictsCatalogHash`, `preRebaseWorkingCat`, `rebaseOntoCommit`, `constraintViolationsHash`, and the live catalog roots — but never `pBt->headCommit`.

Repro (verified):
```
commit on branch b2, note its tip H
dolt_branch('-D','b2')
open db/<H>            -- detached, dolt_log shows 3 rows, reads work
  SELECT dolt_gc();    -- "15 chunks removed"
open db/<H> again      -- fails: unable to open database (SQL logic error)
```
Plain `VACUUM` reproduces identically (VACUUM is gc in this engine). Data reads survive because the catalog roots and the revision catalog are seeded — only the session's live HEAD commit object is collected, permanently.

Bounded by needing a detached session at a commit no ref reaches (tag/branch/`~`-suffix revisions stay reachable), and detached sessions are read-only so no dangling ref results — this is loss of a commit object under a live session, not store corruption.

Note `gcVerifySessionResolvable` shares the same seed list, so the `DOLTLITE_PROLLY_CHECK` invariant structurally cannot catch this omission — the fix should seed `headCommit` and consider making the verifier derive its list independently.

timsehn:
Closes #2176.

## Problem

`doltliteSeedSessionHashes` seeds every live session hash as a GC root — `stagedCatalog`, `mergeCommitHash`, `conflictsCatalogHash`, `preRebaseWorkingCat`, `rebaseOntoCommit`, `constraintViolationsHash`, and every table root in the live catalog — but never `pBt->headCommit`.

A branch reaches its own tip, so nothing goes wrong until a session is **detached** at a commit no ref reaches. Then its own `dolt_gc()` collects the commit it is reading through:

```sql
-- commit on branch feat, note its tip H, then:
SELECT dolt_checkout('main');
SELECT dolt_branch('-D','feat');

-- open db/<H>: works, dolt_log shows 3 rows
SELECT dolt_gc();          -- "15 chunks removed"

-- open db/<H> again:
-- Error: unable to open database ".../<H>": SQL logic error
```

The commit object is gone permanently. Plain `VACUUM` does the same, being the same sweep. Reads keep working *in the session that ran the sweep* because the catalog roots it holds are seeded — that is what keeps the loss quiet until something reopens.

Bounded by needing a detached session at an unreachable commit, and detached sessions are read-only, so no dangling ref results: this is loss of a commit object under a live session, not store corruption.

## Fix

Seed `pBt->headCommit` alongside the other session hashes. The sweep walks a commit's children, so this covers the detached session's whole view — its catalog and table data — transitively.

## Testing

Four cases in `doltlite_gc_session_state.sh`: the revision reads before the sweep, survives its own sweep, keeps its log, and leaves the branch it forked from intact. The two post-sweep cases fail on master; the pre-sweep read passes either way, which separates the collection from an unrelated open failure.

Amalgamation and lint clean. Full battery run twice under identical conditions, once on this branch and once on a clean master worktree at the same base — identical results, assertion for assertion.

## Note on the invariant check

`gcVerifySessionResolvable` (the `DOLTLITE_PROLLY_CHECK` gate) calls `doltliteSeedSessionHashes` for its own list, so it structurally cannot catch an omission from that list — it verifies exactly what the sweep already seeded. Making the verifier derive its roots independently is a real improvement but a separate change from this fix; #2176 notes it and I have left it there.

### DIFF
--- a/src/prolly_btree_state.c
+++ b/src/prolly_btree_state.c
@@ -940,7 +940,10 @@ int doltliteSeedSessionHashes(
     Btree *pBt = db->aDb[i].pBt;
     if( !pBt || pBt->pOps!=&prollyBtreeOps || !pBt->pBt ) continue;
     if( &pBt->pBt->store!=cs ) continue;
-    rc = xPush(pCtx, &pBt->vc.stagedCatalog);
+    /* A detached session's head is reachable from no ref, so nothing else
+    ** keeps the commit it reads through alive. */
+    rc = xPush(pCtx, &pBt->headCommit);
+    if( rc==SQLITE_OK ) rc = xPush(pCtx, &pBt->vc.stagedCatalog);
     if( rc==SQLITE_OK ) rc = xPush(pCtx, &pBt->vc.mergeCommitHash);
     if( rc==SQLITE_OK ) rc = xPush(pCtx, &pBt->vc.conflictsCatalogHash);
     if( rc==SQLITE_OK ) rc = xPush(pCtx, &pBt->preRebaseWorkingCat);
--- a/test/doltlite_gc_session_state.sh
+++ b/test/doltlite_gc_session_state.sh
@@ -121,13 +121,13 @@ run_rev_test() {
 # Pairs with the case below: same database, same open form, but a commit a
 # branch still reaches. If both fail, the revision open is at fault; if only
 # the unreachable one does, the sweep is.
-run_rev_test "detached_reachable_head_before_gc" "SELECT count(*) FROM t;" "1" "$DB/$MAIN_TIP"
-run_rev_test "detached_head_before_gc" "SELECT count(*) FROM t;" "2" "$DB/$FEAT_TIP"
+run_rev_test "detached_reachable_head_before_gc" "SELECT count(*) FROM t;" "1" "$DB@$MAIN_TIP"
+run_rev_test "detached_head_before_gc" "SELECT count(*) FROM t;" "2" "$DB@$FEAT_TIP"
 
-"$DOLTLITE" "$DB/$FEAT_TIP" "SELECT dolt_gc();" > /dev/null 2>&1
+"$DOLTLITE" "$DB@$FEAT_TIP" "SELECT dolt_gc();" > /dev/null 2>&1
 
-run_rev_test "detached_head_survives_own_gc" "SELECT count(*) FROM t;" "2" "$DB/$FEAT_TIP"
-run_rev_test "detached_head_log_survives" "SELECT count(*) FROM dolt_log;" "3" "$DB/$FEAT_TIP"
+run_rev_test "detached_head_survives_own_gc" "SELECT count(*) FROM t;" "2" "$DB@$FEAT_TIP"
+run_rev_test "detached_head_log_survives" "SELECT count(*) FROM dolt_log;" "3" "$DB@$FEAT_TIP"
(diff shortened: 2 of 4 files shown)


timsehn:
Two CI reds, both traced rather than assumed, plus a deliberate baseline raise that needs a reviewer's eye.

**The fault-injection reds are caused by this fix, and the divergence file predicted them.** Its own comment on those entries says gc runs a post-sweep session-resolvability check that re-reads chunks and treats injected faults as inconclusive, and that *"the exact indices move when commit-path allocation counts change."* That check is `gcVerifySessionResolvable`, which calls `doltliteSeedSessionHashes` — the function this PR changes. One more seeded root means one more re-read, so the fault window extends by exactly one ordinal. That is what showed up: `.32.3` in all four `vacuum3-ioerr` loops and in `ioerr-2`, immediately after the known-diverging `.29/.30/.31`, uniformly across five independent loops.

I added those five ordinals in the existing style and extended both comment blocks to say the range tracks the session root set, so the next person does not re-derive it. **That grew the exception ratchet by 5, which I raised deliberately (`gates` 4846→4851, `intentional` 3547→3552) — the file asks for that to be called out in the PR description, so: called out.** This is the one edit here that could hide a real bug, and my grounds for it are the documented mechanism, the exact +1, and the uniformity across loops — not that the red was inconvenient. Worth a look.

**Windows was not this fix at all.** The reachable-commit control I added answered it: that case failed too, and no sweep can explain a commit a branch still reaches. A revision opens read-only, and on Windows one landing straight after a write session cannot replay what that session left; `doltlite_open_branch.sh` gets away with the same open only because an ordinary session happens to run first. I reordered an ordinary open ahead of the detached cases and kept the reachable/unreachable pair permanently, so a future failure there says which of the two is at fault instead of leaving it ambiguous.

Rebased onto master to pick up the current ratchet baseline.

timsehn:
`Build-Test / ubuntu` here is not this PR. `concurrent_continue_adoption_exactly_one_winner` failed with parent=`rebase failed — branch restored to pre-rebase state`, child=`no rebase in progress` — neither side won.

I built `multi_process_merge_rebase_test` from a clean master worktree and looped it: **4 failures in 140 runs (~3%)**, with the same symmetric shape and a worse local variant (`rebase recovery failed — pre-rebase state may not have been fully restored`). Filed as #2206 with the rate and both messages.

Nothing in this PR reaches `dolt_rebase('--continue')` — it adds one hash to the gc root seed list. Rerunning the job.

timsehn:
Thanks for the `@hash` fix — that is the actual cause, and my explanation was wrong.

I had it as the shared helper piping SQL in on stdin, since `doltlite_open_branch.sh` opens hash revisions by argument and passes on Windows while the helper form failed. That fit the observations but was the wrong mechanism: MSYS resolves the slash-qualified hash as a filesystem path before doltlite sees a revision at all. I have corrected the comment I left in the suite, which credited stdin.

Two earlier guesses on this same failure that also did not hold, in case it saves anyone repeating them: it is not the sweep (the reachable-commit control fails identically), and it is not WAL recovery on a read-only open after a write session (putting an ordinary open first changed nothing).

Issue #2204, opened 2026-08-13
The shell battery is never run against an assert-enabled build; 6 suites fail on master when it is


timsehn:
`ci-build.yml`'s `assert-enabled` job builds only two targets — `catalog_serialize_determinism_test` and `doltlite_regression_test_c` — and `build-test.yml` says so outright:

> The regression suite is the only place most of these assertions are ever evaluated -- every other job builds with NDEBUG.

`test/run_doltlite_tests.sh` is never among them. Running it against a `--with-debug` build of master (`f00884405a`) on macOS:

| build | result |
|---|---|
| NDEBUG (`configure`) | 101 passed, 2 failed — both rig-only (`./sqlite3` absent, `libdoltlite.a` not built) |
| assert (`configure --with-debug`) | 95 passed, 8 failed |

The six extra failures are assertion aborts, not wrong answers, in five distinct assertions across four subsystems. Each now has a one-line repro:

- #2200 — `prollyNodeIntKey` INTKEY on a blobkey node, 16 cases (`dolt_history`/`dolt_at`/pushdown over WITHOUT ROWID and text keys)
- #2201 — `cs->lockDepth>0` in `doltliteCompareAndAdvanceBranch` on `:memory:` commit
- #2202 — `pushSavepoint` `inTrans==TRANS_WRITE` on ROLLBACK TO after a spilling insert
- #2203 — stock `seekAndRead` and VDBE `szRow` bounds on 1MB blob reads and 100KB rows

None of them changes an answer under NDEBUG, which is exactly why they have survived: the release build reads a mistyped node, advances a ref unlocked, or issues an oversized read, and returns the right result anyway.

Proposal: add an assert-enabled battery bucket to CI. It cannot go green today, so the sequencing is to fix or explicitly exempt the five assertions first, then add the bucket to keep them fixed. Filing this as the umbrella so the four bugs above have somewhere to point.

Found while running the battery under `--with-debug` as a control for an unrelated fix. #2198 is the same class of gap from the other direction: an assert-only abort a suite was hiding by discarding stderr.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

timsehn:
All four clusters are fixed and merged: #2200 (#2207), #2201 (#2208), #2202 (#2209), #2203 (#2210). #2198 — the FK-violation merge abort — is still open and is the fifth.

**The premise here is now verified rather than inferred.** On master with all four in, an assert-enabled build (`configure --with-debug`) passes:

[...]

Issue #2201, opened 2026-08-13
dolt_commit on an in-memory database asserts cs->lockDepth>0


timsehn:
Pre-existing on master (`f00884405a`). Assert-enabled build only.

```
doltlite ":memory:" "CREATE TABLE t(a INTEGER PRIMARY KEY); INSERT INTO t VALUES(1); SELECT dolt_commit('-Am','x');"
```

```
Assertion failed: (cs->lockDepth>0),
function doltliteCompareAndAdvanceBranch, file doltlite_core.c, line 740.
```

The branch advance runs without the graph lock held. For an in-memory store there is no peer to race, so the missing lock costs nothing today — but `doltliteCompareAndAdvanceBranch` is the ref-advance primitive whose whole contract is that the caller holds the lock across read-confirm-advance (#1098), and `PROLLY_ASSERT_GRAPH_LOCKED` exists to enforce exactly that. An in-memory path that reaches it unlocked either needs the lock taken or needs an explicit, documented exemption; silently asserting is neither.

Hit by `doltlite_memory_db.sh` cases `memory_supports_dolt_commit` and `memory_supports_dolt_checkout_isolates_branches`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #2200, opened 2026-08-13
dolt_history/dolt_at over a WITHOUT ROWID table asserts INTKEY on a blobkey node


timsehn:
Pre-existing on master (`f00884405a`). Only visible in an assert-enabled build (`configure --with-debug`); the NDEBUG build returns correct results, so this has never shown up in CI.

```sql
CREATE TABLE t(k TEXT PRIMARY KEY, v TEXT) WITHOUT ROWID;
INSERT INTO t VALUES('x','one');
SELECT dolt_commit('-Am','c1');
SELECT * FROM dolt_history_t;
```

```
Assertion failed: ((pNode->flags & PROLLY_NODE_INTKEY)!=0),
function prollyNodeIntKey, file prolly_node.c, line 196.
```

The historical vtabs read a blobkey node through the intkey accessor. The non-assert build then reads an integer out of a node that never stored one — the rendered rows happen to come out right in these cases, which is what has kept it quiet, but the read is not type-correct and there is no reason to believe the coincidence holds for every key shape.

16 cases across three suites hit it, all on non-integer or WITHOUT ROWID keys:

- `doltlite_history.sh`: `keyshape_history_eq_filters`, `keyshape_history_upper_bound`, `keyshape_history_values_render`
- `doltlite_at.sh`: `keyshape_eq_filters`, `keyshape_eq_matches_text`, `keyshape_upper_bound_scans_all`, `keyshape_values_render`
- `doltlite_dolt_table_pushdown.sh`: `nonint_history_filter_correct`, `nonint_history_total`, `worid_at_pk_eq_absent_is_empty`, `worid_at_pk_eq_selects_that_row`, `worid_at_renders_pk`, `worid_composite_at_renders_key`, `worid_history_pk_eq`, `worid_history_pk_range`, `worid_history_renders_pk`

Reproduce by running any of those suites against a `--with-debug` build. See the umbrella issue on assert coverage for why CI misses all of these.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #2206, opened 2026-08-13
Concurrent dolt_rebase('--continue') can leave no winner, and the loser's recovery reports it may not have restored


timsehn:
Pre-existing on master (`7760299976`), reproduces on both macOS and Ubuntu. Found when `multi_process_merge_rebase_test` Test 6 went red on an unrelated PR's CI; I then reproduced it on a clean master build.

`concurrent_continue_adoption_exactly_one_winner` requires that when two processes race `dolt_rebase('--continue')`, exactly one reports `Successfully` and the other reports `no rebase in progress`. Sometimes neither wins.

## Rate

Clean master build, macOS, `./multi_process_merge_rebase_test` in a loop: **4 failures in 140 runs (~3%)**. Also seen once on an Ubuntu CI runner.

## What the two sides report

The loser is correct — it sees no rebase in progress. The winner fails instead of succeeding, in two variants:

```
FAIL trial 2 parent=[STEP_ERR: rebase recovery failed — pre-rebase state may not have been fully restored] child=[STEP_ERR: no rebase in progress]
FAIL trial 4 parent=[STEP_ERR: no rebase in progress] child=[STEP_ERR: rebase recovery failed — pre-rebase state may not have been fully restored]
```

and on the Ubuntu runner the milder form:

```
FAIL trial 3 parent=[STEP_ERR: rebase failed — branch restored to pre-rebase state] child=[STEP_ERR: no rebase in progress]
```

Note which side says what: it is symmetric, so this is not parent-vs-child ordering. The **"recovery failed — pre-rebase state may not have been fully restored"** variant is the one worth attention: that is not a clean loss, it is a failed rollback of the rebase it was in the middle of finalizing.

## Reproduce

```bash
cd build
make multi_process_merge_rebase_test
for i in $(seq 1 100); do ./multi_process_merge_rebase_test 2>&1 | grep "FAIL trial"; done
```

## Notes

The test's own post-conditions (branch back on `main`, rows present, no leftover `dolt_rebase_feat` branch) passed in the runs I saw, so the observable damage in these trials was confined to the error paths. That does not settle whether the failed recovery can leave the temporary branch or a partial working set behind under other timings — the message says it may.

[...]

PR #2210, opened 2026-08-13
Slice oversized VFS reads, and stop asserting a page bound on a record

### LINKED ISSUE #2203: Reading a 1MB blob back after reopen asserts in stock seekAndRead; a 100KB row asserts szRow<=65536
Pre-existing on master (`f00884405a`). Assert-enabled build only. Two stock-SQLite invariants, both reached through doltlite's chunk reads.

```sql
-- session 1
CREATE TABLE t(id INTEGER PRIMARY KEY, data BLOB);
INSERT INTO t VALUES(1, randomblob(1048576));
SELECT dolt_commit('-A','-m','1MB');
-- session 2 (reopen)
SELECT length(data) FROM t WHERE id=1;
```

```
Assertion failed: (cnt==(cnt&0x1ffff)),
function seekAndRead, file os_unix.c, line 3470.
```

The read only asserts after a reopen — in the writing session it is served from cache — so this is the on-disk chunk being pulled in a single `read()` larger than the 128KiB that stock's VFS layer asserts as a page-sized bound. Stock never issues a read that big because it reads a page at a time; doltlite reads whole chunks, so the assertion no longer describes the caller. Either the read needs chunking or the assertion is stock's and does not apply to this path — but it should be decided, not left firing.

Separately, a ~100KB row trips the VDBE's own row-size bound:

```
Assertion failed: (pC->szRow<=65536), function sqlite3VdbeExec, file vdbe.c, line 3144.
```

`doltlite_unicode_blob.sh` cases: `1mb_blob_size`, `1mb_blob_on_branch`, `1mb_blob_after_merge`, `blob_count_after_merge` (seekAndRead) and `100kb_length` (szRow).

These are the two places where doltlite's "a row is a record in a chunk" model runs past bounds stock derives from its page format, so they are worth answering together.

timsehn:
Closes #2203.

Two bounds, and they turned out to be different kinds of problem.

## 1. `seekAndRead` — doltlite's own convention, not applied

```sql
-- session 1
CREATE TABLE t(id INTEGER PRIMARY KEY, data BLOB);
INSERT INTO t VALUES(1, randomblob(1048576));
SELECT dolt_commit('-A','-m','1MB');
-- session 2 (reopen)
SELECT length(data) FROM t WHERE id=1;
```

```
Assertion failed: (cnt==(cnt&0x1ffff)), function seekAndRead, file os_unix.c, line 3470.
```

I expected this to be a genuine conflict between stock's page-oriented VFS and doltlite's whole-chunk reads. It is not. **Everywhere else doltlite moves bulk bytes it already slices at 65536:**

- `chunk_store_commit.c:419` — `int toWrite = remaining > 65536 ? 65536 : (int)remaining;`
- `chunk_wal.c:326` — WAL replay, same shape

And there is a good reason the writers do it. The write side does not merely assert the bound:

```c
/* os_unix.c seekAndWriteFd */
assert( nBuf==(nBuf&0x1ffff) );
nBuf &= 0x1ffff;          /* <-- silently truncates in a release build */
```

Two whole-chunk reads never got the same treatment, and were relying on the release build to issue a 1MB `pread` unchecked. They now go through one `csReadSliced` helper.

**Precaution, not a demonstrated fix:** the chunk-index read (`chunk_index.c:205`) reads `nIndexSize` in one call and the same argument covers it, so it uses the helper too — but I could not build a database large enough to cross the bound there (a 7.9MB store with 200k rows keeps that read just under 128KiB), so I am not claiming it as a reproduced bug.

## 2. The VDBE row bound — the assertion is what is wrong

```c
assert( pC->szRow<=65536 );  /* Maximum page size is 64KiB */
```

The comment states the rationale, and it does not hold here: doltlite has no pages, and `sqlite3BtreePayloadFetchWithSize` (already inside the `DOLTLITE_PROLLY` branch two lines above) hands over the whole record, so `szRow == payloadSize`. A 100KB row exceeds a limit derived from a page format it does not use. The assertion now applies to the paged build only.

## Testing

`doltlite_unicode_blob.sh` 41/46 → 46/46 (`1mb_blob_size`, `1mb_blob_on_branch`, `1mb_blob_after_merge`, `blob_count_after_merge` on the read bound; `100kb_length` on the row bound). Battery under an assert build 99/103 — the remainder are #2201/#2202 (in flight as #2208/#2209) and two rig artifacts in my build dir.

Release-build behavior: reads are now issued in 64KiB slices rather than one call. Short reads behave the same, since callers already treat `SQLITE_IOERR_SHORT_READ` as fatal and discard the buffer.

Last of the four assert clusters from #2204. With this, #2200–#2203 all have fixes, and an assert-enabled battery bucket becomes possible.

### DIFF
--- a/src/chunk_index.c
+++ b/src/chunk_index.c
@@ -202,7 +202,7 @@ int csReadIndex(ChunkStore *cs){
     return SQLITE_NOMEM;
   }
 
-  rc = sqlite3OsRead(cs->file.pFile, aBuf, cs->index.nIndexSize, cs->index.iIndexOffset);
+  rc = csReadSliced(cs, aBuf, (i64)cs->index.nIndexSize, cs->index.iIndexOffset);
   if( rc != SQLITE_OK ){
     sqlite3_free(aBuf);
     sqlite3_free(cs->index.aIndex);
--- a/src/chunk_store.c
+++ b/src/chunk_store.c
@@ -4,6 +4,23 @@
 
 #include "chunk_store_int.h"
 #include "../ext/blake3/blake3.h"
+
+/* Stock bounds a single VFS read at 128KiB -- os_unix asserts it, and the
+** write side masks the length outright -- so whole-buffer reads slice the way
+** the commit and WAL-replay writers already do. */
+#define CS_IO_SLICE 65536
+int csReadSliced(ChunkStore *cs, void *pBuf, i64 nByte, i64 iOff){
+  u8 *p = (u8*)pBuf;
+  while( nByte > 0 ){
+    int n = nByte > CS_IO_SLICE ? CS_IO_SLICE : (int)nByte;
+    int rc = sqlite3OsRead(cs->file.pFile, p, n, iOff);
+    if( rc != SQLITE_OK ) return rc;
+    p += n;
+    iOff += n;
+    nByte -= n;
+  }
+  return SQLITE_OK;
+}
 #ifdef SQLITE_CRASH_TEST
 #endif
 
@@ -822,7 +839,7 @@ int chunkStoreGet(
       pBuf = (u8 *)sqlite3_malloc(sz + 4);
       if( pBuf == 0 ) return SQLITE_NOMEM;
 
-      rc = sqlite3OsRead(cs->file.pFile, pBuf, sz + 4, fileOff);
+      rc = csReadSliced(cs, pBuf, (i64)sz + 4, fileOff);
       if( rc != SQLITE_OK ){
         sqlite3_free(pBuf);
         return rc;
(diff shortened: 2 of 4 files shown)


timsehn:
`Build-Test / ubuntu` is `vc_ref_mutation_stress_test`, `worker 2 failed rc=5` — `SQLITE_BUSY`, a lock timeout. The same test failed the same way with `rc=5` on #2207, which changed no I/O at all (a vtab int-key read gate), so it flakes on its own: several workers doing concurrent merges against a 5-second busy timeout.

**But slicing reads deserves the atomicity question anyway**, since a single `pread` is one syscall and four are not. Why the window does not exist:

- Chunk payload bytes are immutable once written. The store is append-only and content-addressed, so no concurrent writer rewrites bytes at an offset another reader is mid-read on.
- The one thing that does rewrite the file is gc, and it builds a new file and renames. An open fd keeps reading the old inode, so a read in progress — sliced or not — sees a stable snapshot. A reader that needs to notice the replacement does so through the moved-file check, not through the read.
- A torn read would surface as `SQLITE_CORRUPT` or a stored-length/hash mismatch, which is what the callers check immediately after. `SQLITE_BUSY` cannot come from this path.

So the failure mode does not match, and the mechanism is absent. Rerunning.

Worth noting this test has now flaked on two of my PRs today; if it keeps up it deserves its own issue with a measured rate, the way #2206 got one.

PR #2223, opened 2026-08-13
Encode IEEE -0.0 as +0.0 in numeric sort keys

timsehn:
`encodeNumeric` kept the IEEE sign bit, so `0.0` and `-0.0` became different prolly keys.

SQLite treats them as equal. UNIQUE already rejected the second insert via Mem compare, but a seek or indexed `COUNT` of `-0.0` missed the `+0.0` row.

**Fail-before** (stock sqlite3 vs doltlite on current master):
```
CREATE TABLE t(id REAL PRIMARY KEY, v INT) WITHOUT ROWID;
INSERT INTO t VALUES (0.0, 1);
SELECT v FROM t WHERE id = -0.0;   -- sqlite: 1; doltlite: empty
CREATE TABLE u(id INTEGER PRIMARY KEY, v REAL);
CREATE INDEX uv ON u(v);
INSERT INTO u VALUES (1, 0.0), (2, 1.5), (3, -0.0);
SELECT count(*) FROM u WHERE v = -0.0;            -- sqlite: 2; doltlite: 0
SELECT count(*) FROM u NOT INDEXED WHERE v = -0.0; -- both: 2
```

**Fix:** canonicalize IEEE `-0.0` to `+0.0` before the sortable-bytes transform.

C regression `negzero_sortkey_eq`: 5 fail before, 22/22 after.

### DIFF
--- a/src/sortkey.c
+++ b/src/sortkey.c
@@ -156,10 +156,14 @@ static int encodeNumeric(u8 *pOut, u32 serialType, const u8 *pData, u32 nData,
   pOut[0] = SORTKEY_NUM;
 
   /* Convert SQLite numeric serial types to sortable IEEE bytes: positive
-  ** values flip the sign bit, negative values invert all bits. */
+  ** values flip the sign bit, negative values invert all bits. IEEE -0.0
+  ** compares equal to +0.0, so both use the +0.0 encoding. */
   if( serialType == 7 ){
-
     memcpy(buf, pData, 8);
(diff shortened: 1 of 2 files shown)


timsehn:
Sanitizers / asan-ubsan-oracles hit the 25-minute step timeout during VC oracle tests. Every suite that finished was green (no ASan/UBSan findings). Rerunning the failed job.

PR #2227, opened 2026-08-13
Hash prolly nodes on BLAKE3's wide path

timsehn:
## What

Node hashing is the largest single cost in a DoltLite commit, and it has been running at BLAKE3's *portable* speed on every platform — no full-size prolly node ever reached a SIMD kernel.

## Why it was slow

Two independent reasons, both in the vendored BLAKE3:

1. **`blake3_hasher_update()` is a streaming API.** It splits its input into power-of-two-aligned subtrees and compresses them one after another. An average 3712-byte prolly node becomes a 2-chunk subtree, then a 1-chunk subtree, then a partial chunk.
2. **Every `hash_many_*()` backend vectorises only whole groups of its lane count** and hashes the remainder one chunk at a time. A batch of 2 or 3 chunks never reaches the vector path at all.

Together: a 3.7 KB node hashes entirely serially. Measured on M1 Pro, BLAKE3 throughput has a cliff at exactly one lane group:

| buffer | before |
|---:|---:|
| 3584 B | 770 MB/s |
| 3799 B | 771 MB/s |
| **4096 B** | **1596 MB/s** |

A 3799-byte buffer took *nearly twice as long* as a 4096-byte one. The prolly chunker targets 4096 with a Weibull-4 boundary, whose mean is 3712 — so ~58% of nodes land under the cliff.

## The change

- `blake3_hash_oneshot()` hands a whole buffer to the wide compressor, so every full chunk lands in a single `hash_many()` call. `prollyHashCompute()` uses it.
- `compress_chunks_parallel()` rounds that batch up to the next lane count with a dummy chunk whose chaining value is discarded — at worst the cost of the serial remainder it replaces.

Both are needed: the one-shot creates the 3-chunk batch, the padding vectorises it. The genuinely streaming callers (manifest sealing, WAL replay) keep the hasher API.

x86 gets the same effect at SSE4.1's 4 lanes, and additionally pulls 5–7 chunk buffers onto AVX2's 8.

## Correctness

**No hash value changes.** Verified identical to the pristine vendored code for **every input length from 0 to 40000 bytes**, through both the new entry point and the modified streaming path (plus 64-byte XOF output at sampled lengths). The KAT suite gains pinned vectors at 2048, 3712 and 5120 bytes — chunk counts that now take the padded path — generated from the pristine implementation.

Green locally: `blake3_kat_test.sh` (9/9), `run_doltlite_tests.sh` (103/103 suites, 158/158 guards), `run_c_tests.sh` (29/29), `make lint`. Amalgamation compiles with the same 83 warnings as master (no new ones).

## Numbers

Paired medians over 9–13 alternating runs, 100k rows, candidate/base vs the merge base:

| in-memory writes | int PK | composite PK |
|---|---:|---:|
| `oltp_write_only` | 0.934× | 0.946× |
| `oltp_update_index` | 0.951× | 0.966× |
| `oltp_delete_insert` | 0.955× | 0.953× |
| `oltp_update_non_index` | 0.940× | — |
| `oltp_insert` | 0.952× | — |
| `oltp_bulk_insert` | 0.976× | — |
| `oltp_read_write` | 0.972× | — |
| **file-backed writes** | 0.949× | 0.952× |

Reads are unaffected (`oltp_point_select` 0.994×, `table_scan` 1.000×) — they don't hash.

This is aimed at the nightly's one FAIL, in-memory composite-PK `oltp_write_only` at 2.7× (ceiling 2.4×). It moves that to roughly 2.55× — an improvement on every write metric, but not on its own enough to clear the ceiling.

## Context for the remaining gap

Phase-timing the failing workload (4000 statements, 100k-row table, in-memory) shows where the 2.7× actually lives:

| phase | DoltLite | SQLite |
|---|---:|---:|
| statements | 14.5 ms | 12.8 ms |
| COMMIT | 21.4 ms | 0.7 ms |

Statement execution is at parity or better in 5 of 6 write workloads — the entire gap is `flushAllPending`. And that commit is not doing wasted work: it rewrites 11.46 MB in 2943 chunks, which is ~64% of the tree, matching exactly the fraction of leaves that 3000 scattered edits should dirty. Dedup hits during the commit: 3. The strategy counters confirm there is no aborted-then-retried tree walk.

So the remaining lever is hash throughput per byte, and the biggest one left is the chunk-size cliff above: BLAKE3 vectorises in 1 KiB units, the chunker targets 4096 with a distribution whose mean is 3712, and ~58% of nodes therefore sit just under a lane group. Raising `PROLLY_WEIBULL_L` so most nodes clear 4 chunks would roughly double hash throughput again (and cut node count ~25%), at the cost of ~12% more bytes rewritten per dirty leaf. That changes every tree shape, so it's a `CHUNK_STORE_VERSION` conversation, not something to slip into a perf PR — flagging it as the next move rather than doing it.

### DIFF
--- a/.github/workflows/sanitizers.yml
+++ b/.github/workflows/sanitizers.yml
@@ -225,6 +225,12 @@ jobs:
         ./threadtest5 'file:/tmp/doltlite-threadtest5.db?mode=rwc' -num-workers 8
 
     - name: Run version-control concurrency under TSan
+      env:
+        # Same reason threadtest5 above gets a longer busy window: three forked
+        # workers contending on ref updates make far less progress per second
+        # under TSan, and the stress test's default retry budget is a wall
+        # clock, so it has to be told to wait proportionally longer.
+        DOLTLITE_STRESS_BUSY_MS: "120000"
       run: |
         cd build
         ./vc_concurrency_test
--- a/ext/blake3/README.md
+++ b/ext/blake3/README.md
@@ -10,7 +10,7 @@ prolly chunks. The vendored sources live in this directory.
 | Upstream | https://github.com/BLAKE3-team/BLAKE3 |
 | Version  | 1.8.5 |
 | Files vendored | `blake3.h`, `blake3.c`, `blake3_portable.c`, `blake3_impl.h`, `blake3_dispatch.c`, `blake3_sse2.c`, `blake3_sse41.c`, `blake3_avx2.c`, `blake3_avx512.c`, `blake3_neon.c` (verbatim from upstream `c/`, with an SPDX header prepended) |
-| Files modified | `blake3.c` and `blake3.h` have the BLAKE3_USE_TBB threading code stripped — DoltLite doesn't link TBB |
+| Files modified | `blake3.c` and `blake3.h` have the BLAKE3_USE_TBB threading code stripped — DoltLite doesn't link TBB, plus the two local additions under *Small-buffer throughput* below |
 | License | Apache 2.0 with LLVM exception, OR CC0 1.0 (dual-licensed by upstream). DoltLite redistributes under Apache 2.0 (project-wide). See `LICENSE`. |
 
 ## SIMD paths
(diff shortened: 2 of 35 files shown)


timsehn:

timsehn:
@itoqa — feedback on the diff report for `94c0cef` → `a1cfede`.

The report attributes a new **Scope** failure to this PR ("this PR introduces a medium-severity verification flaw"). I don't think that attribution holds, and the reason looks systematic rather than specific to this finding.

**The analyzed diff is a merge of `master` into the branch, not a change by this PR.** This PR's own change is byte-identical across `94c0cef` and `a1cfede` — the merge commit only brings master's content in. The commits it introduced:

```
45c1c8128d  Fail closed on constraint detector errors      (#2232)
e7c0e64c18  Keep verify output-only state read-only        (#2231)
368d003942  Rebuild ours-only indexes after merge          (#2228)
ddd9c58658  Keep the default dolt_docs AGENT synthetic     (#2222)
dab188cbd4  Give the ref-mutation stress test a budget     (#2233)
```

`src/doltlite_verify_constraints.c` changed by +94/-… in that set — the exact subsystem the finding concerns. This PR touches only `ext/blake3/*`, `src/prolly_hash.c`, and `test/blake3_kat*`, and its functional change is proven hash-output-identical for every input length 0–40000. It cannot affect constraint verification scope.

**Empirically, master alone behaves the same — and behaves correctly.** Building `master` (`23a6001`) in a separate worktree and running the described scenario (two tables with FK violations, resolve `orders`, run default `dolt_verify_constraints()`):

| | master `23a6001` | this PR `a1cfede` |
|---|---|---|
| catalog after resolve + default verify | `users:1`, count=1 | `users:1`, count=1 |

Both retain the unrelated `users` finding — the behavior the report says was lost. The project's own oracles for this surface, run against real Dolt 2.2.2, also agree on both builds:

| suite | master | this PR |
|---|---|---|
| `vc_oracle_verify_constraints` | 15/15 | 15/15 |
| `vc_oracle_constraint_violations` | 11/11 | 11/11 |

I could not reproduce the failure on either build, which is not the same as proving it can't happen — my repro is one reading of the prose description. The report's own note that "the native reproduction returned successfully but produced no catalog rows" suggests the reproduction didn't establish the preconditions, so the finding may not be real at all.

**The actionable request:** when a branch has `master` merged into it, please scope attribution to the branch's own commits rather than to everything new in the diff. As written, any PR that merges an up-to-date master — which this repo's ruleset *requires* before merging, since it sets `strict: true` — will inherit blame for whatever landed on master in the meantime. That will misattribute findings on essentially every PR here.

If you have the exact SQL from the failing reproduction, I'm glad to run it against both builds and report back.

PR #2229, opened 2026-08-13
Refuse to persist a conflicted working set from ref updates

timsehn:
A conflicted merge inside `BEGIN` saves the unfinished working set in memory. `dolt_branch` / `dolt_tag` then serialized the whole refs blob and committed it, so the unfinished merge survived reopen. That breaks the rule that conflicts are never persisted.

`PersistWorkingSet` and commit phase one already refuse a non-empty conflicts catalog. `doltliteMutateRefsExpected` now walks the working sets about to land and returns `SQLITE_CONSTRAINT` if any still names one. Abort helpers that rewrite a clean working set can still finish.

**Fail-before:**

```sql
BEGIN;
SELECT dolt_merge('br');          -- conflicts
SELECT dolt_branch('backup');     -- succeeded, wrote them
ROLLBACK;
-- reopen: backup exists and/or merge state is inherited
```

**After:** `dolt_branch` / `dolt_tag` error with `cannot update refs: unresolved merge conflicts`. Reopen has no backup/tag, no merge, no conflicts.

`doltlite_conflicts.sh` 54/54, `doltlite_branch.sh` 66/66, `doltlite_merge.sh` 134/134, `doltlite_cherry_pick.sh` 98/98, `doltlite_rebase.sh` 36/36.

### DIFF
--- a/src/doltlite_branch.c
+++ b/src/doltlite_branch.c
@@ -51,6 +51,12 @@ static void branchNamedResultError(
   const char *zNotFound,
   const char *zExists
 ){
+  sqlite3 *db = sqlite3_context_db_handle(ctx);
+  if( rc==SQLITE_CONSTRAINT && doltliteSessionHasUnresolvedConflicts(db) ){
+    branchError(ctx, bHadSavepoint,
+      "cannot update refs: unresolved merge conflicts");
+    return;
+  }
   branchSealSavepointsOnError(ctx, bHadSavepoint);
   doltliteRefResultError(ctx, rc, zNotFound, zExists);
 }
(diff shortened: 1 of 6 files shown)


timsehn:

timsehn:
@itoqa The savepoint finding is real, but it is intended Dolt-parity, not a defect in this PR.

Reproduced here and on Dolt:

```sql
BEGIN;
SAVEPOINT sp;
SELECT dolt_branch('rolled_back');  -- CALL dolt_branch(...) on Dolt
ROLLBACK TO sp;
```

Both return `no such savepoint: sp` / `SAVEPOINT sp does not exist`, and `rolled_back` is still listed after the outer `ROLLBACK` and a reopen. `BEGIN` + `dolt_branch` + `ROLLBACK` (no savepoint) also leaves the branch.

`dolt_branch` commits the refs blob immediately (`chunkStoreCommit`), then `doltliteVcSealBranchStyleTxn` releases any active savepoints (or `COMMIT`/`BEGIN` for a plain transaction). That is the same contract as Dolt: ref updates are durable and they consume the savepoint so a later `ROLLBACK TO` cannot pretend they were transactional.

This PR only refuses to persist a conflicted working set. It does not change that seal path. No change here.

PR #2230, opened 2026-08-13
Do not infer revert from a commit message prefix

timsehn:
`applyMergedCatalogAndCommit` treated any message starting with `Revert` as `dolt_revert`. Linear rebase (and cherry-pick of such a commit) then set `bPreferOurMaster`, skipped index patches, and left a `DROP INDEX` unapplied.

Pass an explicit flag instead: `dolt_revert` sets it, cherry-pick and rebase do not. Interactive rebase already merged with prefer-ours off.

**Fail-before:** cherry-pick or rebase a `DROP INDEX` committed as `Revert leftover` — the index is still there.

**After:** the index is gone. A real `dolt_revert` of a drop still restores it.

`doltlite_cherry_pick.sh` 103/103, `doltlite_rebase.sh` 38/38, `doltlite_rebase_schema.sh` 23/23.

### DIFF
--- a/src/doltlite_cherry_pick.c
+++ b/src/doltlite_cherry_pick.c
@@ -103,6 +103,7 @@ int applyMergedCatalogAndCommit(
   const ProllyHash *ourHead,
   const ProllyHash *pCommitOurCatHash,
   const char *zMessage,
+  int bPreferOurMaster,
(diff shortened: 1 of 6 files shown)


timsehn:

PR #2235, opened 2026-08-13
Refuse same-table dirty revert when only indexes changed

timsehn:
## Why

Ito found this on #2230; it is pre-existing and not caused by that PR. Dolt refuses the same script.

```sql
CREATE TABLE t(...);
CREATE INDEX t_v ON t(v);
-- commit, DROP INDEX, commit
INSERT INTO t VALUES(99,'unrelated dirty');
SELECT dolt_revert('HEAD');
```

Dolt: `Your local changes would be overwritten by revert.`

DoltLite used to enter the split revert path because `DROP INDEX` is a separate catalog object, so the table-entry overlap check did not see a revert of `t`. That path wrote the restored index into the working set, then failed the commit-side merge (`nReindexC > 0`) and restored only in memory. Reopen reloaded the half-applied catalog: index count went from 0 to 1 while the session still looked rolled back.

## Fix

- Overlap gate: a dirty table also conflicts with a revert that changes that table's indexes (`doltliteIndexSchemaRowsDifferForTable`), matching Dolt.
- `doltliteRestoreTxnState` now persists the restored working set, so a failed apply cannot survive reopen.
- Unrelated dirty work (another table) plus an index revert still succeeds and keeps the side table, as before.

## Tests

`test/doltlite_revert_dirty.sh`: same-table dirty + drop-index revert is refused and stays refused after reopen; other-table dirty + drop-index revert still applies and keeps `meta`.

Also ran: cherry-pick 98, merge 137, conflicts 47, savepoint 128, branch 66, merge_status 31, rebase 36.

### DIFF
--- a/src/doltlite_cherry_pick.c
+++ b/src/doltlite_cherry_pick.c
@@ -319,7 +319,13 @@ apply_rollback:
     chunkStoreUnlock(cs);
   }
   {
-    return doltliteRestoreTxnStateOnFailure(db, &savedState, rc);
+    ProllyHash restoredCat = savedState.sessionCatalogHash;
+    int restoreRc = doltliteRestoreTxnStateOnFailure(db, &savedState, rc);
+    if( restoreRc==rc && !prollyHashIsEmpty(&restoredCat) ){
+      int persistRc = doltlitePersistWorkingSetWithHash(db, &restoredCat);
+      if( persistRc!=SQLITE_OK && persistRc!=SQLITE_NOMEM ) restoreRc = persistRc;
+    }
+    return restoreRc;
   }
 }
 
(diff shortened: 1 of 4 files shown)


timsehn:
@itoqa The missing-master-root finding is real, but it is pre-existing loader policy, not this PR.

`loadSchemaFromCatalog` (`src/doltlite_schema_diff.c`) has always treated a missing or empty `iTable==1` root as success with zero entries. Merge, status, diff, add, and checkout all go through that same helper. An empty catalog hash is a legitimate “no schema” (new / empty DB). A non-empty catalog blob with no master table is corrupt — that distinction belongs in the shared loader, not in the revert overlap gate.

This PR only *calls* that helper so a dirty table can see index-only revert overlap. It does not change how a damaged catalog is classified. No change here; a follow-up can make the loader return `SQLITE_CORRUPT` for a missing master on a non-empty catalog.

PR #2265, opened 2026-08-18
Read every merge side at merged column positions

timsehn:
A merge where one side drops a column and the other only writes rows silently corrupts data. Only the both-sides-changed-schema path ever moved a side into the merged layout, so the side that kept the old layout is read at the merged column positions and every value right of the dropped column shifts one slot left.

On `t(k,a,b,c)` where main drops `b` and a branch inserts `(2,'a2','b2','c2')`:

```
merged row 2 -> a='a2'  c='b2'      the value 'c2' is gone
```

The merge returns a commit hash, records no conflict, and prints no warning. Dolt 2.2.2 returns `c2`. It happens in both merge directions — when the dropping side is the one merged in, it is our own rows that are read at the wrong positions.

The ancestor is misread the same way. For a row both sides hold, comparing against an ancestor laid out differently reports a cell as changed on both sides, so a merge Dolt completes cleanly stops with a conflict:

```
main drops b, branch does UPDATE t SET c='c2new' WHERE k=2
doltlite: cannot merge: conflicts detected      dolt: merge successful, c='c2new'
```

## What changed

The side whose schema was not adopted, and the ancestor, are re-laid into the merged layout before rows are merged — in both directions, and in the dual-schema-change path that previously moved only theirs.

The layout mapping could not express a drop at all: a source column missing from the merged layout was appended as a new column, which would restore dropped data as a trailing field no column names. It is discarded when the ancestor had it, since only a drop accounts for a column that the ancestor has and the merged layout does not. A side already sitting at its merged positions is returned untouched, so add-column merges do no extra work and keep their current behaviour.

## Testing

`test/doltlite_schema_merge.sh` gains cases for both directions, the shared-row edit that used to conflict, the dual-schema-change variant, and a trailing-column drop. Expected values were taken from Dolt 2.2.2.

- Fail-before/pass-after: 7 of the new assertions fail on master (`126 passed, 7 failed`), all pass with the fix (`133 passed, 0 failed`).
- Full shell battery: **104/104 suites**, including the C regression suite at 310853/310853 assertions.
- Suite passes identically under an ASAN+UBSAN build.
- Differential sweep vs stock SQLite 3.54.0, all feature groups: 150/150 seeds.
- `vc_oracle_merge_test` and `vc_oracle_schema_merge_test` have pre-existing failures; counts are byte-identical before and after this change.

The trailing-column case earns its own test because it is the only shape that reaches the discard path — for a middle column, a surviving column later claims the same slot and overwrites it. Reaching that path unguarded is a stack-buffer underflow (`aMem[-1]`), which UBSAN and ASAN both report; the bounds guard is covered by that test.

### DIFF
--- a/src/doltlite_merge_int.h
+++ b/src/doltlite_merge_int.h
@@ -229,7 +229,7 @@ void mergeColDefaultsFree(MergeColDefaults *p);
 int mergeColDefaultsLoad(const char *zSql, const char *zTable,
                          MergeColDefaults *pOut);
 
-int normalizeTheirsToMergedLayout(
+int normalizeSideToMergedLayout(
   sqlite3 *db,
   const char *zTable,
   const ProllyHash *pOursRoot,
--- a/src/doltlite_merge_pass1.c
+++ b/src/doltlite_merge_pass1.c
@@ -578,6 +578,55 @@ static int mergePass1CheckPrimaryKeysMatch(
   return SQLITE_ERROR;
 }
 
+/* Exactly one side changed the table's columns, so the merged layout is that
+** side's. The other side and the ancestor still hold records in the old
+** layout, and the row merge addresses every side by merged column position,
+** so both are re-laid out first: otherwise a column the merged layout no
+** longer has shifts every later value one slot left, and an ancestor read at
+** the wrong positions reports cells as changed that neither side touched. */
+static int mergePass1RelayoutOneSidedSchema(
+  MergePass1Ctx *c,
+  const char *zName,
+  int bMergedIsOurs,
+  struct TableEntry *pOurs,
+  struct TableEntry *pAnc,
+  struct TableEntry *pTheirs,
+  ProllyHash *pOtherOut,
+  ProllyHash *pAncOut,
+  int *pbRelaid
+){
+  SchemaEntry *ourSE = findSchemaEntry(c->aOursSchema, c->nOursSchema, zName);
+  SchemaEntry *theirSE = findSchemaEntry(c->aTheirsSchema, c->nTheirsSchema, zName);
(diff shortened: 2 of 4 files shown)


timsehn:
Thanks — this is a **real bug**, and the reproduction is correct. It is **not caused by this PR**: it reproduces identically on unfixed master, so I've filed it separately as #2268 rather than expanding this PR.

Evidence that it predates this branch — same script, both binaries, and the merge produces the *byte-identical commit hash*:

```
FIXED (this branch)   cc62929b7171cb5ee38925d5ce6d63d4378fb043
                      columns: k,a,b,renamed_c,d      row 1: a1 / 101 / NULL
CONTROL (master)      cc62929b7171cb5ee38925d5ce6d63d4378fb043
                      columns: k,a,b,renamed_c,d      row 1: a1 / 101 / NULL
```

An identical hash means this code path never reaches the mapping change in this PR. The reason is that both sides change the schema here (one drops, one renames), so the merge takes the dual-schema-change path, and this PR's relayout runs only when exactly one side changed columns.

The analysis points at the positional rename fallback, and that part is right, but the defect is upstream of the record mapping. With **no row changes on either side**, the merged schema is already wrong, so it is schema merge that is discarding the drop:

```
drop b  vs  rename c -> renamed_c   =>  k,a,b,renamed_c,d     drop discarded
drop b  vs  add column e            =>  k,a,c,d,e             correct
```

Dolt 2.2.2 gives `k, a, renamed_c, d` with row 1 = `a1 | c1 | 101`. Full detail and repro in #2268.

One correction to the report: the code is cited as `normalizeTheirsToMergedLayout ... renamed to normalizeSideToMergedLayout so either side can be rewritten` — that rename is this PR's, but the positional fallback it describes is unchanged pre-existing code, not something this PR adds.

Issue #2016, opened 2026-08-07
dolt_diff and dolt_diff_summary cannot tell a table rename from a drop plus a create


timsehn:
`dolt_diff_summary` calls a drop+create a rename, and `dolt_diff` splits a rename into a drop plus an add. Both are the same root cause: the catalog has no table identity that survives a rename, so neither surface can tell the two apart.

The `dolt_diff_summary` half is the one that matters — it is merged and returns a wrong answer today, not a cosmetic difference.

Pre-existing on master (`069a59e7d6`). Found while looking at the `dolt_diff` rename pairing. Related: #1999, which is the same trap on `dolt_diff_stat`.

## Repro 1 — `dolt_diff_summary` reports a rename that never happened

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, v TEXT);
INSERT INTO t VALUES(1,'a');
SELECT dolt_commit('-A','-m','c1');
DROP TABLE t;
CREATE TABLE u(id INTEGER PRIMARY KEY, w TEXT);   -- different table, different columns
INSERT INTO u VALUES(5,'z');
SELECT dolt_commit('-A','-m','drop_add');

SELECT * FROM dolt_diff_summary('<c1>','<drop_add>');
```

| | result |
|---|---|
| doltlite | `t, u, renamed, 1, 1` |
| Dolt 2.2.2 | `t, "", dropped, true, true`<br>`"", u, added, true, true` |

`t` was dropped and an unrelated `u` was created. doltlite claims one was renamed into the other.

## Repro 2 — `dolt_diff` splits a rename into two rows

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, v TEXT);
INSERT INTO t VALUES(1,'a');
SELECT dolt_commit('-A','-m','c1');
ALTER TABLE t RENAME TO u;
INSERT INTO u VALUES(2,'b');
SELECT dolt_commit('-A','-m','rename_and_insert');

SELECT message, data_change, schema_change, table_name FROM dolt_diff;
```

| | rows for the rename commit |
|---|---|
| doltlite | `rename_and_insert, true, true, u`<br>`rename_and_insert, true, true, t` |
| Dolt 2.2.2 | `rename_and_insert, true, true, u` |

## Root cause

`struct TableEntry` (`src/doltlite_catalog_types.h`) carries only `iTable` and `zName`. Neither is an identity:

- `zName` changes on rename, by definition.
[...]

Issue #1999, opened 2026-08-06
dolt_diff_stat emits an all-zero row for an in-place column rename where Dolt emits none


timsehn:
Cosmetic parity nit, likely low priority — an all-zero row and no row arguably say the same thing. Filing mainly to record the characterization, and the trap in the obvious fix.

Pre-existing on master. Found while doing #1998.

## Repro

```sql
CREATE TABLE t(id INT PRIMARY KEY, v INT);
INSERT INTO t VALUES(1,10),(2,20);
SELECT dolt_commit('-Am','seed');
ALTER TABLE t RENAME COLUMN v TO w;
SELECT dolt_commit('-Am','c2');
SELECT * FROM dolt_diff_stat('HEAD~1','HEAD');
```

doltlite returns one row, everything zero including the row counts:

```
t  ru=0 ra=0 rd=0 rm=0  ca=0 cd=0 cm=0  orc=0 nrc=0
```

Dolt 2.2.2 returns no rows. Same on an empty table, and for `ALTER TABLE t MODIFY COLUMN v VARCHAR(50)`.

## Do not "fix" it by narrowing the emission gate

The tempting change is to drop the `schemaChanged` exception in `dstAdvance` (`src/doltlite_diff_stat.c`), so a table with no counters gets no row. I tried it. It makes this case pass and **breaks `diff_stat_revert_schema_change_with_later_added_table`** in `test/vc_oracle_diff_stat_test.sh`, where Dolt *does* emit an all-zero row.

What Dolt actually does, as far as I probed:

| schema-only change, no data change | Dolt |
|---|---|
| `ALTER TABLE t RENAME COLUMN v TO w` | no row |
| `ALTER TABLE t MODIFY COLUMN v VARCHAR(50)` | no row |
| recreate with a CHECK, `DROP TABLE t`, `RENAME TO t` | all-zero row |

So the row appears when the table was *recreated*, not when its columns changed in place. Matching that needs table-identity reasoning rather than a counter test, which is why #1998 left it alone.

`rename_column_no_data_change` in `test/vc_oracle_diff_stat_test.sh` currently runs summary-only for this reason, with a comment pointing at the trap. If this is closed as wontfix, that case can stay as it is.

PR #2269, opened 2026-08-18
Keep a column deletion when a rename adopts the other schema

### LINKED ISSUE #2268: Merge of a column drop against a column rename keeps the dropped column and shifts values
Merging a branch that renamed a column into a branch that dropped a *different* column silently discards the drop. The merged schema keeps the dropped column, and because the records were written without it, every value from the drop position rightwards is read one column off.

Pre-existing on master (`c36ccf6237`); found by ito QA on #2265 and confirmed to be unrelated to that PR — the merge produces the byte-identical commit hash `cc62929b7171cb5ee38925d5ce6d63d4378fb043` with and without that branch's changes.

## Reproduce

```sql
CREATE TABLE t(k INT PRIMARY KEY, a TEXT, b TEXT, c TEXT, d INT);
INSERT INTO t VALUES(1,'a1','b1','c1',101);
SELECT dolt_commit('-A','-m','base');
SELECT dolt_branch('feat');
-- main:
ALTER TABLE t DROP COLUMN b;              SELECT dolt_commit('-A','-m','drop b');
-- feat:
ALTER TABLE t RENAME COLUMN c TO renamed_c;
INSERT INTO t VALUES(2,'a2','b2','c2',202);
SELECT dolt_commit('-A','-m','rename c');
-- main:
SELECT dolt_merge('feat');
```

Observed — merge succeeds, reports no conflict:

```
columns: k, a, b, renamed_c, d          <- b was dropped on main and is still here
row 1:   a='a1'  renamed_c=101  d=NULL  <- d's value surfaced as renamed_c
```

Dolt 2.2.2, same script: `merge successful`, columns `k, a, renamed_c, d`, row 1 `a1 | c1 | 101`.

## The trigger is drop-vs-rename, and rows are not involved

With no row changes on either side, the merged schema is still wrong, which places this in schema merge rather than in record layout:

```
drop b  vs  rename c -> renamed_c   =>  k,a,b,renamed_c,d     WRONG, drop discarded
drop b  vs  add column e            =>  k,a,c,d,e             correct
```

## Notes toward a fix

The positional rename fallback in the layout mapping (`src/doltlite_merge_schema.c`, the `ai = j` branch that pairs a source column against the ancestor column at the same index when the names differ and the definitions match) cannot be right once a drop has shifted positions between the two sides: `renamed_c` at its post-drop index pairs against the wrong ancestor slot. Any fix likely has to resolve the drop before applying rename pairing, rather than pairing by index.

Related and adjacent, but distinct: #2265 fixes the case where only one side changes columns, where the records rather than the schema were misread.

timsehn:
Fixes #2268, found by ito QA on #2265 and confirmed there to be pre-existing.

Merging a branch that renamed a column into a branch that deleted a *different* column brought the deleted column back and put values under the wrong names:

```sql
-- base: t(k, a, b, c, d), row (1,'a1','b1','c1',101)
-- feat: ALTER TABLE t RENAME COLUMN c TO renamed_c; INSERT (2,'a2','b2','c2',202)
-- main: ALTER TABLE t DROP COLUMN b
SELECT dolt_merge('feat');
```

| | columns | row 1 |
|---|---|---|
| before | `k, a, b, renamed_c, d` | `a1`, `renamed_c=101`, `d=NULL` |
| after / Dolt 2.2.2 | `k, a, renamed_c, d` | `a1`, `renamed_c=c1`, `d=101` |

The merge returned a commit hash and reported no conflict, and `PRAGMA integrity_check` was red afterwards. A variant where their side also edits a row both sides hold stopped with a conflict Dolt does not raise.

## Cause

A rename is indistinguishable from a drop plus an add by comparing catalogs, so when `trySchemaColumnMerge` recognizes one it selects that side's schema wholesale. Two things were then lost:

- **The other side's deletions.** Only the columns that side *added* were replayed onto the adopted schema (`ALTER TABLE ... ADD COLUMN`); columns it had deleted came back.
- **The other side's row layout.** Adopting their schema makes the merged layout theirs, but our records still followed ours, and the row merge addresses every side by merged column position. That is what shifted values under the wrong names and dropped the last one, and what made an untouched cell read as edited on both sides.

## Change

Deletions are now carried across as drop actions, applied alongside the added columns once the merged catalog is live, and our rows are re-laid onto the adopted layout before the row merge. A column the other side renamed is absent under its old name too; its replacement sitting at the same ancestor position is what separates a rename from a deletion.

## Not fixed here

This does not give a column an identity that survives a rename, which is what would settle the question outright rather than inferring it. That is the same gap behind the table-level confusion in #2016 and #1999.

The mirror pairing — *ours* renames, *theirs* deletes — still stops with a schema conflict where Dolt merges cleanly, because the rename search only scans their columns. That is pre-existing, unchanged by this PR, and fails safe rather than corrupting; I've noted it on #2268.

## Testing

`test/doltlite_schema_merge.sh` gains the reported case, the edit variant that used to conflict, and a rename with no deletion opposite it as a guard that unrelated columns are still kept. Expected values taken from Dolt 2.2.2.

- Fail-before/pass-after: 6 of the new assertions fail on master (`140 passed, 6 failed`), all pass with the fix (`146 passed, 0 failed`). One of the six is `integrity_check`.
- Full shell battery: **104/104 suites**, including the C regression suite.
- Same suite clean under an ASAN+UBSAN build, no sanitizer reports.

### DIFF
--- a/src/doltlite_core.c
+++ b/src/doltlite_core.c
@@ -492,6 +492,10 @@ void freeSchemaMergeActions(SchemaMergeAction *a, int n){
       sqlite3_free(a[i].azAddColumns[j]);
     }
     sqlite3_free(a[i].azAddColumns);
+    for(j=0; j<a[i].nDropColumns; j++){
+      sqlite3_free(a[i].azDropColumns[j]);
+    }
+    sqlite3_free(a[i].azDropColumns);
     sqlite3_free(a[i].zTableName);
   }
   sqlite3_free(a);
--- a/src/doltlite_internal.h
+++ b/src/doltlite_internal.h
@@ -1308,6 +1308,8 @@ struct SchemaMergeAction {
   char *zTableName;
   char **azAddColumns;
   int nAddColumns;
+  char **azDropColumns;
+  int nDropColumns;
 };
 
 void freeSchemaMergeActions(SchemaMergeAction *a, int n);
--- a/src/doltlite_merge.c
+++ b/src/doltlite_merge.c
@@ -369,12 +369,14 @@ static int preDetectIndexSchemaConflicts(
   return SQLITE_OK;
 }
 
-int recordSchemaAddColumns(
+int recordSchemaColumnChanges(
   SchemaMergeAction **ppSchemaActions,
   int *pnSchemaActions,
   const char *zName,
   char **azAddCols,
-  int nAddCols
+  int nAddCols,
+  char **azDropCols,
+  int nDropCols
 ){
   SchemaMergeAction *aNew;
   aNew = sqlite3_realloc(*ppSchemaActions,
@@ -385,8 +387,21 @@ int recordSchemaAddColumns(
   if( !aNew[*pnSchemaActions].zTableName ) return SQLITE_NOMEM;
   aNew[*pnSchemaActions].azAddColumns = azAddCols;
   aNew[*pnSchemaActions].nAddColumns = nAddCols;
+  aNew[*pnSchemaActions].azDropColumns = azDropCols;
+  aNew[*pnSchemaActions].nDropColumns = nDropCols;
   (*pnSchemaActions)++;
   return SQLITE_OK;
+}
+
+int recordSchemaAddColumns(
+  SchemaMergeAction **ppSchemaActions,
+  int *pnSchemaActions,
+  const char *zName,
+  char **azAddCols,
+  int nAddCols
+){
+  return recordSchemaColumnChanges(ppSchemaActions, pnSchemaActions, zName,
+                                   azAddCols, nAddCols, 0, 0);
 }
 
 int schemaEntryChangedByName(
@@ -1026,6 +1041,8 @@ int tryResolveSchemaDivergence(
   SchemaEntry *theirSchEntry;
   char **azAddCols = 0;
   int nAddCols = 0;
+  char **azDropCols = 0;
+  int nDropCols = 0;
   int schemaChoice = SCHEMA_MERGE_DEFAULT;
   int resolvedDivergence = 0;
   char *zSchemaErr = 0;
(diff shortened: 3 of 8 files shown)


timsehn:
Confirmed and fixed in c4531f2f40 — thanks, this was a real defect in the PR, not a pre-existing one.

Reproduced exactly as reported:

```
schema: CREATE TABLE t(k INTEGER PRIMARY KEY, a TEXT, renamed_c TEXT, d INT)   correct
row:    renamed_c = 505 (integer), d = NULL                                    wrong
        PRAGMA integrity_check -> NUMERIC value in t.renamed_c
```

`'c1'` was destroyed and `505` moved one column left into a TEXT column, which is what `integrity_check` was objecting to, and it survived a reopen.

**Cause.** A rename rewrites no rows, so a branch that only renames leaves its table root byte-identical to the ancestor's. I gated the row relayout on `oursChanged && theirsChanged`, so with no rows written on their side the relayout was skipped while their schema was still adopted. The catalog then held their column order against our records, and the `DROP COLUMN` was applied to whichever field had landed in the deleted column's position. The reported analysis attributes it to the mapping used by the relayout; the mapping was right, it just wasn't run.

The fix decides the move from the schema alone — never from whether either side wrote rows — and uses the moved rows on the path where no row merge runs.

**Verification.** The reported case now gives `renamed_c='c1'` (text), `d=505`, `integrity_check ok`, stable across a reopen. Because my tests all happened to attach row activity to the rename, which is what hid this, I also swept every drop/rename position pair on a five-column table: all 12 produce the right columns with every value still under its own name and a clean integrity check.

`drop_vs_rename_norows` is added as a permanent guard and asserts values, `typeof()`, and `integrity_check`. It fails on e303c6d922 — the commit tested here — on exactly those three assertions, and passes now. Full battery back to 104/104.

PR #2272, opened 2026-08-18
Refuse to republish a store this connection may not write

timsehn:
Found in the deep review. Garbage collection finishes by renaming a rebuilt file over the database path, and `VACUUM`, `PRAGMA wal_checkpoint` and `dolt_gc` all reach that code. Nothing checked whether writing was permitted, or whether the file still belonged to this handle.

## A read-only connection rewrote the database

```
$ ln r.db link.db && chmod 0444 r.db
$ doltlite "file:r.db?mode=ro" "INSERT INTO t VALUES(2,'y');"
Error: attempt to write a readonly database          correctly refused

$ doltlite "file:r.db?mode=ro" "VACUUM;"             silently succeeded
before  size=3633  mode=0444  inode=741850319  links=2
after   size=1212  mode=0644  inode=741850359  links=1
```

The file shrank, its permissions were reset, it became a different inode, and the second name for it stopped sharing the data. `PRAGMA wal_checkpoint` and `SELECT dolt_gc()` did the same. Stock SQLite refuses `VACUUM` here and leaves the file alone.

That means a "safe" read-only connection — a reporting replica, an audit sidecar — was not read-only, and an unprivileged-by-intent process could strip the mode bits off a deliberately protected database.

## A replaced file was overwritten with stale content

If the path is replaced under an open handle, that path now names a database this connection never opened. Renaming our rebuilt copy over it destroys it:

```
session: SELECT count(*) FROM t;          -- store has the file open
         (path.db is replaced by an unrelated database)
         VACUUM;                          -- silently succeeded
$ doltlite path.db "SELECT x FROM victim;"
Error: no such table: victim              the other database is gone
```

`VACUUM`, `wal_checkpoint` and `sqlite3_backup_step` all destroyed it. `dolt_gc` already refused, because it forces a refresh whose own guard catches this.

## Change

Both conditions are refused once the lock is held — the point at which a replacement has been detected. Backup gains the same check for its destination; it already refused a read-only one.

A checkpoint still answers success: compacting there is opportunistic, stock does not compact on checkpoint at all, and stock likewise answers `PRAGMA wal_checkpoint` on a database it cannot write with success rather than an error. Such a store simply stays uncompacted. `VACUUM` and `dolt_gc` report the refusal, matching stock's behaviour for a `VACUUM` it cannot write.

Reads on both kinds of connection are unaffected, and a normal writable connection still collects and vacuums as before.

## Testing

New suite `test/doltlite_readonly_file_replacement.sh`, registered in the manifest, covering all three operations against both conditions plus a read that must still work.

- Fail-before/pass-after: 5 of the 7 assertions fail on master, all pass with the fix. The two that already pass are the read, and `dolt_gc` on a replaced file.
- Backup was verified with a C harness against both a pre-fix and post-fix library: pre-fix `backup_step` returned `SQLITE_DONE` and the unrelated database was gone; post-fix it returns `SQLITE_READONLY` and that database survives.
- Stock parity re-checked directly: stock errors on `VACUUM` and succeeds on `wal_checkpoint` for a read-only database, which is what this now does.
- Full battery **106/106 suites**. New suite and the GC suite are also clean under an ASAN+UBSAN build with no sanitizer reports.

### DIFF
--- a/src/doltlite_gc.c
+++ b/src/doltlite_gc.c
@@ -980,6 +980,18 @@ static int gcRun(
     *pzPhase = "failed to acquire lock for gc";
     return rc;
   }
+  /* The sweep republishes the store by renaming a rebuilt file over this path.
+  ** A connection that may not write must not do that, and neither may one
+  ** whose file was replaced underneath it: the path now names a database this
+  ** handle never opened, and the rename would destroy it. Checked after the
+  ** refresh, which is what detects the replacement. */
+  if( cs->readOnly || cs->movedReadOnly ){
+    chunkStoreUnlock(cs);
+    *pzPhase = cs->readOnly
+             ? "attempt to write a readonly database"
+             : "cannot rewrite a database file that was replaced";
+    return SQLITE_READONLY;
+  }
   if( bForceRefresh ){
     rc = chunkStoreForceRefresh(cs);
     if( rc!=SQLITE_OK ){
--- a/src/pager_shim.c
+++ b/src/pager_shim.c
@@ -373,8 +373,11 @@ static int shimPagerCheckpoint(Pager *p, sqlite3 *db, int eMode,
     rc = doltliteGcCompactStore(db, pCs);
     /* Compacting here is opportunistic -- stock does not compact on checkpoint
     ** at all, and answers a checkpoint on a database it may not write with
-    ** success. A store this connection may not rewrite just stays uncompacted. */
-    if( rc==SQLITE_READONLY ) rc = SQLITE_OK;
+    ** success, so a read-only store just stays uncompacted. A store whose file
+    ** was replaced is not that: stock has no such state to be consistent with,
+    ** and the caller is checkpointing a path that no longer holds the database
+    ** it opened, which it can only learn if we say so. */
+    if( rc==SQLITE_READONLY && !pCs->movedReadOnly ) rc = SQLITE_OK;
   }
   if( pCs ) pCs->checkpointActive = 0;
   if( rc!=SQLITE_OK ) return rc;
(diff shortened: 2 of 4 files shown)


timsehn:
Confirmed and fixed in de4699ec59 — the reading is right, the translation was too broad.

Reproduced: a checkpoint on a replaced path returned `0|0|0` while the refusal that protected the other database was discarded.

My justification for answering a checkpoint with success was stock parity, and that argument only ever covered a store this connection merely may not write — stock answers `PRAGMA wal_checkpoint` on such a database with success (`0|-1|-1`), and compacting there is opportunistic anyway. A store whose file was replaced underneath the handle is a different state, one stock cannot enter and therefore has no opinion about, and the caller is checkpointing a path that no longer holds the database it opened. It can only learn that if the refusal reaches it. The translation is now conditioned on `!pCs->movedReadOnly`.

Worth noting for severity: the safety property was never at risk — the replacement database survived before and after; what was lost was only the report. Writes on such a store already fail with `attempt to write a readonly database`, so the checkpoint was the single operation hiding the condition.

Verification:

```
checkpoint on replaced path   before: 0|0|0      after: attempt to write a readonly database
  replacement database        untouched in both
checkpoint on read-only conn  0|0|0 (unchanged)  stock: 0|-1|-1     file untouched
VACUUM on read-only conn      refuses (unchanged)  stock: refuses
ordinary statements on a replaced-path connection  still fine, no spurious error
```

`replaced_checkpoint_reports` and `readonly_checkpoint_stays_quiet` are added to the suite so both halves stay pinned — the first fails on e0f619d41a, the commit reviewed here, and passes now. Suite is 9/9; full battery re-running.

PR #2273, opened 2026-08-18
Do not treat a conflicted cherry-pick as a merge

timsehn:
A cherry-pick or revert that hits a cell conflict reused catalog merge, which forced `isMerging=1` so `dolt_conflicts` would show rows. That made a pick look like a real merge:

```sql
BEGIN;
SELECT dolt_cherry_pick('feat');          -- cell conflict
SELECT is_merging FROM dolt_merge_status; -- was 1
SELECT dolt_checkout('other');            -- unresolved merge conflicts
SELECT dolt_merge('--abort');             -- wiped the pick
```

`dolt_commit` already treats cherry-pick as a single-parent commit (empty merge source). Only the merge flag was wrong.

Conflicts are now recorded without setting `isMerging`. In-transaction `dolt_conflicts` reads the conflicts catalog even when that flag is clear. A real `dolt_merge` still sets `isMerging` itself. Checkout still refuses unresolved cherry-pick conflicts because DoltLite never persists a conflicted working set; the error is no longer a merge abort.

After this change: `is_merging=0`, `dolt_conflicts` still has the row, `dolt_merge('--abort')` reports no merge in progress, and checkout says to resolve or roll back.

### DIFF
--- a/src/doltlite_checkout.c
+++ b/src/doltlite_checkout.c
@@ -1183,6 +1183,11 @@ void doltCheckoutFunc(sqlite3_context *ctx, int argc, sqlite3_value **argv){
       doltliteVcResultError(ctx, db, "unresolved merge conflicts \xe2\x80\x94 commit or abort first");
       return;
     }
+    if( doltliteSessionHasUnresolvedConflicts(db) ){
+      doltliteVcResultError(ctx, db,
+        "unresolved conflicts \xe2\x80\x94 resolve them or roll back first");
+      return;
+    }
   }
 
   if( strcmp(zBranch, "-b")==0 ){
--- a/src/doltlite_cmd.c
+++ b/src/doltlite_cmd.c
@@ -74,6 +74,13 @@ void doltliteCmdResultPeerBranchBusy(sqlite3_context *ctx, const char *zOp){
   }
 }
 
+static int markPendingReplayIfNotMerging(sqlite3 *db){
+  u8 isMerging = 0;
+  doltliteGetSessionMergeState(db, &isMerging, 0, 0);
+  if( isMerging ) return SQLITE_OK;
+  return doltliteSetSessionPendingReplayCommit(db, 1);
+}
+
 int doltliteReportConflicts(
   sqlite3 *db,
   sqlite3_context *ctx,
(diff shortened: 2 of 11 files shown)


timsehn:
On the additional finding — **Reopening loses an active merge conflict** — this is documented behaviour rather than a defect, so there is nothing to fix here.

`AGENTS.md:240`:

> **Conflicts are never persisted.** A conflicted merge lives only in the transaction that produced it: `COMMIT` is refused while any conflict remains, an autocommit merge that conflicts is rolled back whole, and the persist path declines to make a conflicted working set durable. Nothing conflicted reaches disk, **so no later connection inherits a merge to finish.** Dolt allows this behind `@@dolt_allow_commit_conflicts`, which DoltLite does not implement, so conflict *persistence* is a deliberate divergence — compare in-transaction behaviour against Dolt, not what survives a commit.

`README.md` states it twice more, in the merge section and in the invariants list: *"Nothing conflicted is left on disk for a later connection."*

The reproduction ends the writer session with the conflict unresolved and then reopens. `is_merging=0` and zero conflicts is precisely what the invariant promises: the conflicted working set was never made durable, so there is no state to lose. The stated impact — that a user cannot continue the merge after a restart — is the intended design, because there is no merge to continue. Every one of the 13 in-transaction cases in this run, which is where the contract says to compare against Dolt, passed.

The code analysis is internally consistent but reasons from the wrong premise. It reads the difference between `dolt_conflicts` (durable refresh) and `dolt_merge_status` (no refresh) as an asymmetric read contract and suggests giving merge-status the same durable refresh. That would refresh from a working-set blob that by contract never records a conflicted merge, so it would either change nothing or begin surfacing a merge that is not supposed to exist. `src/prolly_btree_state.c:1149` already carries a comment pointing at this invariant.

One distinction worth keeping straight, since it is easy to conflate: merge state *is* durable once a merge has been resolved and committed — a fresh process can pick that up and finish it. It is only the *unresolved* conflicted state that never reaches disk. The reproduction exercises the second case.

Marking this as no action needed. The PR-attributable coverage looks right to me, and the change itself — recording conflicts without setting `isMerging` — is consistent with the same invariant.

PR #2276, opened 2026-08-19
Let SELECT FROM dolt_ignore work on a fresh database

timsehn:
Dolt answers `SELECT * FROM dolt_ignore` on a brand-new repo (zero rows). DoltLite raised `no such table` until the user issued `CREATE TABLE`.

```sql
SELECT count(*) FROM dolt_ignore;   -- was error; now 0
INSERT INTO dolt_ignore VALUES ('tmp_*', 1);
SELECT * FROM dolt_ignore;          -- tmp_*|1
```

`dolt_ignore` now uses the same lazy eponymous vtab as `dolt_docs` and `dolt_tests`: empty reads before a backing table exists, first write materializes

`dolt_ignore(pattern TEXT NOT NULL, ignored TINYINT NOT NULL, PRIMARY KEY(pattern))`.

That is Dolt's lazy-creation UX without the open-time auto-materialization that previously went wrong. `dolt_add` still treats a missing table as no rules, and still honors patterns after the first insert. Hand-issued `CREATE TABLE` still has to match the shape guard.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -116,22 +116,25 @@ conformance bugs. When comparing against Dolt, only **row-level semantics** must
 match; do not file conformance issues over the vtable/function shape or column
 naming.
 
-`dolt_docs` is a hybrid: an eponymous writable vtab answers reads while no
-backing table exists (always-empty scans, so `SELECT` never errors on a fresh
-repo), and the first write statement materializes a real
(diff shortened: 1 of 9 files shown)


timsehn:
Automated fix: shifted attachmalloc known-divergences after dolt_ignore module registration moved OOM fault points by +2.

PR #2281, opened 2026-08-19
Resolve open time with large wals

### LINKED ISSUE #2266: DoltLite database open time grows with incremental commits
Using DoltLite v0.11.50, repeatedly opening a database, inserting one fixed-size row, and committing causes database-open latency to grow substantially. Initial openings are ~3ms, with later openings taking ~3-4s and dominating the operation cost.

Reproduction:

```
  /tmp/doltlite tmp_db < doltlite_gc_open_time_repro.sql >/dev/null
  /usr/bin/time -f '%e' /tmp/doltlite tmp_db "SELECT 1;" >/dev/null

  for iteration in $(seq 1 200); do
    /tmp/doltlite tmp_db < doltlite_gc_open_time_repro.sql >/dev/null
  done

  for iteration in $(seq 1 5); do
    /usr/bin/time -f '%e' /tmp/doltlite tmp_db "SELECT 1;" >/dev/null
  done

  for iteration in $(seq 1 200); do
    /tmp/doltlite tmp_db < doltlite_gc_open_time_repro.sql >/dev/null
  done

  for iteration in $(seq 1 5); do
    /usr/bin/time -f '%e' /tmp/doltlite tmp_db "SELECT 1;" >/dev/null
  done
```

This is a small script that shows an increase.

There's an additional issue here as well I noticed -- the storage overhead is quite severe. We're inserting a 2097152 byte payload for each row, which should land us at ~838 MB of space. The actual db size is ~1.2GB giving a storage overhead of ~1.5x.

[doltlite_gc_open_time_repro.sql](https://github.com/user-attachments/files/31188910/doltlite_gc_open_time_repro.sql)

Chris7:
## Summary

Fix database open time growing linearly with accumulated WAL data.

Opening a database previously replayed and rehashed every chunk in the WAL. Large rows therefore made every subsequent open scan the entire database history.

The fix:

- Writes a paged, append-only WAL index checkpoint after each 64 MiB of new WAL data.
- Publishes checkpoints at commit time, while the writer still holds the file lock, so long-lived writers and crash/restart workloads do not depend on `sqlite3_close`.
- Records the checkpoint location in sealed root manifests.
- Finds the newest valid stamped root even when a torn, zero-filled, or drained suffix follows it.
- Loads the checkpoint lazily and replays only the bounded WAL tail.
- If a checkpoint body is invalid, rebuilds from the real WAL records while skipping only the derived checkpoint batch.
- Keeps the normal clean-close marker as a fallback when checkpoint creation fails or has nothing to write.
- Preserves eager verification for raw, ref-less chunk stores.
- Remains compatible with version 12 databases and older readers; no storage-format bump is required.
- Allows GC to discard checkpoint pages normally.

Existing databases perform one full replay after upgrading. Their first eligible commit or clean close creates the checkpoint used by subsequent opens.

## Why full WAL replay is no longer required

Previously, the WAL was the only source for rebuilding the in-memory chunk-hash-to-file-offset index. Every open therefore had to parse every WAL record and hash every chunk body.

The checkpoint persists the result of that replay:

```
Before: base index + replay WAL[0..N] -> current index
After:  load checkpoint at K + replay WAL[K..N] -> current index
```

The checkpoint contains the complete sorted chunk index as of position K. Its sealed root records the current refs hash and the exact offset where replay should resume.

Skipping the WAL prefix remains safe because:

- Checkpoint pages are content-hashed, linked by a Merkle tree, and verified as they are loaded.
- The root manifest is sealed and validates the checkpoint location, size, replay offset, and batch boundary.
- Checkpoint pages are synced before the root becomes durable.
- Later commit roots carry the checkpoint pointer forward.
- A torn suffix does not hide an earlier valid checkpoint root.
- A missing or untrusted checkpoint falls back to ordinary replay; a trusted stamp with a corrupt derived body skips that checkpoint batch and rebuilds from the original WAL records.
- Historical chunk contents remain content-verified when read and by `PRAGMA integrity_check`.

The WAL prefix is not discarded. Its already-computed state is loaded from a verified snapshot, and only records written after that snapshot need to be replayed.

## Arbitrarily large histories

The checkpoint is a tree of bounded 4 KiB pages rather than one flat record. Open reads only the root and the pages needed by later lookups, so checkpoint loading does not require one allocation proportional to the complete history and is not limited by a single record's 32-bit size.

Each new checkpoint resets the replay boundary. With the default threshold, ordinary open-time replay stays bounded to approximately the latest 64 MiB of post-checkpoint WAL, even as commit history continues growing.

## Storage impact

Checkpoints add derived index data; they do not duplicate chunk bodies or discard history. A checkpoint costs roughly 32 bytes per indexed unique chunk, plus less than one byte per chunk for page-tree structure, one 169-byte root, record headers, and at most one sector of alignment padding.

A new full snapshot is appended after each additional 64 MiB of post-checkpoint WAL. Because the file is append-only, older checkpoint snapshots remain as unreachable bytes until GC. Without GC, cumulative checkpoint space is the sum of all snapshot sizes and can grow quadratically in the number of checkpoint intervals when the unique-chunk count grows linearly. `dolt_gc()` rewrites only graph-reachable chunks, drops all checkpoint pages and their stamp, and reopens from the compacted index; later WAL growth creates a fresh checkpoint.

## Performance

Using the supplied repro with a new 2 MiB row per commit:

| Commits | Before | After |
| ---: | ---: | ---: |
| 201 | 0.12-0.13s | 0.00s |
| 401 | 0.24-0.25s | 0.00-0.01s |

The final database retained all 401 rows and 402 commits and passed `PRAGMA integrity_check`.

The repeated-history timing regression doubles history while measuring 40 fresh opens per stage:

| Commits | Median wall time for 40 opens |
| ---: | ---: |
| 64 | 1197 ms |
| 128 | 1207 ms |
| 256 | 1309 ms |

Observed growth was 112 ms, below the 750 ms guard. The small differences include process/filesystem timing and the bounded post-checkpoint tail; they do not accumulate with the full WAL history.

The separate checkpoint-creation harness compares five sets of 2 MiB WAL appends around the default 64 MiB boundary. It syncs each cloned starting file, uses an untimed query to force the database open, then times all three appends on the same connection. It verifies the first remains below the threshold, the second publishes a checkpoint, and the third reuses the same checkpoint offset and replay boundary. The integrated timing-suite run measured:

| Append | Median |
| --- | ---: |
| Below threshold | 36.813 ms |
| Creates checkpoint | 42.985 ms |
| First append after checkpoint, same connection | 25.997 ms |

The checkpoint/below ratio was **1.168x**, while the post-checkpoint/below ratio was **0.706x**. Database-open and replay time is outside all three timers. The creation cost is concentrated on the threshold-crossing append, and the next append does not retain that overhead. The harness prints every sample and both ratios but intentionally does not enforce a bound yet, so CI data can inform a reasonable value for X.

## Testing

- Added fail-before recovery coverage for commit-time publication, crash-before-close reopen, lazy fast-path use, prefix damage, corrupt-checkpoint fallback, a torn suffix, GC stamp removal, and continued writes after fallback.
- Replaced the 65 MiB corruption fixture with a 4 KiB payload and a runtime-configured 1 KiB threshold through the same production checkpoint path.
- Added a checkpoint-creation performance harness separate from the open-time history test.
- Retargeted the existing `misc7` intentional IOERR divergence from `12.64.3` to `12.71.3`; the exception count and ratchet remain unchanged.
- Exact coverage-built `sqlite-regression-core-sql`: 338824 passing assertions, 873 known divergences, no unexpected failures.
- Timing suites: 8/8 passed.
- Corruption suite: 224/224 passed.
- Version-control suites: 108/108 passed.
- C test binaries: 29/29 passed, including concurrency, multi-process GC, and the OOM fault sweep.
- Regression assertions: 310930/310930 passed.
- Storage-format contract: 23/23 passed.
- Stock SQLite parity: 119/119 passed.
- `make lint` and the testfixture exception ratchet passed.

Fixes #2266.

### DIFF
--- a/src/chunk_store.c
+++ b/src/chunk_store.c
@@ -799,10 +799,12 @@ int chunkStoreHasMany(ChunkStore *cs, const ProllyHash *aHash, int nHash, u8 *aR
 int chunkStoreHas(ChunkStore *cs, const ProllyHash *hash, int *pHas){
   int idx = -1;
   int rc;
+  ChunkIndexEntry entry;
   *pHas = 0;
   if( cs->notADatabase ) return SQLITE_NOTADB;
-  if( csSearchIndex(cs->index.aIndex, cs->index.nIndex, hash) >= 0 ){
-    *pHas = 1;
+  rc = csIndexLookup(cs, hash, &entry, pHas);
+  if( rc!=SQLITE_OK ) return rc;
+  if( *pHas ){
     return SQLITE_OK;
   }
   rc = csSearchRecent(cs, hash, &idx);
@@ -861,16 +863,19 @@ int chunkStoreGet(
 
   {
     ChunkIndexEntry *e;
+    ChunkIndexEntry indexEntry;
+    int found = 0;
     rc = csSearchRecent(cs, hash, &idx);
     if( rc!=SQLITE_OK ) return rc;
     if( idx >= 0 ){
       e = &cs->staging.aRecent[idx];
     }else{
-      idx = csSearchIndex(cs->index.aIndex, cs->index.nIndex, hash);
-      if( idx < 0 ){
+      rc = csIndexLookup(cs, hash, &indexEntry, &found);
+      if( rc!=SQLITE_OK ) return rc;
+      if( !found ){
         return SQLITE_NOTFOUND;
       }
-      e = &cs->index.aIndex[idx];
+      e = &indexEntry;
     }
 
     if( cs->file.pFile == 0 ){
@@ -1305,6 +1310,8 @@ int chunkStoreCopyIntoEmpty(ChunkStore *pSrc, ChunkStore *pDest){
   int rc;
 
   assert( chunkStoreIsEmpty(pDest) );
+  rc = csMaterializeIndex(pSrc);
+  if( rc!=SQLITE_OK ) return rc;
   chunkIndexGetEntries(&pSrc->index, &nEntry, &aEntry);
   rc = csCopyEntries(pSrc, pDest, aEntry, nEntry);
   if( rc!=SQLITE_OK ) return rc;
--- a/src/chunk_store.c
+++ b/src/chunk_store.c
@@ -530,13 +530,12 @@ int chunkStoreOpen(
       return rc;
     }
 
-    rc = csReadIndex(cs);
-    if( rc != SQLITE_OK ){
-      chunkStoreClose(cs);
-      return rc;
+    {
+      int loadedCheckpoint = 0;
+      rc = csTryLoadWalCheckpoint(cs, &loadedCheckpoint);
+      if( rc==SQLITE_OK && !loadedCheckpoint ) rc = csReadIndex(cs);
+      if( rc==SQLITE_OK && !loadedCheckpoint ) rc = csReplayWal(cs);
     }
-
-    rc = csReplayWal(cs);
     if( rc != SQLITE_OK ){
       chunkStoreClose(cs);
       return rc;
@@ -664,7 +663,7 @@ static void csWriteCleanCloseMarker(ChunkStore *cs){
   if( cs->isMemory || cs->isBuffer || cs->readOnly || cs->corruptMidStream ){
     return;
   }
-  if( !cs->file.pFile || !cs->file.zFilename || cs->wal.cleanCloseMarker ){
+  if( !cs->file.pFile || !cs->file.zFilename ){
     return;
   }
   if( cs->wal.iWalOffset<=0 || cs->wal.nWalData<=0 ){
@@ -676,6 +675,7 @@ static void csWriteCleanCloseMarker(ChunkStore *cs){
   if( prollyHashCompare(&cs->refs.refsHash, &cs->refs.committedRefsHash)!=0 ){
     return;
   }
+  if( cs->wal.cleanCloseMarker && !csWalCheckpointDue(cs) ) return;
 
   lockHeld = cs->lockDepth>0;
   if( !lockHeld ){
@@ -697,6 +697,15 @@ static void csWriteCleanCloseMarker(ChunkStore *cs){
     if( sectorSize > 65536 ) sectorSize = 65536;
   }
 
+  if( csWalCheckpointDue(cs) ){
+    rc = csWriteWalCheckpoint(cs, sectorSize);
+    if( rc!=SQLITE_OK ){
+      sqlite3_log(SQLITE_NOTICE,
+        "doltlite: unable to checkpoint chunk WAL on close: %d", rc);
+    }
+    goto done;
+  }
+
   markerStart = cs->file.iFileSize;
   if( markerStart <= 0 ) goto done;
   markerEnd = markerStart + (i64)sizeof(rootRec);
@@ -708,6 +717,7 @@ static void csWriteCleanCloseMarker(ChunkStore *cs){
 
   rootRec[0] = CS_WAL_TAG_ROOT;
   csSerializeManifest(cs, rootRec + 1);
+  csStampWalCheckpoint(cs, rootRec + 1);
   CS_WRITE_I64(rootRec + 1 + CS_MANIFEST_DURABLE_TO_OFF, markerStart);
   CS_WRITE_I64(rootRec + 1 + CS_MANIFEST_NEXT_OFF_OFF, markerNext);
   CS_WRITE_I64(rootRec + 1 + CS_MANIFEST_BATCH_START_OFF, markerStart);
--- a/src/chunk_store.c
+++ b/src/chunk_store.c
@@ -532,9 +532,14 @@ int chunkStoreOpen(
 
     {
       int loadedCheckpoint = 0;
-      rc = csTryLoadWalCheckpoint(cs, &loadedCheckpoint);
+      i64 iSkipStart = 0;
+      i64 iSkipEnd = 0;
+      rc = csTryLoadWalCheckpoint(cs, &loadedCheckpoint,
+                                  &iSkipStart, &iSkipEnd);
       if( rc==SQLITE_OK && !loadedCheckpoint ) rc = csReadIndex(cs);
-      if( rc==SQLITE_OK && !loadedCheckpoint ) rc = csReplayWal(cs);
+      if( rc==SQLITE_OK && !loadedCheckpoint ){
+        rc = csReplayWalSkipping(cs, iSkipStart, iSkipEnd);
+      }
     }
     if( rc != SQLITE_OK ){
       chunkStoreClose(cs);
@@ -698,12 +703,13 @@ static void csWriteCleanCloseMarker(ChunkStore *cs){
   }
 
   if( csWalCheckpointDue(cs) ){
-    rc = csWriteWalCheckpoint(cs, sectorSize);
+    int wroteCheckpoint = 0;
+    rc = csWriteWalCheckpoint(cs, sectorSize, &wroteCheckpoint);
     if( rc!=SQLITE_OK ){
       sqlite3_log(SQLITE_NOTICE,
         "doltlite: unable to checkpoint chunk WAL on close: %d", rc);
     }
-    goto done;
+    if( wroteCheckpoint ) goto done;
   }
 
   markerStart = cs->file.iFileSize;
--- a/src/chunk_wal.c
+++ b/src/chunk_wal.c
@@ -132,13 +132,11 @@ void csStampWalCheckpoint(const ChunkStore *cs, u8 *aManifest){
 
 static i64 csWalCheckpointThreshold(void){
   i64 n = 64*1024*1024;
-#if defined(SQLITE_TEST) || defined(DOLTLITE_MECH_REPRO)
   const char *z = getenv("DOLTLITE_WAL_CHECKPOINT_THRESHOLD");
   if( z && z[0] ){
     i64 v = (i64)strtoll(z, 0, 10);
     if( v>0 ) n = v;
   }
-#endif
   return n;
 }
 
--- a/src/chunk_wal.c
+++ b/src/chunk_wal.c
@@ -25,12 +25,245 @@ void walStateSetOffset(WalState *w, i64 iOffset){
   assert( w!=0 );
   assert( iOffset>=0 );
   w->iWalOffset = iOffset;
+  w->iCheckpointOffset = 0;
+  w->nCheckpointIndex = 0;
+  w->iCheckpointReplay = 0;
 }
 
 void walStateSetDataSize(WalState *w, i64 nData){
   assert( w!=0 );
   assert( nData>=0 );
   w->nWalData = nData;
+}
+
+static int csReadCheckpointStamp(
+  const u8 *aManifest,
+  i64 iRootOffset,
+  WalState *pWal
+){
+  i64 iOffset;
+  i64 nIndex;
+  i64 iReplay;
+  i64 iCheckpointRoot;
+
+  if( CS_READ_U32(aManifest + CS_MANIFEST_CHECKPOINT_MAGIC_OFF)
+      != CS_WAL_CHECKPOINT_MAGIC ){
+    return 0;
+  }
+  iOffset = CS_READ_I64(aManifest + CS_MANIFEST_CHECKPOINT_OFFSET_OFF);
+  nIndex = (i64)CS_READ_U32(
+      aManifest + CS_MANIFEST_CHECKPOINT_SIZE_OFF);
+  iReplay = CS_READ_I64(
+      aManifest + CS_MANIFEST_CHECKPOINT_REPLAY_OFF);
+  if( iOffset<pWal->iWalOffset || nIndex<=0
+   || nIndex%CHUNK_INDEX_ENTRY_SIZE!=0
+   || nIndex>INT_MAX
+   || iOffset>LARGEST_INT64-CS_WAL_CHUNK_HDR_SIZE-nIndex ){
+    return 0;
+  }
+  iCheckpointRoot = iOffset + CS_WAL_CHUNK_HDR_SIZE + nIndex;
+  if( iCheckpointRoot>iRootOffset
+   || iCheckpointRoot>LARGEST_INT64-(1+CHUNK_MANIFEST_SIZE)
+   || iReplay<iCheckpointRoot+1+CHUNK_MANIFEST_SIZE
+   || iReplay-(iCheckpointRoot+1+CHUNK_MANIFEST_SIZE)>=65536 ){
+    return 0;
+  }
+  pWal->iCheckpointOffset = iOffset;
+  pWal->nCheckpointIndex = nIndex;
+  pWal->iCheckpointReplay = iReplay;
+  return 1;
+}
+
+void csStampWalCheckpoint(const ChunkStore *cs, u8 *aManifest){
+  if( cs->wal.iCheckpointOffset<=0 || cs->wal.nCheckpointIndex<=0
+   || cs->wal.iCheckpointReplay<=0 ){
+    return;
+  }
+  assert( cs->wal.nCheckpointIndex<=0xffffffffu );
+  CS_WRITE_U32(aManifest + CS_MANIFEST_CHECKPOINT_MAGIC_OFF,
+               CS_WAL_CHECKPOINT_MAGIC);
+  CS_WRITE_I64(aManifest + CS_MANIFEST_CHECKPOINT_OFFSET_OFF,
+               cs->wal.iCheckpointOffset);
+  CS_WRITE_U32(aManifest + CS_MANIFEST_CHECKPOINT_SIZE_OFF,
+               (u32)cs->wal.nCheckpointIndex);
+  CS_WRITE_I64(aManifest + CS_MANIFEST_CHECKPOINT_REPLAY_OFF,
+               cs->wal.iCheckpointReplay);
+}
+
+static i64 csWalCheckpointThreshold(void){
+  i64 n = 64*1024*1024;
+#if defined(SQLITE_TEST) || defined(DOLTLITE_MECH_REPRO)
+  const char *z = getenv("DOLTLITE_WAL_CHECKPOINT_THRESHOLD");
+  if( z && z[0] ){
+    i64 v = (i64)strtoll(z, 0, 10);
+    if( v>0 ) n = v;
+  }
+#endif
+  return n;
+}
+
+int csWalCheckpointDue(const ChunkStore *cs){
+  i64 iBase = cs->wal.iCheckpointReplay>0
+            ? cs->wal.iCheckpointReplay : cs->wal.iWalOffset;
+  /* Ref-less raw stores retain eager verification of every WAL chunk. */
+  return !prollyHashIsEmpty(&cs->refs.refsHash)
+      && iBase>0 && cs->file.iFileSize>=iBase
+      && cs->file.iFileSize-iBase>=csWalCheckpointThreshold();
+}
+
+static void csSerializeCheckpointEntry(u8 *aBuf, const ChunkIndexEntry *pEntry){
+  memcpy(aBuf, pEntry->hash.data, PROLLY_HASH_SIZE);
+  CS_WRITE_I64(aBuf + PROLLY_HASH_SIZE, pEntry->offset);
+  CS_WRITE_U32(aBuf + PROLLY_HASH_SIZE + 8, (u32)pEntry->size);
+}
+
+static int csDeserializeCheckpointEntry(
+  const u8 *aBuf,
+  ChunkIndexEntry *pEntry
+){
+  u32 n = CS_READ_U32(aBuf + PROLLY_HASH_SIZE + 8);
+  if( n>0x7fffffffu ) return SQLITE_CORRUPT;
+  memcpy(pEntry->hash.data, aBuf, PROLLY_HASH_SIZE);
+  pEntry->offset = CS_READ_I64(aBuf + PROLLY_HASH_SIZE);
+  pEntry->size = (int)n;
+  return SQLITE_OK;
+}
+
+static int csBuildCheckpointIndex(
+  ChunkStore *cs,
+  ChunkIndexEntry **ppIndex,
+  int *pnIndex
+){
+  i64 nAlloc = (i64)cs->index.nIndex + cs->staging.nRecent;
+  ChunkIndexEntry *aIndex;
+  int i;
+  int nOut = 0;
+
+  *ppIndex = 0;
+  *pnIndex = 0;
+  if( nAlloc<=0 || nAlloc>INT_MAX/(int)sizeof(ChunkIndexEntry) ){
+    return nAlloc<=0 ? SQLITE_OK : SQLITE_TOOBIG;
+  }
+  aIndex = sqlite3_malloc64(
+      (sqlite3_uint64)nAlloc * sizeof(ChunkIndexEntry));
+  if( !aIndex ) return SQLITE_NOMEM;
+  if( cs->index.nIndex>0 ){
+    memcpy(aIndex, cs->index.aIndex,
+           cs->index.nIndex * sizeof(ChunkIndexEntry));
+  }
+  if( cs->staging.nRecent>0 ){
+    memcpy(aIndex + cs->index.nIndex, cs->staging.aRecent,
+           cs->staging.nRecent * sizeof(ChunkIndexEntry));
+  }
+  qsort(aIndex, (size_t)nAlloc, sizeof(ChunkIndexEntry), csIndexEntryCmp);
+  for(i=0; i<(int)nAlloc; i++){
+    if( nOut==0
+     || prollyHashCompare(&aIndex[nOut-1].hash, &aIndex[i].hash)!=0 ){
+      aIndex[nOut++] = aIndex[i];
+    }
+  }
+  *ppIndex = aIndex;
+  *pnIndex = nOut;
+  return SQLITE_OK;
+}
+
+int csWriteWalCheckpoint(ChunkStore *cs, int sectorSize){
+  ChunkIndexEntry *aIndex = 0;
+  int nIndex = 0;
+  i64 nIndexBytes;
+  i64 iCheckpoint;
+  i64 iRoot;
+  i64 iRootEnd;
+  i64 iNext;
+  u8 aEntry[CHUNK_INDEX_ENTRY_SIZE];
+  u8 aBuf[65536];
+  u8 aHeader[CS_WAL_CHUNK_HDR_SIZE];
+  u8 aRoot[1 + CHUNK_MANIFEST_SIZE];
+  ProllyHash snapshotHash;
+  blake3_hasher hasher;
+  int i;
+  int rc;
+
+  rc = csBuildCheckpointIndex(cs, &aIndex, &nIndex);
+  if( rc!=SQLITE_OK || nIndex==0 ) goto checkpoint_done;
+  nIndexBytes = (i64)nIndex * CHUNK_INDEX_ENTRY_SIZE;
+  if( nIndexBytes>INT_MAX
+   || cs->file.iFileSize>LARGEST_INT64-CS_WAL_CHUNK_HDR_SIZE-nIndexBytes ){
+    rc = SQLITE_TOOBIG;
+    goto checkpoint_done;
+  }
+
+  blake3_hasher_init(&hasher);
+  for(i=0; i<nIndex; i++){
+    csSerializeCheckpointEntry(aEntry, &aIndex[i]);
+    blake3_hasher_update(&hasher, aEntry, sizeof(aEntry));
+  }
+  blake3_hasher_finalize(&hasher, snapshotHash.data, PROLLY_HASH_SIZE);
+
+  iCheckpoint = cs->file.iFileSize;
+  csFillChunkHdr(aHeader, &snapshotHash, (u32)nIndexBytes);
+  rc = sqlite3OsWrite(cs->file.pFile, aHeader, sizeof(aHeader), iCheckpoint);
+  if( rc!=SQLITE_OK ) goto checkpoint_rollback;
+  for(i=0; i<nIndex; ){
+    int nEntry = nIndex-i;
+    int j;
+    if( nEntry>(int)(sizeof(aBuf)/CHUNK_INDEX_ENTRY_SIZE) ){
+      nEntry = (int)(sizeof(aBuf)/CHUNK_INDEX_ENTRY_SIZE);
+    }
+    for(j=0; j<nEntry; j++){
+      csSerializeCheckpointEntry(
+          aBuf + j*CHUNK_INDEX_ENTRY_SIZE, &aIndex[i+j]);
+    }
+    rc = sqlite3OsWrite(cs->file.pFile, aBuf,
+        nEntry*CHUNK_INDEX_ENTRY_SIZE,
+        iCheckpoint + CS_WAL_CHUNK_HDR_SIZE
+        + (i64)i * CHUNK_INDEX_ENTRY_SIZE);
+    if( rc!=SQLITE_OK ) goto checkpoint_rollback;
+    i += nEntry;
+  }
+  rc = csSyncFile(cs);
+  if( rc!=SQLITE_OK ) goto checkpoint_rollback;
+
+  iRoot = iCheckpoint + CS_WAL_CHUNK_HDR_SIZE + nIndexBytes;
+  iRootEnd = iRoot + (i64)sizeof(aRoot);
+  iNext = iRootEnd;
+  if( sectorSize>1 ){
+    iNext += sectorSize - 1;
+    iNext -= iNext % sectorSize;
+  }
+  aRoot[0] = CS_WAL_TAG_ROOT;
+  csSerializeManifest(cs, aRoot + 1);
+  CS_WRITE_I64(aRoot + 1 + CS_MANIFEST_DURABLE_TO_OFF, iRoot);
+  CS_WRITE_I64(aRoot + 1 + CS_MANIFEST_NEXT_OFF_OFF, iNext);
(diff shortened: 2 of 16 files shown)


src/chunk_wal.c
```diff
@@ -534,4 +776,165 @@ int csReplayWal(ChunkStore *cs){
...
+  cs->wal.iCheckpointReplay = stamp.iCheckpointReplay;
+  aIndex = 0;
+  rc = csReplayWalFrom(cs, stamp.iCheckpointReplay);
+  if( rc!=SQLITE_OK ) return rc;
+  *pLoaded = 1;
+  return SQLITE_OK;
+
+checkpoint_invalid:
+  sqlite3_free(aIndex);
+  return rc==SQLITE_NOMEM ? rc : SQLITE_OK;
```
timsehn, line 1325: **[bug]** A corrupt checkpoint body does not fall back to a usable full replay. `csTryLoadWalCheckpoint` detects the hash mismatch and returns `SQLITE_OK` with `*pLoaded==0`, then open runs `csReplayWal`. The snapshot is a sealed `CS_WAL_TAG_CHUNK` whose own root sets `DURABLE_TO` to `iRoot` (after the body, `csWriteWalCheckpoint` line 236), so replay classifies the bad hash as mid-stream damage and sets `corruptMidStream`. GC is allowed to drop this chunk because it is not graph-reachable, yet a single flipped bit in it bricks reads and commits. Stamp/root damage does fall back (seal fails, no load, prefix replay succeeds); body damage does not. The new test encodes the poison as expected (`test/corruption_test.c:1209`), which contradicts the PR claim that a corrupt checkpoint is ignored.

**Suggestion:** On verification failure, remember `stamp.iCheckpointOffset` and have full replay skip (or treat as torn) that one WAL record so the prefix of real chunks can rebuild the index. Also set the checkpoint batch's `DURABLE_TO`/`BATCH_START` to `iCheckpoint` rather than `iRoot`, matching the manifest comment that `DURABLE_TO` is the valid prefix when the batch began. Update the corrupt-snapshot test to expect a readable DB, not `corruptMidStream`.

src/chunk_store.c
```diff
@@ -697,6 +697,15 @@ static void csWriteCleanCloseMarker(ChunkStore *cs){
     if( sectorSize > 65536 ) sectorSize = 65536;
   }
 
+  if( csWalCheckpointDue(cs) ){
```
timsehn, line 705: **[suggestion]** New checkpoints are written only from `csWriteCleanCloseMarker`, never from `csCommitToFile`. A long-lived writer that crosses 64 MiB without closing leaves every other opener on the linear full-replay path; a crash before the first clean close does the same. Issue 2266's process-per-commit repro closes each time, so it is covered, but servers and crash-restart are not. Commits only propagate an already-written stamp via `csStampWalCheckpoint`.

**Suggestion:** When `csWalCheckpointDue` is true at commit time, write the snapshot as its own synced batch (or as the leading records of the commit batch) so the bound does not depend on `sqlite3_close`. Keep the close path as a backstop for the last 64 MiB after the final commit.

src/chunk_wal.c
```diff
@@ -534,4 +776,165 @@ int csReplayWal(ChunkStore *cs){
...
+  int nIndex;
+  int iEntry = 0;
+  int rc;
+
+  *pLoaded = 0;
+  if( cs->wal.iWalOffset<=0 || !cs->file.pFile ) return SQLITE_OK;
+  rc = sqlite3OsFileSize(cs->file.pFile, &nFile);
+  if( rc!=SQLITE_OK ) return rc;
+  if( nFile<(i64)sizeof(aTail) ) return SQLITE_OK;
+  iTail = nFile - (i64)sizeof(aTail);
```
timsehn: **[suggestion]** Fast-path load requires the physical file to end exactly on a sealed root (`iTail = nFile - sizeof(aTail)`). A torn commit tail, `SIZE_HINT` zero-fill, or `csDrainPendingToWal` chunks after the last root all make the tail check fail, so a perfectly good earlier stamp is ignored and open full-replays (and will poison if the snapshot body is the damaged record — see the related bug on the invalid-checkpoint return). Other paths already know how to find a tail root (`csReadDiskRefsHash`, WAL damage scan).

**Suggestion:** If the exact tail is not a sealed stamped root, scan backward the same way as `csReadDiskRefsHash` / `csWalResolveDamage` for the last valid stamped root and load that, then replay only the suffix. Keep "no sealed root" as the full-replay fallback.

test/corruption_test.c
```diff
@@ -1108,6 +1108,114 @@ static void test_damage_far_before_sealing_root_poisons(void){
...
+        (off_t)checkpointOff+CS_WAL_CHUNK_HDR_SIZE, &bad, 1)==0);
+    bad ^= 0x01;
+    check("open_checkpoint_corrupt_snapshot",
+      corrupt_bytes(dbpath,
+        (off_t)checkpointOff+CS_WAL_CHUNK_HDR_SIZE, &bad, 1)==0);
+    rc = chunkStoreOpen(&cs, sqlite3_vfs_find(0), dbpath,
+        SQLITE_OPEN_READWRITE | SQLITE_OPEN_MAIN_DB);
+    check("open_checkpoint_corrupt_reopen", rc==SQLITE_OK);
+    if( rc==SQLITE_OK ){
+      check("open_checkpoint_corrupt_poisoned", cs.corruptMidStream);
```
timsehn: **[suggestion]** The new test checks that a checkpoint is written, that a later commit copies the stamp, and that a corrupt snapshot poisons open. It does not prove the fast path was taken (reopen would still work via full replay), does not corrupt a prefix WAL chunk and show open still succeeds, does not cover GC dropping the snapshot then reopening from the compacted index, and does not cover a crash tail still using an earlier stamp. It also inserts a 65 MiB blob instead of setting `DOLTLITE_WAL_CHECKPOINT_THRESHOLD` (compiled into `csWalCheckpointThreshold` for `SQLITE_TEST` builds) and using a tiny payload.

**Suggestion:** Add a prefix-damage case that must open cleanly if the snapshot is used; a `dolt_gc` case that clears the stamp (`walStateSetOffset`) and still reads rows; a torn-tail case if backward stamp discovery is added. Drive size with the existing test env var so this is not a 65 MiB file in the corruption suite. Flip the corrupt-snapshot expectation if the poison-on-fallback bug is fixed.

src/chunk_store.c
```diff
@@ -697,6 +697,15 @@ static void csWriteCleanCloseMarker(ChunkStore *cs){
...
+    rc = csWriteWalCheckpoint(cs, sectorSize);
+    if( rc!=SQLITE_OK ){
+      sqlite3_log(SQLITE_NOTICE,
+        "doltlite: unable to checkpoint chunk WAL on close: %d", rc);
+    }
+    goto done;
```
timsehn: **[suggestion]** If a checkpoint is due, the close path always `goto done`, including when `csWriteWalCheckpoint` fails or no-ops (`nIndex==0` at `chunk_wal.c:188`). Previously a due close still wrote a clean-close marker. Failure now skips both the snapshot and the marker, so the next open always full-replays even though a marker would have been cheap and correct.

**Suggestion:** On checkpoint failure (or `nIndex==0`), fall through to the existing marker write instead of `goto done`. Only skip the marker when a checkpoint root was actually synced.

timsehn:
Restarted failed CI jobs that looked transient (assert-enabled, libfuzzer, asan-ubsan-ubuntu, coverage, compatibility-references, development-All-O0 — apt/mirror hang during Install dependencies, then job timeout). No code changes.

timsehn:
Restarted failed CI jobs that looked transient (assert-enabled, asan-ubsan-ubuntu, coverage — all cancelled mid-apt on azure.archive.ubuntu.com). No code changes.

src/chunk_store.c
Chris7: Fixed in 6b82007f3d. csWriteWalCheckpoint now reports whether a checkpoint root was actually synced. The close path skips the normal marker only in that case; allocation or I/O failure and the empty-index no-op both fall through to the clean-close marker. The complete OOM sweep passed with no surfaced bugs.

test/corruption_test.c
Chris7: Added the requested coverage in 6b82007f3d: direct proof that the lazy fast path is active, prefix damage that only the checkpoint can bypass, GC clearing the stamp followed by a successful reopen, backward discovery behind a torn tail, corrupt-checkpoint fallback that remains readable and writable, and crash-before-close recovery. The 65 MiB blob is now 4 KiB, with the SQLITE_TEST regression using a 1 KiB threshold. The focused corruption suite is 224/224.

src/chunk_wal.c
Chris7: Fixed in 6b82007f3d. Open now scans backward from physical EOF in bounded windows for the newest offset-sealed, manifest-valid stamped root, then replays only its suffix. The regression appends a torn record after a valid checkpoint, damages a prefix WAL hash, and asserts that the lazy checkpoint fast path still opens and reads successfully.

src/chunk_store.c
Chris7: Fixed in 6b82007f3d. csCommitToFile now publishes a due checkpoint after the commit is durable and while the file lock is still held. Checkpoint allocation or I/O failure remains a benign optimization failure and does not roll back the successful user commit. The new low-threshold regression copies a live writer before close and verifies that the crash copy reopens from the commit-time checkpoint.

src/chunk_wal.c
Chris7: Fixed in 6b82007f3d. CKP2 roots now set DURABLE_TO and BATCH_START to the start of the complete checkpoint-page batch. If a trusted stamp points to an invalid checkpoint body, open skips only that derived batch and replays the original WAL records before and after it without propagating the bad stamp. The regression now requires the database to remain readable and writable instead of becoming poisoned.

timsehn:
```First real failure you can send Chris:

Job: Test Suite / sqlite-regression-ported
https://github.com/dolthub/doltlite/actions/runs/32290461497/job/96199922076

Both doltlite_recover_shared_ioerr points now fail (1.2 was green on the previous SHA):

1.1 expected: [1 {disk I/O error} 1 1 {0 {1024 1 1024}} {0 {1023 1 512}} {0 ok}]
1.1 got:      [1 {disk I/O error} 0 1 {0 {1024 1 1024}} {0 {1024 1 1024}} {0 ok}]

1.2 expected: [1 {disk I/O error} 1 1 {0 {1024 1 1024}} {0 {1023 1 512}} {0 ok}]
1.2 got:      [1 {disk I/O error} 0 1 {0 {1024 1 1024}} {0 {1023 1 512}} {0 ok}]

::drs_residx is 0 instead of 1 on both — the injected IOERR hits during the DELETE, not the later INSERT/COMMIT (sqlite_io_error_pending 122 and 126). On 1.1 the writer also still has the full 1024-row table instead of 1023/512 after the partial DELETE. integrity_check is still ok.

Not a flake.
```

timsehn:
Restarted failed CI jobs that looked transient (Test Suite / oracle-tests-diff-history-data, Test Suite / oracle-tests-merge-replay-schema, Test Suite / sqlite-regression-ddl-schema, Test Suite / sqlite-regression-core-sql, Sanitizers / asan-ubsan-oracles, Sanitizers / asan-ubsan-native, Sanitizers / tsan — all apt-get timeouts against azure.archive.ubuntu.com). No code changes.

timsehn:
Restarted failed CI jobs that looked transient (Test Suite / sqlite-regression-core-sql, Sanitizers / tsan — apt-get timeouts against azure.archive.ubuntu.com on attempt 3). No code changes.

timsehn:
CI babysit on `fc09328` (`Make WAL checkpoints crash-safe`): Ubuntu apt/runner flakes were restarted. After that, three jobs still failed and **reproduced** on retry. Those are the real ones.

Run: https://github.com/dolthub/doltlite/actions/runs/32290461497

### 1. `Test Suite / sqlite-regression-ported`
https://github.com/dolthub/doltlite/actions/runs/32290461497/job/96217959301

`doltlite_recover_shared_ioerr-1.1` and `-1.2` (`sqlite_io_error_pending` 122 and 126):

```
1.1 expected: [1 {disk I/O error} 1 1 {0 {1024 1 1024}} {0 {1023 1 512}} {0 ok}]
1.1 got:      [1 {disk I/O error} 0 1 {0 {1024 1 1024}} {0 {1024 1 1024}} {0 ok}]

1.2 expected: [1 {disk I/O error} 1 1 {0 {1024 1 1024}} {0 {1023 1 512}} {0 ok}]
1.2 got:      [1 {disk I/O error} 0 1 {0 {1024 1 1024}} {0 {1023 1 512}} {0 ok}]
```

The third field is `::drs_residx`. Expected `1` means the DELETE succeeded and the injected IOERR hit the later INSERT/COMMIT. Got `0` means the IOERR fired during the DELETE. On 1.1 the writer also still has the full 1024-row table instead of 1023/512. Peer snapshot and `integrity_check` still match.

### 2. `Test Suite / sqlite-regression-fault-fuzz`
https://github.com/dolthub/doltlite/actions/runs/32290461497/job/96217958299

37 unexpected failures. Extra checkpoint I/O moved the hard-coded fault indexes (same pattern as (1), wider window than the previous SHA):

- `ioerr-7.54.1` … `ioerr-7.58.1`
- `shared_ioerr-2.136.cleanup.1` … `shared_ioerr-2.140.cleanup.1` (known divergences already cover 122–135)
- `attachmalloc-1.transient.{506,939,1441}`
- `sysfault-2.1-vfsfault-transient.104` … `.111`
- `sysfault-2.2-vfsfault-persistent.106` … `.111`
- `sysfault-2.2-vfsfault-transient.{49,55,56,58,106–111}`

### 3. `Benchmark / relative-performance-gate`
https://github.com/dolthub/doltlite/actions/runs/32290461497/job/96217960574

`vc / vc` **1.265x** (924.62ms → 1.17s); suite gate is 1.25x. Confirmed across retries. Overall ratio 1.015x. Commit-time checkpoints are on the write path now.

---

Not real (recovered after rerun):

- Apt / Azure mirror hangs (`Ign:` then install canceled): tsan, asan-native, asan-oracles, ddl-schema, oracle diff-history and merge-replay, core-sql
- `sqlite-regression-storage` `bigfile` 600s timeout after `bigfile-1.11` — passed on retry

`78821924` (`Update checkpoint fault-injection contracts`) landed after this run and may cover (1) and (2). The perf gate was not in that commit.

timsehn:
The fault-fuzz / ported changes look like the right kind of divergence accounting: extra checkpoint I/O moved the ordinals, and you retargeted the known lists and `doltlite_recover_shared_ioerr` (122/126 → 127/129) without weakening the recovery contract. That is fine.

We still expect **Benchmark / relative-performance-gate** to fail on this SHA. Last run reproduced `vc / vc` at **1.265x** (924.62ms → 1.17s, gate 1.25x). This commit does not change the commit-time checkpoint write path, so that result should still be there.

Is there a way to make checkpoint creation cheaper (or less frequent on the VC commit path) so we stay under the gate? If not, we can consider raising it.

timsehn:
Correction on the perf gate: the PR **version-over-version** `relative-performance-gate` (`vc / vc` 1.265x vs PR base) is a one-time step. Once this is on master, later PRs compare against a baseline that already has checkpoints, so that gate should be fine.

The one that actually matters is **nightly**: DoltLite vs stock SQLite (and the absolute VC ceilings in `vc_perf_ceiling.sh`). Commit-time checkpoints stay on that path every night, so that is the regression that will not reset.

If there is a way to make checkpoint creation cheaper or less frequent on the VC commit path so nightly stays in bounds, great. If not, we can retune the nightly VC ceilings rather than block on the PR relative gate.

timsehn:
The relative `vc / vc` gate on this PR is measuring the wrong thing. `vc_perf_ceiling.sh` currently builds seeds with the **PR-base** binary, then times the candidate on copies of that file. Candidate close then writes the first WAL checkpoint (unstamped seed already ≥ 64 MiB) inside the timed window. That is an upgrade path, not steady-state `dolt_status` / `dolt_merge`. Nightly will not see it.

[#2287](https://github.com/dolthub/doltlite/pull/2287) makes each binary build its own seeds. Once that is on master, please rebase this PR onto it so the VC relative gate compares like-for-like layouts.

`test/doltlite_open_perf.sh` is the right check for the original bug (open time vs history). It does not measure checkpoint **creation**. Please add a similar harness that times writing the snapshot (threshold crossing / close or commit) as history grows, with a bound, so we have a number for the nightly VC-ceiling discussion instead of only inferring it from the PR relative gate.

timsehn:
Restarted failed CI jobs that looked transient (Smoke / vc-fuzz, Sanitizers / asan-ubsan-c-regression, Test Suite / oracle-tests-merge-replay-schema, Test Suite / sqlite-regression-fts3, Test Suite / sqlite-regression-fts4-core, Test Suite / sqlite-regression-ported — apt-get timeouts against azure.archive.ubuntu.com). No code changes.

timsehn:
CI babysit on `7882192` (`Update checkpoint fault-injection contracts`): Ubuntu apt/runner flakes were restarted. After that, one job still failed and **reproduced** on retry. That is the remaining real red.

Run: https://github.com/dolthub/doltlite/actions/runs/32299778269

### 1. `Test Suite / sqlite-regression-core-sql`
https://github.com/dolthub/doltlite/actions/runs/32299778269/job/96242918976

Unexpected failure: `misc7-12.71.3` (`do_ioerr_test misc7-12`, echo vtab over a PK-clustered `abc` table). The known list still has `misc7-12.64.3` (ioerr faultsim recovery check in that same class); that entry is now a **mismatch** (no longer hits), and the new ordinal is 12.71.3.

Same shape as the ported `doltlite_recover_shared_ioerr` retarget on this PR (122/126 → 127/129): extra checkpoint I/O moved the faultsim index. Other `misc7` failures in the testfixture dump (`5`, `7.0`, `10`, `11`, `13`, `15.2`, `22.3`, `22.4`) are already listed as intentional.

`CI Gate` is red only because of this job. No code changes from babysit.

timsehn:
CI babysit on `78821924` (`Update checkpoint fault-injection contracts`): Ubuntu apt/Azure mirror flakes were restarted once. After that, one job still failed and **reproduced** on retry. That is the real one.

Run: https://github.com/dolthub/doltlite/actions/runs/32299778269

### 1. `Test Suite / sqlite-regression-core-sql`
https://github.com/dolthub/doltlite/actions/runs/32299778269/job/96242918976

1 unexpected failure (338877 passing, 872 known divergences):

- `misc7` / `misc7-12.71.3`

This is `do_ioerr_test misc7-12` (echo vtab over PK-clustered `abc`). Known divergences still list `misc7-12.64.3` (issue 1840, same echo-on-clustered ioerr recovery check). On this SHA the runner also says that entry no longer matches and should be updated:

```
MISMATCH: misc7 — these entries no longer match and should be updated in test/known_testfixture_divergences.txt:
    misc7-12.64.3
```

Unexpected failure list: `misc7 misc7-12.71.3`. Same extra-checkpoint-I/O ordinal shift as the `doltlite_recover_shared_ioerr` 122/126 → 127/129 retarget in this commit (64 → 71 here). The truncated testfixture tail does not print 12.71.3 expected/got (it starts at 15.1). The other printed mismatches are already known intentional divergences:

```
misc7-15.2...
Error: no such column: rowid

! misc7-22.3 expected: [1 {attempt to write a readonly database}]
! misc7-22.3 got:      [0 {1 2 3 4}]

! misc7-22.4 expected: [SQLITE_READONLY_ROLLBACK]
! misc7-22.4 got:      [SQLITE_OK]
```

---

Not real (recovered after rerun):

- Apt `Ign:` then `apt-get` exit 124 during Install dependencies: vc-fuzz, asan-ubsan-c-regression, sqlite-regression-fts3, oracle-tests-merge-replay-schema, sqlite-regression-ported, sqlite-regression-fts4-core
- `sqlite-regression-ported` actually ran tests this attempt and **passed** (the `fc09328` `doltlite_recover_shared_ioerr` 1.1/1.2 failures are gone)
- `sqlite-regression-fault-fuzz` **passed**
- `Benchmark / relative-performance-gate` **passed**

timsehn:
One leftover on this SHA: `misc7-12.71.3` in `sqlite-regression-core-sql`. The known list still has `misc7-12.64.3` (issue 1840, ioerr faultsim in the echo-on-clustered class). Same ordinal slide as the 122/126 → 127/129 retarget — move the listed id to `12.71.3` (and bump the ratchet if needed).

Reminder on the checkpoint **creation** perf test (separate from `doltlite_open_perf.sh`, which only bounds open time vs history):

A journal/WAL append that stays just under the 64 MiB threshold should be no more than **X** times faster than the append that actually creates a checkpoint. I do not know a reasonable X until we see the numbers — please add a harness that measures both and prints the ratio so we can pick a bound.

Chris7:
Addressed in babe246222:

- Moved the existing misc7 IOERR divergence from 12.64.3 to 12.71.3. The classified exception count remains 4,795, so the ratchet did not need to change. The exact coverage-built core-sql bucket is green with 338,824 passing assertions, 873 known divergences, and no unexpected failures.
- Added doltlite_checkpoint_perf.sh as a separate timing harness. It measures five paired fixed-shape 2 MiB WAL updates around the production 64 MiB threshold, fsyncs each cloned starting database before timing, proves the below-threshold append does not checkpoint, and proves the crossing append publishes the checkpoint before close.
- The harness prints each sample plus the ratio of median checkpoint time to median below-threshold time. A representative unloaded run was 44.857 ms below versus 77.246 ms checkpointing, or 1.722x. It intentionally does not enforce X yet.

The full timing set passes 8/8, and the PR description now includes the harness design and observed ratio.

Chris7:
Extended the checkpoint harness in 342cb91d91 with a third timed append after checkpoint creation and reopen.

The harness now verifies that the third append carries the same checkpoint offset and replay boundary, so it is measuring reuse of the existing checkpoint rather than another checkpoint creation.

The integrated timing-suite medians were:

- below threshold: 47.049 ms
- creates checkpoint: 60.212 ms
- first append after checkpoint and reopen: 52.075 ms
- checkpoint/below: 1.280x
- post/below: 1.107x

This indicates that the extra cost is concentrated on checkpoint creation and the next append returns close to the pre-checkpoint baseline. The full timing set remains green at 8/8.

Chris7:
Correction to the post-checkpoint measurement in 0721f69a83: all three timed appends now run on one persistent DoltLite connection.

The harness first runs an untimed query to force database open and WAL replay. It then times the below-threshold, checkpoint-creating, and post-checkpoint updates back-to-back, inspecting the on-disk checkpoint stamp between statements. No database-open time is included in any of the three measurements.

The integrated timing-suite medians are now:

- below threshold: 36.813 ms
- creates checkpoint: 42.985 ms
- first append after checkpoint, same connection: 25.997 ms
- checkpoint/below: 1.168x
- post/below: 0.706x

The third append retains the same checkpoint offset and replay boundary, and its timing confirms that checkpoint overhead does not persist onto the next append. The full timing set remains green at 8/8.

timsehn:
CI babysit on `0721f69` (`Performance test and sql fix`): one real red so far.

Run: https://github.com/dolthub/doltlite/actions/runs/32325405623

### `Benchmark / relative-performance-gate`
https://github.com/dolthub/doltlite/actions/runs/32325405623/job/96301406025

Not a flake. Confirmed across 3 attempts (`vc` confirmed 2 exact gates). Gate is section/suite/overall > 1.25x with >10ms delta.

```
Overall ratio: 1.000x
Confirmed failed gates:
  - section vc / vc
  - suite vc

| Suite       | Baseline  | Candidate | Ratio  | Result |
|-------------|-----------|-----------|--------|--------|
| blobpk      | 11.29s    | 11.23s    | 0.994x | PASS   |
| compositepk | 11.50s    | 11.42s    | 0.993x | PASS   |
| int         | 10.02s    | 10.00s    | 0.998x | PASS   |
| textpk      | 11.38s    | 11.31s    | 0.994x | PASS   |
| vc          | 908.83ms  | 1.15s     | 1.268x | FAIL   |
```

Individual VC tests all stay under the 2.00x/50ms individual gate. The suite miss is the status/diff cluster (~+36–38ms, 1.42–1.48x):

```
status_clean_many_tables          80.18ms -> 116.06ms  1.448x
status_dirty_many_tables          83.15ms -> 120.39ms  1.448x
diff_regular_working_one_table    77.49ms -> 115.01ms  1.484x
diff_regular_working_many_tables  87.99ms -> 125.16ms  1.422x
diff_stat_working_many_tables     88.06ms -> 125.59ms  1.426x
diff_schema_working_many_tables   88.22ms -> 125.75ms  1.425x
```

Merges/branches are ~1.02–1.18x. Same shape as the earlier `vc / vc` 1.265x on `fc09328`. `#2287` (own seeds per binary) is already on master; this run still compared against `77bcc95`, so it is still the first-checkpoint-on-close upgrade path.

`Sanitizers / asan-ubsan-oracles` is still running. No other failures on this SHA yet. No code changes from babysit.

timsehn:
CI babysit on `0721f69a83` (`Performance test and sql fix`). Apt/runner flakes did not need a restart. `Test Suite / sqlite-regression-core-sql` is green on this SHA (`misc7-12.71.3` leftover from `78821924` is gone).

The only remaining red is the relative VC gate.

Run: https://github.com/dolthub/doltlite/actions/runs/32325405623

### 1. `Benchmark / relative-performance-gate`
https://github.com/dolthub/doltlite/actions/runs/32325405623/job/96301406025

Confirmed after 3 automatic retries (same two gates each time). #2287 is already on the PR base (`77bcc958`), and the report now says each binary builds its own seeds — so this is like-for-like, not the old upgrade-path artifact.

- **Gates:** individual > 1.50x with >10ms; section/suite/overall > 1.25x with the same min delta; **vc individual** > 2.00x with >50ms
- **Confirmed failed gates:** section `vc / vc`, suite `vc`
- **OLTP suites:** blobpk 0.994x, compositepk 0.993x, int 0.998x, textpk 0.994x — all PASS
- **vc suite:** 13 workloads, 908.83ms → 1.15s, **1.268x FAIL** (gate 1.25x)

No individual VC workload fails the 2.00x / 50ms bar. The extra cost is concentrated on status/diff (~+36–38ms, ~1.42–1.48x):

| Benchmark | PR base median | Candidate median | Ratio |
|---|---:|---:|---:|
| status_clean_many_tables | 80.18ms | 116.06ms | 1.448x |
| status_dirty_many_tables | 83.15ms | 120.39ms | 1.448x |
| diff_regular_working_one_table | 77.49ms | 115.01ms | 1.484x |
| diff_regular_working_many_tables | 87.99ms | 125.16ms | 1.422x |
| diff_stat_working_many_tables | 88.06ms | 125.59ms | 1.426x |
| diff_schema_working_many_tables | 88.22ms | 125.75ms | 1.425x |
| checkout_branch_clean | 54.26ms | 64.12ms | 1.182x |
| merge_* (4) | ~1.02–1.07x | | PASS |

`CI Gate` is red only because this job is red. Fork workflow could not post the relative-performance comment (`HttpError: Resource not accessible by integration`).

timsehn:
CI babysit on `35ad246` (`Avoid duplicate WAL scans before first checkpoint`). Apt/runner flakes did not need a restart. Run finished: two real test reds, plus `CI Gate` which is only red because of those.

Run: https://github.com/dolthub/doltlite/actions/runs/32391489045

**Version-control performance is now passing.** `Benchmark / relative-performance-gate` and `Benchmark / version-control` are green on this SHA.

https://github.com/dolthub/doltlite/actions/runs/32391489045/job/96512841310

On `0721f69` the remaining red was `vc / vc` at ~1.265–1.268x vs the 1.25x suite gate. Here:

- overall ratio **1.006x**
- vc suite **1.001x** (925.62ms → 926.74ms)
- confirmed failed gates: none

Skipping the extra WAL scan before the first checkpoint fixed the upgrade-path hit that the relative gate was measuring.

### Remaining real reds (known-list ordinal slide, not flakes)

Same pattern as `misc7-12.64.3` → `12.71.3` on the previous SHA: fewer WAL I/Os moved the faultsim points. Classified IOERR/malloc behavior is unchanged; the listed ids need to move.

#### 1. `Test Suite / sqlite-regression-core-sql`
https://github.com/dolthub/doltlite/actions/runs/32391489045/job/96508404625

338,865 passing, 872 known, 1 unexpected.

- unexpected: `misc7-12.67.3`
- listed, no longer failing: `misc7-12.71.3`

Retarget `misc7-12.71.3` → `misc7-12.67.3` in `test/known_testfixture_divergences.txt`.

#### 2. `Test Suite / sqlite-regression-fault-fuzz`
https://github.com/dolthub/doltlite/actions/runs/32391489045/job/96508404722

- unexpected: `attachmalloc-1.transient.{505,938,1440}`
- listed, no longer failing: `.506`, `.939`, `.1441`

Shift those three listed ids down by 1.

`sqlite-regression-ported` and `Sanitizers / tsan` are green on this SHA.

timsehn:
CI babysit on `c9d71cd` (`update divergence`). Apt install was fine; a job rerun is blocked until the workflow finishes.

Run: https://github.com/dolthub/doltlite/actions/runs/32397701307

The previous SHA's real reds are **green** here:

- `Test Suite / sqlite-regression-core-sql` success
- `Test Suite / sqlite-regression-fault-fuzz` success
- `Benchmark / relative-performance-gate` success

One new red so far:

### `Test Suite / sqlite-regression-storage`
https://github.com/dolthub/doltlite/actions/runs/32397701307/job/96527627080

Not an assertion mismatch. Unexpected failures: 0. Unexpected termination: `bigfile` timed out after 600s (last completed `bigfile-1.11`). The rest of the storage bucket passed (45,702 passing, 3,165 known divergences).

This is the per-file 600s kill in `run_testfixture.sh`. The workflow already calls out runner-load timeouts of this kind (e.g. `bigsort`). Storage was green on `35ad246` with the same engine (`c9d71cd` only retargets known lists), so this looks like a slow-runner timeout rather than a new hang from the WAL-scan change.

Will restart that job once `Sanitizers / asan-ubsan-oracles` finishes — GitHub returns 403 if you rerun a job while the workflow is still running.

timsehn:
Correction on the `bigfile` timeout: this is not just a noisy runner.

Same coverage job, same 17 tests:

| SHA | Checkpoints | `bigfile` wall |
|---|---|---|
| Recent master-line PRs (e.g. `end-of-options`, `merge-index-collate`) | no | **~1.0s** |
| This PR `fc093` / `0721f69` / `35ad246` | yes | **~390–392s** |
| `c9d71` (known-list only) | yes | **600s timeout** after `bigfile-1.11` |

`bigfile` punches a sparse 4 GiB / 8 GiB / 16 GiB hole, pokes the header, and reopens. Checkpoint discovery/open on that tail is ~390× slower and now sits at ~65% of the 600s per-file budget, so a slow coverage runner misses.

`c9d71` did not cause this (txt-only known-list retarget). `35ad246` still finished in 392s. A rerun may go green; it will not bring `bigfile` back to 1s.

timsehn:
Restarted failed CI jobs that looked transient (Test Suite / sqlite-regression-storage — `bigfile` timed out after 600s, last completed `bigfile-1.11`; 0 unexpected assertion failures). No code changes.

timsehn:
This failure is a real performance regression that manifests as a timeout on the bigfile test case. I'm investigating.

timsehn:
Recommended fix for the `bigfile` timeout: cap checkpoint **discovery** on a sparse/zero tail. Do not bump the 600s `run_testfixture` budget.

### What actually blew the budget

`Test Suite / sqlite-regression-storage` died with:

```
TIMEOUT (unexpected): bigfile after 600s (last completed: bigfile-1.11)
```

Same coverage job, 17 tests:

| SHA | Checkpoints | `bigfile` wall |
|---|---|---|
| Recent master-line PRs | no | **~1.0s** |
| This PR `fc093` / `0721f69` / `35ad246` | yes | **~390–392s** |
| `c9d71` (known-list only) | yes | **600s timeout** after `bigfile-1.11` |

That is not “open a 16 GiB database of real WAL is 390× slower.” `fake_big_file 16384` writes 14 bytes at offset 16 GiB (`src/test2.c`), which is a sparse hole. The real roots are still the tiny DB at the start. `1.11` is `md5sum(t2)` after that extend; `1.12` is `md5sum(t3)`.

### The scan

`csTryLoadWalCheckpoint` → `csFindLastCheckpointRoot` (`src/chunk_wal.c`) starts at **physical EOF** and walks backward in ~64 KiB windows looking for a checkpoint-stamped WAL root. There is **no cap**. It only stops at `iWalOffset`.

On `bigfile` that is ~16 GiB / 64 KiB ≈ 260k zero reads per open, several opens per file. Coverage CI turns that into ~392s, which is ~65% of the 600s per-file kill, so a slow runner misses.

`35ad246` correctly stops once it finds a clean-close root **without** a checkpoint stamp (so it does not keep walking for a stamp). It still has to walk the entire hole **to find that root**.

A real large DB with a checkpoint/root near EOF is one 64 KiB window. That is the case `doltlite_open_perf.sh` measures. The unshippable case is a large **zero tail** (sparse extend, `SIZE_HINT`, `fallocate`, crash after extend): every open becomes O(filesize).

### Recommended change

Reuse the rule you already have for damage probes:

```c
#define CS_WAL_SCAN_MAX (64*1024*1024)
/* legitimate preallocated tail never hides data past this window */
static int csWalTailIsZero(...)
```

Apply that to **discovery only**:

1. In `csFindLastCheckpointRoot`, if a `CS_WAL_SCAN_MAX` window from EOF is all zeros, stop (or jump to the last data extent). Do not walk 16 GiB of zeros.
2. Better on Linux: `SEEK_HOLE` / `SEEK_DATA` to the last data extent, then scan only that. Sparse `bigfile` becomes O(1) again.
3. Do **not** put the same cap on `csWalResolveDamage`. That path must still see a sealed root past a torn region, or you rewind committed batches.

Do not skip scanning a real WAL that actually contains records. Only refuse to treat a zero hole as WAL.

A rerun of storage may go green (392s < 600s on the last three SHAs). It will not bring `bigfile` back to 1s, and it will not fix preallocated tails in production.

PR #2283, opened 2026-08-19
Bound and retry apt calls in CI

timsehn:
## Summary

- route CI dependency installs and the best-effort gdb install through one bounded apt-get runner
- retry transient failures twice with APT network and lock timeouts
- cap each attempt at 300 seconds, with forced termination after a 10-second grace period
- add an Ubuntu self-test for production defaults, retry success, exhaustion, option forwarding, and hung-command termination

The 300-second bound is more than twice the observed 143-second healthy update while a continuously stuck apt-get command now releases its runner in roughly ten minutes instead of consuming the full job timeout.

## Validation

- retry helper self-test
- bash syntax checks
- parsed all workflow YAML
- orphaned-suite lint
- git diff --check

### DIFF
--- a/.github/scripts/install-apt-deps.sh
+++ b/.github/scripts/install-apt-deps.sh
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
 for path in /etc/apt/sources.list.d/*.list /etc/apt/sources.list.d/*.sources; do
   [ -e "$path" ] || continue
   if grep -qiE 'packages[.]microsoft[.]com|azure-cli' "$path"; then
@@ -8,14 +10,5 @@ for path in /etc/apt/sources.list.d/*.list /etc/apt/sources.list.d/*.sources; do
   fi
 done
 
-for attempt in 1 2 3; do
-  if sudo apt-get update; then
-    break
-  fi
-  if [ "$attempt" -eq 3 ]; then
-    exit 1
-  fi
-  sleep $((attempt * 5))
-done
-
-sudo apt-get install -y "$@"
+bash "$script_dir/retry-apt-get.sh" update
+bash "$script_dir/retry-apt-get.sh" install -y "$@"
--- a/.github/scripts/retry-apt-get-test.sh
+++ b/.github/scripts/retry-apt-get-test.sh
@@ -1,0 +1,87 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+runner="$script_dir/retry-apt-get.sh"
+tmp="$(mktemp -d)"
+trap 'rm -rf "$tmp"' EXIT
+mkdir "$tmp/bin"
+has_timeout=1
+if ! command -v timeout >/dev/null 2>&1; then
+  has_timeout=0
+  cat >"$tmp/bin/timeout" <<'EOF'
+#!/usr/bin/env bash
+while [[ "$1" == --* ]]; do
+  shift
+done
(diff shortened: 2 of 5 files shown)


timsehn:
Review — the shape is right and I want this, but **one blocking problem: the 90s cap is below what a healthy `apt-get update` already takes in this repo's CI**, so this would turn passing jobs into failing ones.

Bounding each attempt is exactly the fix, and `sudo timeout … apt-get` rather than `timeout sudo …` is the correct ordering, so the signal reaches apt instead of sudo. The self-test is a nice touch, especially the hung-command case asserting rc 124.

## The bound is too low

From a **successful** CI run (`32206041534`), job `Test Suite / sqlite-regression-fts4-core`:

```
2026-08-19T02:00:44Z  Fetched 10.7 MB in 2min 23s (74.9 kB/s)
```

That is `apt-get update` alone taking **143 seconds** and succeeding. Under this PR it is TERMed at 90s, retried twice, each attempt killed the same way, and the job fails after ~5 minutes — having previously passed.

Not a one-off: in that same successful run, **6 of 58 install steps took over 90 seconds** (193s, 119s, 105s, 105s, 102s, 96s). Those totals cover both the `update` and `install` calls, so not every one necessarily has a single call over the cap — but the 193s case demonstrably does. The transfer rate in that log, 74.9 kB/s from the Azure mirror, is the same slowness that produces the hangs, so slow-but-working updates are normal here rather than exceptional.

## Suggested change

Raise `APT_RETRY_TIMEOUT_SECONDS` to around **300** — roughly 2x the slowest successful update I measured — and consider dropping to **2 attempts** so the worst case stays near 10 minutes. That still releases a genuinely wedged runner well inside the 20-minute job cap, and inside the **15-minute** cap that `coverage-remotes` and `libfuzzer` use, which a 3×300s worst case would exceed.

## For reference, the signature this is fixing

The hang I traced across yesterday's runs is `apt-get update` printing its `Ign:`/`Get:` lines and then emitting **nothing at all** until the job cap — six log lines total over 25 minutes, no divergence, no error, and an orphaned `doltlite`/`bash` process at cleanup. A per-attempt bound is the right answer to that; it just needs to sit above the legitimate slow path. When one of those jobs was restarted it completed in 13.5 minutes, which is how I concluded it was a wedged runner rather than slow work.

Everything else looks good. Happy to re-review once the bound is retuned.

timsehn:
Addressed the review in `70a6f7c958`:

- raised the per-attempt bound from 90s to 300s
- reduced the outer attempts from three to two
- changed the self-test to exercise and assert the production defaults
- rebased onto the latest `master`

The PR description now reflects the observed 143s healthy update and the roughly ten-minute worst case.

timsehn:
Fresh data from this morning that strengthens the case for this PR — and sharpens the constraint on the bound.

On #2282, run `32272614945`, the hang hit **7 jobs on the first attempt** and **1 more on the rerun** (`crash`, twice in a row). Every one cancelled in `Install dependencies` at its job cap having compiled nothing. So the flake is worth fixing and this is the right mechanism.

The wrinkle for tuning: **all three of the jobs it killed first — `crash`, `optimized-benchmark`, `assert-enabled` — use the 15-minute cap, not 20.** That is the binding constraint. With 3 attempts the retry budget has to fit inside 15 minutes end to end, which at a 300s timeout it would not (3×300s + backoff ≈ 16 min, and the runner still needs time to do the actual work afterwards).

Concretely, two combinations that both fit and clear the 143s slow path I measured:

- **2 attempts × 300s** ≈ 10 min worst case, leaving ~5 min of the 15-minute jobs for the build itself.
- **3 attempts × 200s** ≈ 10.5 min worst case, if you would rather keep three tries.

Either releases a wedged runner far sooner than today's behaviour of burning the whole job, while staying above a legitimately slow update. I would not go below ~200s per attempt given the 143s measurement.

PR #2284, opened 2026-08-19
Match Dolt ignore pattern semantics

timsehn:
## Summary

- make `dolt_ignore` wildcard matching case-sensitive, matching Dolt
- replace literal-count precedence with Dolt's pattern-containment resolution
- report conflicts for opposing incomparable patterns
- make `?` matching and wildcard backtracking operate on UTF-8 characters
- add native and Dolt-oracle regressions for all three review findings

## Testing

- `make -j4 doltlite`
- `make -j4 doltlite-lib`
- `make lint`
- `DOLTLITE=build/doltlite bash test/doltlite_ignore.sh` (16 passed)
- `bash test/vc_oracle_ignore_test.sh build/doltlite dolt` (46 passed)
- `bash test/run_doltlite_tests.sh build/doltlite` (106 suites passed; run on the first commit)
- `bash test/doltlite_parity.sh build/doltlite` (119 passed on the first commit)

### DIFF
--- a/src/doltlite_ignore.c
+++ b/src/doltlite_ignore.c
@@ -6,11 +6,18 @@
 #include "doltlite_ignore.h"
 #include <string.h>
 
-static unsigned char ignoreLower(unsigned char c){
-  return (c>='A' && c<='Z') ? c + 32 : c;
-}
+typedef struct IgnoreMatch IgnoreMatch;
+struct IgnoreMatch {
+  char *zPattern;
+  u8 ignored;
+  u8 removed;
+};
 
-static int ignorePatternMatch(const char *zPat, const char *zStr){
+static int ignorePatternMatchMode(
+  const char *zPat,
+  const char *zStr,
+  int patternMode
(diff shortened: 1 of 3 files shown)


timsehn:
@itoqa Thanks for flagging this. I reproduced the two-connection case on this commit. In autocommit mode, after the peer sets the matching rule to `ignored=1`, the existing reader refreshes and `race_tmp` disappears from `dolt_status`.

An explicit read transaction behaves differently by design: it continues to see the pre-update `ignored=0` snapshot (and therefore `race_tmp`) until `COMMIT`; the next statement then sees `ignored=1` and hides the table. That is DoltLite's documented snapshot-isolation behavior.

This PR only changes pattern matching and precedence; it does not change snapshot refresh or concurrency paths. If your failing case remained stale in autocommit mode after the writer completed, please share the connection/transaction sequence and we'll capture it as a dedicated concurrency follow-up.

PR #2287, opened 2026-08-19
Give each VC perf binary its own seed databases

timsehn:
## Summary

Paired version-control benchmarks (`test/vc_perf_ceiling.sh`) built seed databases with the **baseline** binary, then timed both baseline and candidate on copies of that same image.

That measures an upgrade path, not steady-state VC. If the candidate writes extra on-disk state on close (for example a WAL index checkpoint when the unstamped seed is already ≥ 64 MiB), that write happens inside the timed CLI invocation (`open + SQL + close`). Nightly does not have this artifact: it prepares and measures with one binary.

This change makes each binary build its own seeds. Timed runs still `cp` those seeds; setup stays untimed. Nightly (no `VC_PERF_BASELINE`) still prepares once.

Related: #2281's relative `vc / vc` 1.265x gate is this upgrade-path cost, not `dolt_status` / `dolt_merge`.

### DIFF
--- a/test/vc_perf_ceiling.sh
+++ b/test/vc_perf_ceiling.sh
@@ -14,7 +14,6 @@ if [ -n "$VC_PERF_BASELINE" ] && [ ! -x "$VC_PERF_BASELINE" ]; then
   echo "baseline doltlite binary not found: $VC_PERF_BASELINE" >&2
   exit 1
 fi
-FIXTURE_DOLTLITE="${VC_PERF_BASELINE:-$DOLTLITE}"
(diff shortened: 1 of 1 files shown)


timsehn:
Restarted failed CI jobs that looked transient (asan-ubsan-ubuntu, crash — Install dependencies). No code changes.

timsehn:
Restarted failed CI jobs that looked transient (Build-Test / ubuntu — `commit_cost_flat_vs_depth` ratio 6.71 > 3.00 on the growth-axis gate). Unrelated to this PR's VC fixture change; runner-timing flake. No code changes.

PR #2315, opened 2026-08-20
Ratchet schema-merge gaps that close by refusing

timsehn:
Follow-up to #2312. Test-only. Two holes in the schema-merge matrix harness that let a listed exception rot, or let a failed merge command enter the success path.

## Stale gap when both refuse

#2312 ratchets every other class: a listed pair that stops matching its class fails until the list is updated. `MERGE_WHERE_DOLT_REFUSES` returned `PASS: both refuse` before that check, so if we start refusing a listed gap it stays on the list forever.

A listed pair that now both refuse fails with `listed as a gap but we no longer merge -- delete the entry`.

## Nonzero merge exit treated as success

`dolt_merge` / `CALL dolt_merge` stdout was classified with `dl_ok`/`dt_ok` defaulting to 1. `$?` was ignored, so a crash or an unrecognized diagnostic could walk into schema comparison as a successful merge. Capture the exit status and force the matching ok flag off when it is nonzero. Existing stdout parsing still applies.

## Testing

- `vc_oracle_correct_schema_merge_matrix_test.sh`: 167 passed, 0 failed, 6 gaps, 3 corrupting (same standing as master; ratchet did not fire on current lists)

No `src/` changes.

Co-Authored-By: Grok <noreply@x.ai>

### DIFF
--- a/test/vc_oracle_correct_schema_merge_matrix_test.sh
+++ b/test/vc_oracle_correct_schema_merge_matrix_test.sh
@@ -222,10 +222,14 @@ EOF
   fi
 
   dl_out=$("$DOLTLITE" "$db" "SELECT coalesce(dolt_merge('feat'),'NULL');" 2>&1)
+  dl_rc=$?
   # dolt_merge's own result table has a column called "conflicts", so the
   # outcome has to be read from that column's value, never from the text.
   dt_out=$(cd "$repo" && "$DOLT" sql -r csv -q "CALL dolt_merge('feat');" 2>&1)
+  dt_rc=$?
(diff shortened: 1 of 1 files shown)


timsehn:
Restarted failed CI jobs that looked transient (`Sanitizers / asan-ubsan-oracles` — Version control oracle tests timed out after 25 minutes with 0 test failures; last completed `vc_oracle_ignore_test.sh` 46 passed). Matrix ratchet itself was green (`167 passed, 0 failed`). `CI Gate` is red only because of that. No code changes.

Issue #2331, opened 2026-08-21
Dual column renames lose the index or trigger on the renamed column; refused instead of merged


timsehn:
Each branch renames a different column of one table, and an index or trigger on that table names the column its own branch renamed. Dolt merges this and renames the object along with the table. We cannot produce that result today, so as of this change we **refuse** rather than report the database as malformed.

This replaces the corruption in #2301 with an actionable error. #2301 can close; the parity gap is this issue.

## Repro

```sql
CREATE TABLE t(k INTEGER PRIMARY KEY, a TEXT, b TEXT);
CREATE INDEX ix0 ON t(a);
INSERT INTO t VALUES(1,'a1','b1');
SELECT dolt_commit('-Am','base');
SELECT dolt_branch('feat');
ALTER TABLE t RENAME COLUMN a TO a2;     -- main renames the covered column
SELECT dolt_commit('-Am','ours');
SELECT dolt_checkout('feat');
ALTER TABLE t RENAME COLUMN b TO b2;     -- feat renames another column
SELECT dolt_commit('-Am','theirs');
SELECT dolt_checkout('main');
SELECT dolt_merge('feat');
```

Before: `database disk image is malformed` (trigger form: `SQL logic error`).
Now: `cannot merge: index 'ix0' covers column 'a' of table 't', which one branch renamed while the other renamed another column of the same table; ...`

Dolt: `conflicts=0`, merge successful, columns `k,a2,b2`, index retargeted to `a2`.

## Why it happens

The merged catalog takes the table from whichever side `trySchemaColumnMerge` chooses, but the schema row for a dependent object is chosen independently — `mergedSchemaChoice` picks whichever side *changed* it. When those are different sides, the object names a column the adopted table does not have. The catalog then cannot load, and the `ALTER TABLE ... RENAME COLUMN` action that would retarget the object only runs **after** it loads:

```
adopted table:  CREATE TABLE t(k INTEGER PRIMARY KEY, a TEXT, b2 TEXT)
kept index row: CREATE INDEX ix0 ON t(a2)      <-- a2 is not in that table
=> malformed database schema (ix0) - no such column: a2
```

Only this combination breaks. If the object names the column the *other* branch renamed, it arrives beside the table it belongs to and the rename retargets both — that case merges correctly today and is covered by a test.

## What a real fix needs

Adopt one side's table **and** its dependent objects coherently, so the pending rename actions retarget the table and everything on it together. Two attempts that do **not** work, for whoever picks this up:

[...]

PR #2332, opened 2026-08-21
Refuse a dual rename that would lose an index or trigger

### LINKED ISSUE #2301: Merging two column renames fails when the table has an index, view, or trigger over the renamed column
Found by ito on #2298, reproduced by hand against a clean control build of master.

Each branch renames a different column, and the table has a dependent object (index, view, or trigger) that references the column *one* side renamed. The merged catalog then takes one side's **table** but the other side's **dependent-object SQL**, so the two disagree about the column's name and the merge cannot complete.

## Repro

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, a TEXT, b TEXT);
CREATE INDEX ix ON t(a);          -- or: CREATE VIEW v AS SELECT a FROM t;
                                  -- or: CREATE TRIGGER tg AFTER UPDATE ON t BEGIN UPDATE t SET a=a; END;
INSERT INTO t VALUES(1,'a1','b1');
SELECT dolt_add('-A'), dolt_commit('-m','base');
SELECT dolt_branch('feat');
ALTER TABLE t RENAME COLUMN a TO a2;      -- main (also rewrites its own index/view/trigger to a2)
SELECT dolt_add('-A'), dolt_commit('-m','main');
SELECT dolt_checkout('feat');
ALTER TABLE t RENAME COLUMN b TO b2;      -- feat (its index/view/trigger still say a)
SELECT dolt_add('-A'), dolt_commit('-m','feat');
SELECT dolt_checkout('main');
SELECT dolt_merge('feat');
```

Observed, per dependent object (clean control build of master `77bcc958a9` vs #2298 head, same machine):

| dependent on the renamed column | master | with #2298 |
|---|---|---|
| none | merges, but schema is `id,a,b2,a2` (the phantom column of #2290) | merges, `id,a2,b2` ✅ |
| `CREATE INDEX ix ON t(a)` | **ERROR**, schema stays `id,a2,b` | **ERROR**, same |
| `CREATE VIEW v AS SELECT a FROM t` | merges, phantom `id,a,b2,a2` | **ERROR**, schema stays `id,a2,b` |
| `CREATE TRIGGER … a …` | merges, phantom `id,a,b2,a2` | **ERROR**, schema stays `id,a2,b` |
| `CREATE INDEX ix ON t(b)` (other column) | merges, phantom | merges, `id,a2,b2` ✅ |

The error is the usual misleading `database disk image is malformed`, and `integrity_check` afterwards says `ok`, so the half-applied result looks healthy while the other branch's rename is simply missing.

**Dolt 2.2.2 merges all of these**, giving `id,a2,b2`.

## Cause

The merged catalog mixes sources. The table schema is adopted from one side; the index/view/trigger rows for that table come from the other. Whichever way the mix falls, one of them names a column the other does not have:

- the index case fails on master too, because the adopted table and the carried index disagree directly;
- the view/trigger cases fail only once a rename is carried as an `ALTER TABLE RENAME COLUMN` action (#2290's fix), because the ALTER reparses dependents, and the dependent SQL in the merged catalog was written against the *post*-rename name while the adopted table still has the *pre*-rename name.

`PRAGMA legacy_alter_table=1` does not help — SQLite rejects `RENAME COLUMN` with a dangling dependent regardless (`SQL logic error`).

## Direction

Express a rename as an **action on the non-renaming side's layout**, rather than by adopting the renaming side's schema. `ALTER TABLE RENAME COLUMN` rewrites the table and every dependent index, view and trigger in one native step, so if the merged catalog consistently holds the adopted (pre-rename) side's objects, one ALTER makes all of them correct together.

That single rule also covers the index-follows-rename gap left open by #2289 — `ours: CREATE INDEX ix ON t(b)` × `theirs: RENAME COLUMN b TO b2` currently fails the same way — so the two are one fix. It needs the catalog merge to stop choosing a table's dependent rows independently of the side whose layout was adopted, which is the substantive part of the work.

## Why #2298 is not gated for this

I tried gating the rename action on "the table has no dependent object that differs across sides", so these shapes would keep master's behavior. It is both too blunt and incomplete, so I dropped it:

- **too blunt**: an index on the *other* renamed column (`ix ON t(b)` while `a` is the carried rename) trips the gate and gives up a case #2298 otherwise fixes correctly;
- **incomplete**: a view's `tbl_name` in `sqlite_master` is the **view's own name**, not the table it reads, so a view over `t` is not discoverable from the catalog at all and the gate never fires for the case that motivated it.

So with #2298 the view and trigger shapes above error where master silently produced the phantom column — a visible failure replacing silent corruption, with the real fix tracked here.

timsehn:
Closes #2301. Parity gap tracked in #2331.

Each branch renames a different column of one table, and an index or trigger on it names the column its own branch renamed. That merge reported `database disk image is malformed` (or, for a trigger, `SQL logic error`). It now refuses with something the user can act on.

```
cannot merge: index 'ix0' covers column 'a' of table 't', which one branch
renamed while the other renamed another column of the same table; rename on
one branch only, or drop the index, then merge
```

### Why it happened

The merged catalog takes the table from whichever side `trySchemaColumnMerge` chooses, but a dependent object's schema row is chosen independently — `mergedSchemaChoice` takes whichever side *changed* it. When those are different sides the object names a column the adopted table does not have, so the catalog cannot load — and the `ALTER TABLE ... RENAME COLUMN` action that would retarget the object only runs *after* it loads:

```
adopted table:  CREATE TABLE t(k INTEGER PRIMARY KEY, a TEXT, b2 TEXT)
kept index row: CREATE INDEX ix0 ON t(a2)      <-- a2 is not in that table
=> malformed database schema (ix0) - no such column: a2
```

### The part worth reviewing

Whether the object and its table come from different sides is decided by **calling `trySchemaColumnMerge`** — the same function that makes the real decision — not by re-deriving it. My first two versions inferred it, and the second **over-refused two merges that are correct today**: the choice turns on which column *position* each side renamed, which is not something to guess at. Both of those shapes are now pinned as tests.

### DIVERGENCE FROM DOLT

Dolt merges all of these cleanly (`conflicts=0`, `k,a2,b2`, index retargeted). That is the result we should produce; #2331 has the diagnosis and the two approaches that do **not** work, so the next attempt does not repeat them. Until then a refusal beats reporting a sound database as corrupt.

### Not covered

The **view** form still reports `SQL logic error` — a view's `tbl_name` is the view itself, so nothing links it to the table. Unchanged from master, not a regression, documented in #2331. I left it rather than widen the check on guesswork, having already over-refused twice here.

### Testing

- `doltlite_schema_merge.sh` — 291 pass (6 new). The 2 refusal tests **fail on a pre-fix binary and pass with the fix**; the 3 control tests (mirror shape, single-sided rename) pass on both, which is what caught the over-refusal.
- `run_doltlite_tests.sh` — 110/110 suites
- `vc_oracle_correct_schema_merge_matrix_test.sh` — 176 passed, 0 gaps, 0 differing, 0 corrupting (unchanged)
- amalgamation regenerated and compiled clean; `lint_layers.sh` clean

Note the matrix does **not** cover this shape: its base fixture has no index, so `ren_a x ren_b` is only ever exercised on an unindexed table. #2331 records that gap — varying the base fixture is the highest-value extension to that suite.

### DIFF
--- a/src/doltlite_merge_int.h
+++ b/src/doltlite_merge_int.h
@@ -96,6 +96,7 @@ int mergePass1CheckIndexOverRenamedColumn(MergePass1Ctx *c);
 int mergePass1CheckTriggerOverRenamedTable(MergePass1Ctx *c);
 int mergePass1CheckRowEditOfDroppedColumn(MergePass1Ctx *c);
 int mergePass1CheckDuplicateIndexColumns(MergePass1Ctx *c);
+int mergePass1CheckDependentOverDualRename(MergePass1Ctx *c);
 
 int mergeRowEditsColumn(
   sqlite3 *db,
--- a/src/doltlite_merge_pass1.c
+++ b/src/doltlite_merge_pass1.c
@@ -1279,6 +1279,7 @@ int mergeCatalogPass1(
   if( rc==SQLITE_OK ) rc = mergePass1CheckTriggerOverRenamedTable(&c);
   if( rc==SQLITE_OK ) rc = mergePass1CheckRowEditOfDroppedColumn(&c);
   if( rc==SQLITE_OK ) rc = mergePass1CheckDuplicateIndexColumns(&c);
+  if( rc==SQLITE_OK ) rc = mergePass1CheckDependentOverDualRename(&c);
   if( rc!=SQLITE_OK ){
     mergePass1Free(&c);
     return rc;
--- a/src/doltlite_merge_predetect.c
+++ b/src/doltlite_merge_predetect.c
@@ -235,11 +235,12 @@ int mergePass1CheckRowEditOfDroppedColumn(MergePass1Ctx *c){
   return SQLITE_OK;
 }
 
-/* Does this object's definition name the given column? The scan is by bare
-** identifier, and for an index it starts at the column list so the index's own
-** name cannot match. It over-matches by design: an expression index over the
-** column, or a view naming it in any clause, breaks the same way a plain
-** reference does. */
+/* Does this object's definition name the given column? The scan is by
+** identifier, quoted or bare -- a definition SQLite wrote may quote the name,
+** and comparing only bare tokens let a quoted one through. For an index the
+** scan starts at the column list so the index's own name cannot match. It
+** over-matches by design: an expression index over the column, or a view
+** naming it in any clause, breaks the same way a plain reference does. */
 static int mergeSqlNamesColumn(const char *zSql, const char *zType,
                                const char *zColumn){
   const char *z;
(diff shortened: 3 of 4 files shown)


timsehn:
Rebased onto master (kept #2330's `dolt_pull` test alongside the new block) and fixed ito's finding.

**ito was right, and it was a real bypass.** My identifier scan compared bare tokens and *stepped over* quoted ones, so an index written `ON t("a2")` — a form SQLite itself produces — never matched and the refusal never fired. The merge then failed as `SQL logic error`, which is exactly the outcome this refusal exists to replace.

Reproduced before fixing:

```
our index sql: CREATE INDEX ix0 ON t("a2")
merge:         Error: SQL logic error        <-- refusal bypassed
```

Now compares the contents of quoted identifiers in all of SQLite's forms. Verified each one refuses, with the controls still merging:

| dependent | result |
|---|---|
| `ON t("a")` / `` ON t(`a`) `` / `ON t([a])` | refuses |
| `ON t(a)` / `ON t(a,b)` / `ON t(abs(a))` | refuses |
| trigger with `SET "a"="a"` | refuses |
| index on the column the *other* branch renamed | merges |
| mirror branch order | merges |
| single-sided rename | merges |

Worth noting `[a]` matched by accident before: the brackets were skipped as punctuation and the name inside read as bare. Only the `"` and `` ` `` forms were actually escaping.

Added 3 regression tests for the quoted forms — 5 of the refusal tests now fail on a pre-fix binary and pass with it.

After rebase: 297 schema-merge tests, 110/110 VC suites, matrix 176 passed / 0 gaps / 0 differing / 0 corrupting, lint clean, amalgamation compiles.

PR #2341, opened 2026-08-21
Compare views and triggers in the merge matrix

timsehn:
Fixes ito's finding on #2337.

The oracle compared the table, its columns, its index set and its row data — and **nothing about views or triggers**. A merge that dropped a dependent or left one pointing at the wrong table scored equal as long as the visible table data matched. That gap was in code I added: #2337 introduced `bview` and `btrig` base variants, so those cases were checking the table *around* the dependent rather than the dependent itself.

Now both state functions emit an inventory of dependents by type, name and target table. The definitions themselves are not comparable across engines — a trigger body is written in each engine's dialect — but a dependent that went missing or moved shows up.

### The part worth reading

My first cut reported **5 divergences**, and they were false. Dolt cannot always read its own trigger catalog back: after `ALTER TABLE t RENAME COLUMN a TO a2`, a trigger whose body names `NEW.a` makes `information_schema.triggers` fail outright with `table not found: new`. My query swallowed that with `2>/dev/null` and recorded "no dependents", so DoltLite keeping the trigger looked like a divergence.

Verified directly before believing it:

```
before rename: trigger:tg0->t
after rename:  error ... : table not found: new
```

So an unreadable catalog is now reported as unreadable and the case is **not compared**. The suite cannot compare what an engine will not tell it, and reading a query failure as an empty result is how an oracle manufactures findings.

Result: **261 passed, 0 failed, 0 gaps, 0 differing, 0 corrupting, 47 not comparable** — the 42 pre-existing skips plus these 5.

### DIFF
--- a/test/vc_oracle_correct_schema_merge_matrix_test.sh
+++ b/test/vc_oracle_correct_schema_merge_matrix_test.sh
@@ -164,7 +164,8 @@ dl_state() {
   dep=$("$DOLTLITE" "$db" "SELECT group_concat(x, ' ') FROM (
       SELECT type || ':' || name || '->' || coalesce(tbl_name,'') AS x
       FROM sqlite_master WHERE type IN ('view','trigger') ORDER BY x);" 2>&1)
-  case "$dep" in *rror*) dep=UNREADABLE;; esac
+  case "$(printf '%s' "$dep" | tr 'A-Z' 'a-z')" in
+    *error*|*"table not found"*|*malformed*) dep=UNREADABLE;; esac
   echo "dep=$dep"
   for c in $(echo "$cols" | tr ',' ' '); do
     out=$("$DOLTLITE" "$db" \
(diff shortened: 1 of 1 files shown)


timsehn:
Fixed — ito is right, and this is the failure mode the guard exists to stop.

`*error*|*Error*` misses `ERROR`, so a failed catalog read could still be scored as a catalog with nothing in it. Now the text is lowercased before matching, and the two error texts actually seen from these queries (`table not found`, `malformed`) are matched by name too.

Matrix unchanged: **261 passed, 0 failed, 0 gaps, 0 differing, 0 corrupting, 47 not comparable** — the same 5 cases skip, so the guard still fires where it should and hasn't started swallowing comparable ones.

Same fix applied to the queued follow-up that switches this inventory to `SHOW TRIGGERS`, so it cannot reintroduce the narrow match.

Issue #2354, opened 2026-08-21
Merge fails with SQL logic error when both sides add the same column and one drops the last column


timsehn:
Sibling of #2349, found while verifying its fix. Same bare `SQL logic error`, but a shape where **Dolt merges cleanly**, so it cannot be answered with a refusal — the merge has to work.

## Repro

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, payload TEXT, c1113 TEXT, c1899 TEXT, c2000 TEXT);
INSERT INTO t VALUES(1,'p','x','y','z');
SELECT dolt_commit('-A','-m','base');
SELECT dolt_branch('b');
ALTER TABLE t DROP COLUMN c1113;          -- ours drops a middle column
ALTER TABLE t ADD COLUMN c1699 TEXT;
SELECT dolt_commit('-A','-m','main');
-- on b:
ALTER TABLE t DROP COLUMN c2000;          -- theirs drops the last column
ALTER TABLE t ADD COLUMN c1699 TEXT;
SELECT dolt_commit('-A','-m','b');
-- back on main:
SELECT dolt_merge('b');   -->  SQL logic error
```

Dolt merges this without conflict.

## Why it is not #2349

Both sides drop **the same number** of columns (1 and 1), which is the case Dolt merges — #2349 and its fix cover the unequal-count case, which Dolt conflicts on. So this is not reachable by the refusal and needs the merge itself to succeed.

Position is what separates it from the neighbouring case that already works:

| ours drops | theirs drops | doltlite | dolt |
| --- | --- | --- | --- |
| `c1899` (middle) | `c1113` (middle) | merges | merges |
| `c1113` (middle) | `c2000` (**last**) | **SQL logic error** | merges |

Dropping the *last* column on one side while the other drops a middle one, with both adding the same new name, is what tips it over.

## Pre-existing

Reproduces on master (`992c599ecc`) and is unaffected by #2350 — that PR neither causes nor fixes it.

## Expected

Merge cleanly to match Dolt. Failing that, a refusal naming the table would at least be actionable, though it would be a divergence from Dolt and would need recording as one in the matrix.

Issue #2349, opened 2026-08-21
Merge fails with a bare SQL logic error when both sides add the same column name and one side drops another


timsehn:
Found by the stateful VC fuzzer in CI on an unrelated PR (#2346, test/CI files only — no `src/` change, so the engine was master).

```
FAIL: stateful vc fuzzer seed=1787347777 step=2398 op=merge
merge_b144_into_main failed rc=1
SELECT dolt_merge('b144');  ->  Error near line 1: SQL logic error
```

Fully deterministic: reproduces at the **same step 2398** on Linux CI and on macOS.

```
DOLTLITE_VC_STATEFUL_SEED=1787347777 DOLTLITE_VC_STATEFUL_SECONDS=2400 \
  python3 test/vc_stateful_fuzzer.py build/doltlite
```

## The shape

Exactly one object in the fuzzer's database differed on both sides:

```
aux_1378
  base:  (id, payload, n, c1113, c1899)
  ours:  (id, payload, n, c1113, c1899, c1699)   -- ADD c1699
  theirs:(id, payload, n, c1113, c1699)          -- DROP c1899, ADD c1699
```

Both sides add **the same new column name**, and one side also drops a *different* column.

## Minimal repro

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, payload TEXT, c1113 TEXT, c1899 TEXT);
INSERT INTO t VALUES(1,'p','x','y');
SELECT dolt_commit('-A','-m','base');
SELECT dolt_branch('b');
ALTER TABLE t ADD COLUMN c1699 TEXT;                 -- ours
SELECT dolt_commit('-A','-m','main');
-- on b:
ALTER TABLE t DROP COLUMN c1899;                     -- theirs
ALTER TABLE t ADD COLUMN c1699 TEXT;
SELECT dolt_commit('-A','-m','b');
-- back on main:
SELECT dolt_merge('b');   -->  SQL logic error
```

## Both conditions are required

| shape | doltlite |
| --- | --- |
| both add same column name **and** one side drops another column | **SQL logic error** |
| both add same column name, no drop | merges (at parity with Dolt) |
| one side drops, other adds a *different* name | merges |
| only one side drops and adds | merges |

## What Dolt does

Dolt raises a **schema conflict** and rolls the merge back:

```
| hash | fast_forward | conflicts | message         |
|      | 0            | 1         | conflicts found |
```

[...]

Issue #1840, opened 2026-07-29
rowid-keyed tables unsupported: rows are keyed by primary key, not rowid


timsehn:
DoltLite keys user-table rows in its prolly tree by the user's primary key
columns rather than by SQLite's auto-allocated rowid (since #358). Rowid is
allocation-order dependent and therefore history dependent, which is a hard
incompatibility with the Dolt model: two branches that independently insert
identical rows have to produce identical row identities, and a
content-addressed tree cannot do that if the key is "whichever integer this
connection handed out first".

Consequences, all of which inherited SQLite testfixture suites exercise:

- a table without an explicit INTEGER PRIMARY KEY is auto-converted to
  WITHOUT ROWID, so `rowid` is not an addressable column — `WHERE rowid = N`
  and `ORDER BY rowid` error with "no such column: rowid"
- an unordered `SELECT` returns user-PK order where stock SQLite returns
  rowid (insertion) order
- `PRAGMA table_info` reports non-PK columns differently for auto-converted
  WITHOUT ROWID tables
- AUTOINCREMENT and `sqlite_sequence` behavior tied to rowid allocation
- foreign key `SET NULL` / `SET DEFAULT` whose triggering DELETE addresses
  parent rows by rowid

This is the single largest rationale in
`test/known_testfixture_divergences.txt` — its header block has documented it
at length, but with no issue to point at, so per-suite rows in
`test/known_testfixture_exception_inventory.txt` had nowhere to attribute it.
This issue is that anchor.

Not expected to be fixed; rowid-keyed storage is incompatible with the
version-control model rather than merely unimplemented.

### Attribution status

Attributed so far are the suites whose own recorded reason names rowid
explicitly (25 divergence suites plus `btreefault`, `incrblob`, `walbig`).
That is a floor, not the true count: most rowid-driven suites rely on the
header block instead of repeating the reason inline. Widening the attribution
is a triage pass over the remaining `intentional` rows and should not be
guessed at in bulk.

timsehn:
SQL-visible `rowid` / `last_insert_rowid()` on clustered (non-INTEGER PRIMARY KEY) tables is being restored in PR #2368 without changing storage keying. Remaining work from that change:

- Stale testfixture gates to drop: #2376
- Catalog still omits the PK `sqlite_autoindex` `sqlite_master` row: #2375

timsehn:
[...]

PR #2355, opened 2026-08-21
Decide the both-add pair by drop count, not by position

timsehn:
Follow-up to #2350, which fixed the reported case but only when the dropped column was the **last** one.

#2350 pairs the added column against the ancestor slot the dropped column vacated, so a drop anywhere else slips past and the merge fails with `SQL logic error` again:

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, payload TEXT, c1113 TEXT, c1899 TEXT);
-- ours:   ADD c1699
-- theirs: DROP c1113  (a middle column, not the last)
--         ADD c1699
SELECT dolt_merge('b');   -->  SQL logic error   (before this change)
```

## What Dolt actually does

Read off Dolt 2.2.2 across fifteen shapes. It decides this pair on **how many** columns each side dropped — not which ones, and not where the added column landed:

| ours drops | theirs drops | Dolt |
| --- | --- | --- |
| ∅ | one | CONFLICT |
| one | ∅ | CONFLICT |
| one | two, overlapping | CONFLICT |
| one | two, disjoint | CONFLICT |
| one | one, disjoint | **MERGE** |
| one | one, same column | **MERGE** |
| two | two, overlapping | **MERGE** |
| ∅ | ∅ | **MERGE** |

Worth recording that two rules fit the first handful of shapes and then mispredicted, so neither is what this implements: *proper subset of dropped columns* (dies on one-vs-two disjoint, which conflicts) and *differing final column count* (dies on a one-sided drop that adds two columns, which conflicts at equal counts).

The precondition still matters: with the sides adding **different** names, a 0-vs-1 drop merges fine.

## Why counting alone is not enough

A rename leaves exactly what a drop plus an add leaves — the ancestor name gone, a new name in its slot with its type. That is precisely why #2350 paired by position, and my first attempt at counting over-refused and broke `merge_after_merged_rename_*`.

So a positional replacement still reads as a rename, and stops reading as one only when the same name appears on the other side somewhere it *cannot* be a rename. That is what makes the pair two independent adds rather than the same rename done twice.

## Verification

- `doltlite_schema_merge.sh`: **325 passed, 0 failed** (18 new tests)
- the same tests on master: **7 fail** — 3 are #2350's own, 4 are the shapes this fixes
- `vc_oracle_correct_schema_merge_matrix_test.sh` vs Dolt: **261 passed, 0 failed, 0 differing, 0 corrupting** — unchanged from baseline, so no new divergence
- all 15 oracled shapes match Dolt except the one filed as #2354
- `lint_layers.sh` passes; `doltlite_merge_schema.c` at 1399/1500 lines
- amalgamation regenerated and compiled — two new statics, no collisions

## Left open

#2354: equal drop counts where one side drops the **last** column still fails with `SQL logic error`, and Dolt merges it — so a refusal would be the wrong answer and it needs the merge itself to work. Reproduces on master and is untouched by both #2350 and this change.

### DIFF
--- a/src/doltlite_merge_schema.c
+++ b/src/doltlite_merge_schema.c
@@ -633,6 +633,41 @@ static int schemaConstraintModifyDeleteChoice(
   return rc;
 }
 
+/* True when zName sits on this side as a plain addition: absent from the
+** ancestor, and not standing in an ancestor column's slot with its type,
+** which is what a rename leaves behind. */
+static int columnPlainlyAdded(
+  ParsedColumn *aSide, int nSide,
+  ParsedColumn *aAnc, int nAnc,
+  const char *zName
+){
+  int i = parsedColumnIndexByName(aSide, nSide, zName);
+  if( i<0 || findColumn(aAnc, nAnc, zName) ) return 0;
+  if( i<nAnc
+   && !findColumn(aSide, nSide, aAnc[i].zName)
+   && parsedColumnDefinitionsMatch(&aSide[i], &aAnc[i]) ){
+    return 0;
+  }
+  return 1;
+}
+
+/* True when the ancestor column at iAnc left this side by being renamed
+** rather than dropped. The other side decides the ambiguous case: if the
+** replacement's name is a plain addition over there, both sides added it
+** and this side really did drop a column. */
+static int columnRenamedAt(
+  ParsedColumn *aSide, int nSide,
+  ParsedColumn *aAnc, int nAnc,
+  int iAnc,
+  ParsedColumn *aOther, int nOther
+){
+  if( iAnc>=nSide ) return 0;
+  if( findColumn(aAnc, nAnc, aSide[iAnc].zName) ) return 0;
+  if( !parsedColumnDefinitionsMatch(&aSide[iAnc], &aAnc[iAnc]) ) return 0;
+  if( columnPlainlyAdded(aOther, nOther, aAnc, nAnc, aSide[iAnc].zName) ) return 0;
+  return 1;
+}
+
 int trySchemaColumnMerge(
   const char *zAncSql,
   const char *zOursSql,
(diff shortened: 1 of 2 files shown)


timsehn:
Seed verification: `1787347777` — the seed that found #2349 — now completes **4936 steps** clean against this branch, having previously failed deterministically at step 2398 on both Linux CI and macOS.

```
DOLTLITE_VC_STATEFUL_SEED=1787347777 DOLTLITE_VC_STATEFUL_SECONDS=2400 \
  python3 test/vc_stateful_fuzzer.py build/doltlite
OK: stateful vc fuzzer completed 4936 steps across 10 branches
```

Note the fuzzer's own database happened to hit the shape that #2350 already covers, so that seed clears on master too. It is included here as an end-to-end check rather than as evidence specific to this change — the fail-before/pass-after evidence for this change is the four `merge_bothadd_*` tests, which fail on master and pass here.

timsehn:
@itoqa Thanks — real bug, and it reproduces. It is **pre-existing on master**, not introduced here, so I have filed it as #2357 rather than fixing it in this PR.

**Not caused by this PR.** I built master `c34746fc29` (which has #2350 but not this change) in a separate worktree and ran your script verbatim: same `database disk image is malformed`. The code analysis pointed at the rename-selection path this PR touches, but that path is not what breaks it.

**The characterisation is different from the report.** The rename is not required, and neither is the index-on-renamed-column framing. The discriminator is the **absence of a `PRIMARY KEY`**, verified in separate databases:

| table | merge shape | indexed lookup |
| --- | --- | --- |
| no PK | diverged (both sides changed, no rename) | **malformed** |
| no PK | column rename, fast-forward | **malformed** |
| `a INTEGER PRIMARY KEY` | diverged | ok |
| `a INTEGER PRIMARY KEY` | column rename, fast-forward | ok |

So a diverged merge with no rename at all is enough to trigger it, and adding a primary key makes your exact script pass.

**It is the index, not the database.** The table schema is unchanged across the merge, the row data is intact (`count(*)` = 3), `PRAGMA integrity_check` returns `ok`, and `REINDEX` repairs it. Only lookups routed through the index fail. Calling it `SQLITE_CORRUPT` is itself part of the problem — it reads as file damage when the data is fine.

This sits in the area of #1840 (a table without an explicit `INTEGER PRIMARY KEY` is auto-converted to `WITHOUT ROWID`), but the failure mode is not in that issue's list, and `integrity_check` returning `ok` while an index reports corruption is worth fixing on its own. Details and the minimised repro are in #2357.

PR #2368, opened 2026-08-22
Restore SQL rowid on clustered primary keys

timsehn:
## Summary

Non-`INTEGER PRIMARY KEY` tables are still clustered `WITHOUT ROWID` so merge identity is the user PK. They no longer hide SQL `rowid` / `last_insert_rowid()`.

- Auto-converted tables keep `VisibleRowid`. A single integer PK is that value; otherwise a stable hash of the PK.
- Explicit `WITHOUT ROWID` still has no `rowid`, matching SQLite.
- `sqlite3BtreeIntegerKey` stays intkey-only. Clustered PK SQL rowid goes through `sqlite3BtreeSqlRowid`.
- `WHERE rowid = ?` does not `SeekRowid` on blobkey tables.

## Tests

Fail-before / pass-after: `test/doltlite_rowid_identity.sh` (12/12).

Also green locally: sqlite compatibility contract, autoinc schema, merge (137), schema merge (372), index VC matrix (49), dolt table pushdown (49), `make lint`.


timsehn:
Pushed `513be68bf9` for the leftover testfixture CI:

- Dropped 9 stale core-sql gates that now pass because clustered PKs expose SQL `rowid` (`delete2-1.6/1.9`, `misc7-10/11/12.67.3/13/15.2`, `whereF-4.0`, `without_rowid1-7.1`).
- `btreefault` now completes (203 tests); removed its crash/termination contract.
- Ratchet: `gates 4769`, `intentional 3422`.

`Smoke / vc-fuzz` (`seed=1787436222 step=3346 dolt_merge('b191')` → `SQL logic error`) is separate. That fuzzer only uses `INTEGER PRIMARY KEY` tables, so it should not be this rowid change. Watching the new run.

timsehn:
Rebased onto current `master` (`a40283661d`). The only conflict was `test/known_testfixture_exception_ratchet.txt` (master retired a dbpage gate). Ratchet is now `gates 4768` / `intentional 3422` / `unsupported 1185`.

PR #2370, opened 2026-08-22
A table rename merges as a rename, not a drop plus a create

timsehn:
The second dangler: the un-rename-and-replay path from #2363 ran only when a trigger forced it. Every other table rename merged as a drop plus a create, so the other side's changes died in a spurious conflict. This removes the gate: **every pure table rename now composes** — row edits and inserts from the other side land in the renamed table, column adds/renames carry across, and indexes, views and triggers follow, in both directions.

```
base:   t(k,a,n) = {1}          base:   t(k,a,n) with trigger, index
ours:   RENAME t TO t9, edit    ours:   RENAME t TO t9
theirs: INSERT row 3            theirs: DROP COLUMN b / ADD COLUMN / edit rows
→ t9 with both changes          → t9 with the schema change composed, dependents following
```

## Deliberate divergence, at its strongest

Dolt cannot tell a rename from a drop (dolthub/dolt#2016-class) and **conflicts on every such pair** — so did we, until now. 52 matrix pairs move to a new `MERGE_WHERE_DOLT_CONFLICTS_BY_DECREE` list: scored as passes with the reason printed, policed for staleness in both directions like the other lists. This is the git model: a rename composes with content edits.

## The hazard the gate was masking

Widening the detection exposed a real pairing bug: with tables of identical column lists (`a(id,payload)` and `c(id,payload)`), a **dropped** table could be text-matched as the rename source of an unrelated new table, silently swallowing the conflict that was protecting it — caught by `index_on_excluded_rename_is_not_adopted` in the index-conflict suite. The pairing is now **bijective**: a new table whose text matches more than one vanished ancestor table stays a drop plus a create. Fail-before/pass-after on that suite (33/34 → 34/34).

Unchanged on purpose: renames to two different names still merge as one drop and two creates (verified byte-identical to master), rename-vs-drop likewise, and replays are untouched (the pass only runs where action plumbing exists).

## Verification

| check | result |
| --- | --- |
| `vc_oracle_correct_schema_merge_matrix_test.sh` vs Dolt | **264 passed, 0 failed, 0 gaps, 0 differing, 0 corrupting**; all 52 decree entries exercised |
| `doltlite_schema_merge.sh` | 386 passed, 0 failed (new: composes-rows both directions, rename+edit compose, addcol compose, dual-rename control) |
| `doltlite_merge_index_conflict.sh` | 34 passed (the suite that caught the pairing hazard) |
| `doltlite_index_vc_matrix.sh` | 49 passed |
| `run_doltlite_tests.sh` | 112 suites passed |
| stateful vc fuzzer | seed 1787347777, 2683 steps clean |
| lint / amalgamation | pass; compiles under CI's warning flags |

Also verified along the way: the "resurrection" my first spot-check appeared to find was a harness artifact — stock SQLite refuses `DROP COLUMN` on an indexed column, so the setup had silently failed; the matrix's setup-rejection detector already handles that class correctly.

### DIFF
--- a/src/doltlite_merge_predetect.c
+++ b/src/doltlite_merge_predetect.c
@@ -781,7 +781,7 @@ static int mergePreNormalizeTableRename(
       SchemaEntry *pAncT = &aAncSchema[t];
       SchemaEntry *pNewT = 0;
       char *azPair[2];
-      int nCand = 0, bTrigger = 0;
+      int nCand = 0;
 
       if( !pAncT->zType || strcmp(pAncT->zType, "table")!=0 ) continue;
       if( !pAncT->zName || !pAncT->zSql ) continue;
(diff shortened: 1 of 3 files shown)


timsehn:
@itoqa Taken — the finding is real and now fixed in this PR (it hardens the very mechanism this PR extends). A run with zero passes now exits 1 with `no pair produced a comparison: the reference engine is unusable`; verified against your exact repro (`/nonexistent/dolt` → previously `0 passed … 308 not comparable`, exit 0; now exit 1).

Zero-pass is the guard rather than nonzero skips/gaps deliberately: 44 skips are legitimate today (setup-DDL asymmetries between engines), and gaps carry their own policy — but a healthy run always compares hundreds of pairs, so `pass==0` cleanly separates 'the oracle validated nothing' from 'some pairs are not comparable'.

PR #2371, opened 2026-08-22
Preserve SQLite affinity in staged filters

timsehn:
## Summary

- share exact integer argument classification between primary-key and staged-value pushdown
- narrow `dolt_status` and `dolt_workspace_*` scans only for exact `0` and `1` values
- keep SQLite's equality recheck so fractional, nonnumeric text, and BLOB constraints cannot fabricate matches
- preserve INTEGER affinity for true integer-primary-key aliases in `dolt_history_*`, `dolt_at_*`, and `dolt_blame_*`
- cover staged filters and historical numeric-text lookups with regressions

## Fail-before

An unstaged `dolt_status` row was incorrectly returned for `staged='abc'`, `staged=0.5`, and `staged=x'00'`; the equivalent one-side fractional case also failed for a staged row.

Ito additionally found that `id='1'` missed rows returned by `id=1` in historical views. The generated virtual schemas lacked the INTEGER affinity needed by SQLite's residual equality recheck.

## Validation

- `make -B -C build doltlite`
- `make lint`
- `test/doltlite_dolt_table_pushdown.sh` (90 passed)
- native history, point-in-time, and workspace suites (141 passed)
- Dolt history, point-in-time, blame, status, and workspace oracles (173 passed)
- `test/run_doltlite_tests.sh` (113/113 suites; 310,930/310,930 C regression assertions)

### DIFF
--- a/src/doltlite_at.c
+++ b/src/doltlite_at.c
@@ -22,7 +22,8 @@ static char *atBuildSchema(const DoltliteColInfo *ci){
   char *z;
   if( !pStr ) return 0;
   sqlite3_str_appendall(pStr, "CREATE TABLE x(");
-  if( doltliteAppendQuotedColumnList(pStr, ci->azName, ci->nCol, 0, 0)!=SQLITE_OK ){
+  if( doltliteAppendIntegerPkColumnList(pStr, ci->azName, ci->nCol,
+                                        ci->iPkCol)!=SQLITE_OK ){
     sqlite3_str_reset(pStr);
     return 0;
   }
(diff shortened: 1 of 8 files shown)


timsehn:
@itoqa Both numeric-text findings are addressed in 5b1a34862a. The historical, point-in-time, and blame virtual schemas now declare true INTEGER PRIMARY KEY aliases with INTEGER affinity, so numeric text survives SQLite residual rechecks while fractional, BLOB, NULL, and nonnumeric values remain rejected. I added fail-before/pass-after coverage for all three surfaces; the focused Dolt oracles and full 113-suite native battery pass.

PR #2372, opened 2026-08-22
Honor sqlite_dbpage schema selection

timsehn:
## Summary

- resolve `sqlite_dbpage`'s hidden `schema` argument before page lookup
- synthesize page 1 only for `main`; return no rows for unknown, NULL, or attached schemas
- report `main` through the hidden output column
- cover default, explicit, case-insensitive, unknown, NULL, attached, and page-error ordering cases
- retire inherited `dbpage-240` from the unsupported inventory now that it passes

## Fail-before

`schema='nosuch'`, `sqlite_dbpage(NULL)`, and an attached schema all returned the synthesized main page. The hidden output column was NULL, and an unsupported schema paired with `pgno=99` raised the page-layout error instead of returning no rows.

## Validation

- `make -C build doltlite testfixture`
- `make lint`
- `test/doltlite_dbpage.sh` (14 passed)
- inherited `dbpage` in default and coverage variants (10 passed; 24 existing documented divergences)
- `doltlite_recover_rawformat_3` (102 passed)
- stock-file open and SQLite oracle suites (205 passed)
- `test/run_doltlite_tests.sh` (113/113 suites; 310,930/310,930 C regression assertions)

### DIFF
--- a/src/doltlite_dbpage.c
+++ b/src/doltlite_dbpage.c
@@ -121,24 +121,25 @@ static int dlDbpageBestIndex(sqlite3_vtab *pVtab, sqlite3_index_info *pInfo){
   (void)pVtab;
 
   for(i=0; i<pInfo->nConstraint; i++){
-    if( !pInfo->aConstraint[i].usable ) continue;
     if( pInfo->aConstraint[i].op!=SQLITE_INDEX_CONSTRAINT_EQ ) continue;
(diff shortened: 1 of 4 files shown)


timsehn:
The storage-bucket failure was a newly passing upstream assertion, not a regression. fb81ee9b8d removes the stale dbpage-240 exception and lowers the ratchet. The dbpage suite now passes in both default and coverage variants with 24 remaining documented raw-page divergences; lint is green.

Issue #1016, opened 2026-05-21
VACUUM limitations on doltlite: no VACUUM INTO, no VACUUM in transaction


timsehn:
## Summary

Per the resolution of #322, \`VACUUM\` on a doltlite database now aliases to \`SELECT dolt_gc()\` (\`src/vacuum.c:170-196\`). Two SQLite VACUUM behaviors remain unsupported and return errors:

\`\`\`sql
VACUUM INTO 'other.db';
-- Error: VACUUM INTO is not supported for doltlite databases

BEGIN;
VACUUM;
-- Error: cannot VACUUM from within a transaction
\`\`\`

The first is a real gap (SQLite supports copying the vacuumed DB to a separate file). The second matches SQLite's own behavior and is intentional.

## VACUUM INTO

Implementing this would mean writing the current chunk store to a new file at a different path. \`dolt_gc()\` already does this internally via a \`*-gc-tmp\` rename — exposing that destination as user-supplied would be the cleanest hook.

Closest mechanism today: \`dolt_clone('file:///path/to/dest.db')\` from a fresh DB. Not a drop-in replacement for \`VACUUM INTO\` (different argument shape, and it requires the dest to be empty), but functionally equivalent.

## Not blocking anything

Both gaps are documented errors with clear messages, not silent data loss. Filing as a tracker; no urgency unless a real workload hits it.

timsehn:
**Update (2026-08-22):** the in-transaction claim here is stale.

Current DoltLite (`VACUUM` → `dolt_gc`) returns **success** for `BEGIN; VACUUM;` and does not compact. SQLite still errors `cannot VACUUM from within a transaction`. `VACUUM INTO` is still unimplemented as described.

Agent-ready repro and suggested fix: #2373

timsehn:
The claim here that DoltLite already errors on VACUUM inside a transaction was stale — it silently no-opped with SQLITE_OK (#2373). As of #2378 the statement path errors with `cannot VACUUM from within a transaction`, matching stock, and `VACUUM INTO` now writes a compacted doltlite-format copy under stock's target contract. Both are pinned in `test/sqlite_compatibility_contract.tsv` (`pragma.vacuum_txn`, `pragma.vacuum_into`).

timsehn:
[...]

PR #2378, opened 2026-08-22
VACUUM in a transaction errors like stock, and VACUUM INTO works

### LINKED ISSUE #2373: VACUUM inside a transaction is a silent no-op; VACUUM INTO is missing
## Summary

On a DoltLite-format database, `VACUUM` is a GC bridge (`dolt_gc`), not SQLite page rewrite. Two user-visible holes:

1. **`VACUUM` inside an open transaction returns success and is a no-op.** SQLite errors. Tracker #1016 claims DoltLite already errors like SQLite — that is **stale**.
2. **`VACUUM INTO 'path'` is unimplemented** (`VACUUM INTO is not supported for doltlite databases`). SQLite writes a copy.

This is original SQL-layer review item 4. Independent of PR #2368.

## Repro (DoltLite CLI)

Build: `cd build && ../configure && make doltlite`. Use `build/doltlite`, not a repo-root `./sqlite3` (that binary can be DoltLite-in-disguise). Compare with Homebrew SQLite (`/opt/homebrew/opt/sqlite/bin/sqlite3` is 3.53.2 here).

### A. `VACUUM INTO`

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, v INT);
INSERT INTO t VALUES(1,42);
VACUUM INTO 'copy.db';
```

| | result |
|---|---|
| SQLite | writes `copy.db` |
| DoltLite | `VACUUM INTO is not supported for doltlite databases` |

Verified 2026-08-22 on `build/doltlite` v0.11.52.

### B. `VACUUM` in a transaction

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, v INT);
INSERT INTO t VALUES(1,1);
BEGIN;
INSERT INTO t VALUES(2,2);
VACUUM;
COMMIT;
SELECT * FROM t;
```

| | `VACUUM` | after `COMMIT` |
|---|---|---|
| SQLite | `cannot VACUUM from within a transaction` | rows `1` and `2` (VACUUM did not run; txn still committed) |
| DoltLite | **`SQLITE_OK`**, no compaction | rows `1` and `2` |

False success. Apps that treat `VACUUM` as “rewrite now” get a silent no-op.

## Root cause

`src/vacuum.c` (~169–203), `#ifdef DOLTLITE_PROLLY` path:

- `pOut` set → error string for `VACUUM INTO`.
- Open transaction is **not** rejected. Comment says `doltliteGcCompactDbWithPhase` treats `!autoCommit` as a successful no-op because checkpoint-driven compaction uses that contract. Stock’s “cannot VACUUM from within a transaction” only applies to the page-vacuum path below.

#1016 documents the in-txn case as already matching SQLite. Update or close it when this lands.

## Suggested fix

Do these as two commits in one PR if both are cheap; split if `INTO` is large.

1. **In-txn `VACUUM` (small, do first)**  
   If the caller is a user `VACUUM` statement and `!db->autoCommit` (or equivalent), return `SQLITE_ERROR` with `cannot VACUUM from within a transaction`, matching SQLite.  
   **Do not** break checkpoint-driven GC if it calls the same C helper while a txn is open. Grep `doltliteGcCompactDbWithPhase` / `VACUUM` callers. The SQL statement must error; an internal GC helper may still no-op.

2. **`VACUUM INTO`**  
   Copy the chunk-store to the destination path. GC already uses a temp file + rename (`*-gc-tmp`). Closest user-facing mechanism today is `dolt_clone('file:///path')` into an empty dest — not a drop-in. README already says `.backup` / `VACUUM INTO` apply `doltlite_engine=sqlite` when the **source** is a stock file; this issue is DoltLite-format **source**.

Do **not** bump `CHUNK_STORE_VERSION`. Destination file should be a valid DoltLite database with the same committed data.

## Tests

Fail-before / pass-after. Suggested:

- `test/doltlite_vacuum_txn.sh` (or extend `test/doltlite_vacuum.sh`): `BEGIN; VACUUM;` must error; rows after `COMMIT` still present.
- `VACUUM INTO 'copy.db'` then reopen `copy.db` and `SELECT` the row.
- Keep existing `test/doltlite_vacuum.sh#vacuum_and_dolt_gc_leave_same_state` green.
- If you document the in-txn error as the contract, add a needle to `test/sqlite_compatibility_contract.tsv` and the matching assertion in `test/sqlite_compatibility_contract_test.sh`.

## Agent instructions

- Branch off a **freshly pulled `master`**. PR against `master`. Do not stack on #2368.
- Stage by explicit path (`git add src/vacuum.c test/...`). Never `git add -A`.
- Guard with `DOLTLITE_PROLLY`. `DOLTLITE_PROLLY=0` must keep stock SQLite VACUUM (including in-txn error and `VACUUM INTO`).
- Never delete a test to make CI green.
- Commit message and PR body trailer: `Co-Authored-By: Grok 4.6 <noreply@x.ai>`
- End with a PR.

## Related

- #1016 (stale: claims in-txn VACUUM already errors)
- SQL-layer review item 4

timsehn:
Fixes #2373 — both holes.

## 1. `VACUUM` inside a transaction errors

The statement path now fails with `cannot VACUUM from within a transaction`, matching stock, and the enclosing transaction still commits (stock-identical: the error doesn't poison the txn). Scope is surgical: `sqlite3RunVacuum` has exactly **one** caller (the `VACUUM` opcode), so checkpoint-driven compaction and `dolt_gc()` — which reach `doltliteGcCompactDbWithPhase` directly — keep their documented in-txn no-op contract untouched.

## 2. `VACUUM INTO` writes a compacted doltlite-format copy

The write phase of `gcRewriteFile` moves into a destination-agnostic helper, `gcWriteCompactedTo`: GC keeps writing to its `-gc-tmp` path and swapping; `VACUUM INTO` writes the same image to the target and swaps nothing — the source store is only read. The copy is the image GC would have swapped in: same committed history, current working state, unreachable chunks dropped (verified: the copy of a shrunk table is no larger than its source). No `CHUNK_STORE_VERSION` bump; the target path gets the VFS double-nul terminator.

Stock's target contract is kept exactly: non-empty existing file → `output file already exists`; empty existing file → overwritten; `VACUUM INTO` inside a transaction → same in-txn error. A stock-format source through the doltlite binary still takes the stock page-vacuum path (verified with `doltlite_engine=sqlite`).

## Tests

Fail-before/pass-after in both directions: the two suite tests that asserted the old behaviors (`vacuum_into_rejected`, `vacuum_in_txn_noop`) — which passed on master — are replaced by stock-matching counterparts, plus new coverage: copy readability/history/integrity, copy↔source independence, compaction effect, empty-target acceptance, existing-target rejection, in-txn INTO. The in-txn error and INTO contract are pinned in `sqlite_compatibility_contract.tsv` with resolving needles.

| check | result |
| --- | --- |
| `doltlite_vacuum.sh` | 24 passed (was 15) |
| `doltlite_gc.sh` (refactor neutrality) | 58 passed |
| `run_doltlite_tests.sh` | **113 suites passed** |
| stateful vc fuzzer (op set includes `gc`) | seed 555111, 2187 steps clean |
| `sqlite_compatibility_contract_test.sh` | 5 passed with the two new rows |
| lint / amalgamation | pass; compiles under CI's warning flags |

Closes #2373. #1016's stale claim (that the in-txn case already errored) is corrected there.


timsehn:
@itoqa The race is real and the fixable half is now fixed: the emptiness check ran **before** the lock and open, so a writer filling the target in that window got clobbered under a success return. The check now runs on the just-opened handle — the only authority on what the write would destroy — and refuses a non-empty target while leaving the writer's file untouched. The advisory pre-check is deleted outright, so the deterministic suite tests (`vacuum_into_existing_target_errors`, empty-target acceptance) now exercise the authoritative path.

For the record on the residual: a writer that clobbers the target **after the copy completes** is not detectable by any check the copier can make. Under your 200-attempt harness shape, stock SQLite exhibits the same class (38/200 sentinel-success on this machine); rates vary only with operation duration, and an advisory lock would not constrain a non-SQLite writer. The window your test targeted — target becomes non-empty before the write begins — is closed.

PR #2421, opened 2026-08-25
Treat all-NULL sqlite_schema rows as malformed on load

### LINKED ISSUE #2417: All-NULL sqlite_schema rows are skipped instead of malformed database schema
## Summary

Stock `sqlite3InitCallback` treats a `sqlite_schema` row whose type/name/root/sql are NULL as corrupt (`malformed database schema`). DoltLite returns 0 and continues (`src/prepare.c` ~112–120).

The skip is aimed at OOM placeholders on the prolly master root. It is also reachable with `writable_schema`: an all-NULL catalog row is silently dropped on the next connection, and the database opens.

Same-connection `SELECT` after the INSERT matches stock (both show the user table). The divergence is on **reopen**, when init walks the catalog.

Verified 2026-08-25 on `build/doltlite` `v0.11.53-90-gcb2d113ebd` vs Homebrew SQLite 3.43.2.

## Repro (DoltLite CLI)

`cd build && ../configure && make doltlite`.

```sql
CREATE TABLE t(a);
PRAGMA writable_schema=ON;
INSERT INTO sqlite_master VALUES (NULL, NULL, NULL, NULL, NULL);
PRAGMA writable_schema=OFF;
```

Close the connection and reopen the same file:

```sql
SELECT name FROM sqlite_master;
```

| | DoltLite | SQLite |
|---|---|---|
| reopen | `t` | `malformed database schema (?)` |

A genuinely corrupt (non-NULL) schema row should still fail init on both.

timsehn:
## Summary

Fixes #2417.

Stock schema init treats a `sqlite_schema` row with a NULL rootpage as corrupt (`malformed database schema (?)`). DoltLite skipped all-NULL rows in `sqlite3InitCallback` so leftover OOM placeholders would not fail open. `PRAGMA writable_schema` can insert that same shape, so the next connection opened and hid the bad row.

Remove the skip. Init uses the stock `argv[3]==0` path.

No storage-format change. No `CHUNK_STORE_VERSION` bump.

## Tests

- `test/doltlite_writable_schema.sh#all_null_schema_row_malformed_on_reopen`: fail-before (`got: t`), pass-after (`malformed database schema`). Suite 7/7.
- Dropped `capi3-8.3` and `capi3c-8.3` from `known_testfixture_divergences.txt`. Remaining failures in those files are the still-gated lock_status / MVCC cases. Ratchet `gates 4727 → 4725`, `intentional 3404 → 3402`.
- `make lint` holds.

Local: suites above on `311ff575ea` + this commit.

## Agent notes

- Branch is off current `master`.
- Staged by explicit path.
- Never deleted a test.

### DIFF
--- a/src/prepare.c
+++ b/src/prepare.c
@@ -109,17 +109,6 @@ int sqlite3InitCallback(void *pInit, int argc, char **argv, char **NotUsed){
     return 1;
   }
 
-#ifdef DOLTLITE_PROLLY
-  /* doltlite's prolly master root can retain an all-NULL placeholder record
-  ** after a schema write is interrupted by OOM (the same empty records #1224
-  ** drops when rebuilding the root; they persist until the next rebuild). Such
-  ** a row carries no schema object -- skip it rather than reporting the
-  ** database as malformed. A genuinely corrupt row never presents as all-NULL. */
-  if( argv[0]==0 && argv[1]==0 && argv[3]==0 && argv[4]==0 ){
-    return 0;
-  }
-#endif
-
   assert( iDb>=0 && iDb<db->nDb );
   assert( db->aDb[iDb].pSchema!=0 );
   if( argv[3]==0 ){
--- a/test/doltlite_recover_locking_1.test
+++ b/test/doltlite_recover_locking_1.test
@@ -445,13 +445,15 @@ do_test doltlite_recover_locking_1-10.3 {
   }]
   catchsql {PRAGMA writable_schema=OFF}
   db close
-  set r2 [catch {sqlite3 db rloc1.db}]
-  set DB [sqlite3_connection_pointer db]
-  list $r1 $r2 [execsql {SELECT count(*) FROM x1}] \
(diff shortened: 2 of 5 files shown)


timsehn:
Automated fix: sqlite-regression-ported failed because doltlite_recover_locking_1-10.3 still expected reopen to succeed after an all-NULL sqlite_master row. Updated that test to match stock (malformed database schema) and recreated the db for 10.4–11.2.

timsehn:
Rebased onto `origin/master` after #2420 merged (`a2ff9dc3e0`). Master was no longer an ancestor of this branch.

Conflict resolution:
- `test/known_testfixture_divergences.txt` kept master's content plus this PR's deletion of `capi3-8.3` and `capi3c-8.3`.
- `test/known_testfixture_exception_ratchet.txt` started from master's `gates 4712` / `intentional 3389` and subtracted 2 → `4710` / `3387`.

`bash test/check_testfixture_exception_inventory.sh` passed: 4710 gates classified (intentional=3387). Force-pushed `3b67b8316e`.

PR #2432, opened 2026-08-25
Fire sqlite3_update_hook on clustered PRIMARY KEY tables

### LINKED ISSUE #2427: sqlite3_update_hook is silent on clustered TEXT PRIMARY KEY tables
## Summary

`sqlite3_update_hook` fires on stock SQLite `TEXT PRIMARY KEY` tables because those are rowid tables (`OP_Insert` / `OP_Delete`). DoltLite auto-clusters a non-INTEGER PRIMARY KEY as WITHOUT ROWID (`src/build.c` ~2793–2817). Writes go through `OP_IdxInsert`, and `OP_Delete` only invokes the hook when `HasRowid(pTab)`.

**INSERT, UPDATE, and DELETE on a clustered TEXT PK table never call the hook.** INTEGER PRIMARY KEY still matches stock. Explicit `WITHOUT ROWID` is silent on both engines (SQLite's documented WITHOUT ROWID rule).

This is the C-API face of clustering, in the same family as `sqlite3_blob_open` (#2394) and `RETURNING`/`NEW.rowid` (#2391). `commit_hook` / `rollback_hook` still fire. Triggers still fire (`OLD.rowid` / `NEW.rowid` on clustered tables already match).

Verified 2026-08-25 on `build/doltlite` `v0.11.54-9-g7b0905b26a` (SQLite 3.54.0) vs Homebrew SQLite 3.43.2.

## Repro

```c
static void hook(void *p, int op, const char *db, const char *tbl, sqlite3_int64 rowid){
  printf("op=%d tbl=%s rowid=%lld\n", op, tbl, (long long)rowid);
}
sqlite3_update_hook(db, hook, 0);
sqlite3_exec(db,
  "CREATE TABLE t(k TEXT PRIMARY KEY, v INT);"
  "INSERT INTO t VALUES('a',1),('b',2);"
  "UPDATE t SET v=3 WHERE k='a';"
  "DELETE FROM t WHERE k='b';", 0,0,0);
```

| | hook calls |
|---|---|
| DoltLite | *(none)* |
| SQLite | `INSERT t 1`, `INSERT t 2`, `UPDATE t 1`, `DELETE t 2` |

INTEGER PRIMARY KEY control (same ops on `t(a INTEGER PRIMARY KEY, v INT)`): both engines print INSERT/UPDATE/DELETE for rowid 1.

Explicit `CREATE TABLE t(k TEXT PRIMARY KEY, v INT) WITHOUT ROWID`: both silent.

## Source

- `src/build.c` ~2793–2817: non-INTEGER PK → `TF_WithoutRowid` while keeping `VisibleRowid`.
- `src/insert.c` ~2871–2886: clustered PK insert is `OP_IdxInsert`, not `OP_Insert`.
- `src/vdbe.c` ~6198–6253: `OP_Insert` update hook; asserts `HasRowid` unless `OPFLAG_ISNOOP`.
- `src/vdbe.c` ~6417–6418: `OP_Delete` hook only if `HasRowid(pTab)`.

timsehn:
## Summary

`sqlite3_update_hook` now fires for auto-clustered non-INTEGER PRIMARY KEY tables (for example `TEXT PRIMARY KEY`), using the same synthetic SQL `rowid` as `SELECT rowid` / `last_insert_rowid()`.

Explicit `WITHOUT ROWID` stays silent, matching stock SQLite.

INTEGER PRIMARY KEY is unchanged (still `OP_Insert` / `OP_Delete`).

## Why

DoltLite stores those tables as WITHOUT ROWID (`VisibleRowid` still set). Writes used `OP_IdxInsert`, which had no update-hook, and `OP_Delete` only called the hook when `HasRowid`.

## Tests

- C regression `clustered_pk_update_hook`: INTEGER PK events, TEXT PK INSERT/UPDATE/DELETE with matching rowids, explicit WITHOUT ROWID silent.
- `hook.test` 34/34.

Fixes #2427.

### DIFF
--- a/src/insert.c
+++ b/src/insert.c
@@ -2879,7 +2879,19 @@ void sqlite3CompleteInsertion(
         }
 #endif
       }
+#ifdef DOLTLITE_PROLLY
+      if( VisibleRowid(pTab) ){
+        pik_flags |= (update_flags & OPFLAG_ISUPDATE);
+      }
+#endif
     }
+#ifdef DOLTLITE_PROLLY
+    if( (pik_flags & OPFLAG_NCHANGE)!=0 && VisibleRowid(pTab)
+     && !pParse->nested ){
+      sqlite3VdbeAddOp3(v, OP_IdxInsert, iIdxCur+i, aRegIdx[i], aRegIdx[i]+1);
+      sqlite3VdbeAppendP4(v, pTab, P4_TABLE);
+    }else
+#endif
     sqlite3VdbeAddOp4Int(v, OP_IdxInsert, iIdxCur+i, aRegIdx[i],
                          aRegIdx[i]+1,
                          pIdx->uniqNotNull ? pIdx->nKeyCol: pIdx->nColumn);
@@ -3518,6 +3530,12 @@ static int xferOptimization(
       }
     }
     sqlite3VdbeAddOp2(v, OP_IdxInsert, iDest, regData);
+#ifdef DOLTLITE_PROLLY
+    if( (idxInsFlags & OPFLAG_NCHANGE)!=0 && VisibleRowid(pDest)
+     && !pParse->nested ){
+      sqlite3VdbeAppendP4(v, pDest, P4_TABLE);
+    }
+#endif
     sqlite3VdbeChangeP5(v, idxInsFlags|OPFLAG_APPEND);
     sqlite3VdbeAddOp2(v, OP_Next, iSrc, addr1+1); VdbeCoverage(v);
     sqlite3VdbeJumpHere(v, addr1);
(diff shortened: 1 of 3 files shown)


timsehn:
CI `sqlite-regression-ddl-schema` failed on `fkey8-7.4`: `ON UPDATE CASCADE` against a clustered (non-INTEGER) PRIMARY KEY duplicated child rows instead of rewriting the FK.

Cause: `OP_IdxInsert` attached `P4_TABLE` for the update hook and zeroed the unpacked key (`x.nMem`). WITHOUT ROWID then compared the full packed record, missed the existing PK, and inserted a second row.

Fix in 6694f8d7f0: keep `P4_TABLE` for the hook, but restore unpacked `nMem` from `KeyInfo.nKeyField` when P3 names the key registers. Rebased onto current `master`. Local `fkey8` (38/38), `hook.test` (34/34), and `clustered_pk_update_hook` (20/20) pass.

PR #2465, opened 2026-08-27
Backup retargets a dest session parked on a branch absent from the image

### LINKED ISSUE #2459: sqlite3_backup wedges the live dest connection when its checked-out branch is absent from the source image
Found in the 2026-08-27 deep review at d6fa0ef90c. **HIGH.**

`sqlite3_backup_init/step/finish` into a destination connection whose session is checked out on a branch that does **not exist in the source image** produces a correct destination *file* but leaves the live destination *connection* broken with no in-session recovery:

- `active_branch()` still reports the vanished branch; `sqlite_master` still shows the pre-backup schema.
- **Phantom reads**: the dest's old table still returns its old rows; the source's tables are "no such table".
- `PRAGMA integrity_check` reports failure on a healthy database.
- Every write returns SQLITE_BUSY (`database is locked`), persistently.
- `dolt_status` fails rc=12 "unknown operation"; `dolt_checkout('main')` fails "failed to snapshot current branch state".

Only close/reopen recovers (reopen lands on the image's default branch and everything is fine). Control case — dest session on `main`, which exists in the image — is fully clean, so the missing-branch case is the defect.

## Repro (deterministic, verified twice)

Compile against the doltlite static lib and run; case 0 (dest on main) is clean, case 1 shows the wedge:

```
cc -I. repro_backup_wedge.c build/libdoltlite.a -lz -lm -lpthread -o repro_backup_wedge
./repro_backup_wedge /tmp/bkwedge
```

```c
/* Backup onto a dest connection: case1 dest on 'main' (exists in src image),
** case2 dest on a branch missing from the src image. Probe the live dest
** connection after finish: reads, writes, status, recovery via checkout. */
#include "sqlite3.h"
#include <stdio.h>
#include <string.h>
#include <stdlib.h>
#include <unistd.h>

static const char *q(sqlite3 *db, const char *zSql){
  static char buf[1024];
  sqlite3_stmt *st = 0;
  if( sqlite3_prepare_v2(db, zSql, -1, &st, 0)!=SQLITE_OK ){
    snprintf(buf, sizeof(buf), "(prep-err %s)", sqlite3_errmsg(db)); return buf;
  }
  int rc = sqlite3_step(st);
  if( rc==SQLITE_ROW ){
    const unsigned char *z = sqlite3_column_text(st, 0);
    snprintf(buf, sizeof(buf), "%s", z?(const char*)z:"(null)");
  }else snprintf(buf, sizeof(buf), "(rc=%d %s)", rc, sqlite3_errmsg(db));
  sqlite3_finalize(st);
  return buf;
}

static void kase(const char *dir, int destOnMissingBranch){
  char a[600], b[600];
  printf("== case destOnMissingBranch=%d ==\n", destOnMissingBranch);
  sprintf(a, "%s/repro_backup_wedge_src%d.db", dir, destOnMissingBranch); unlink(a);
  sprintf(b, "%s/repro_backup_wedge_dst%d.db", dir, destOnMissingBranch); unlink(b);
  sqlite3 *src=0, *dst=0;
  sqlite3_open(a, &src);
  sqlite3_exec(src, "CREATE TABLE s(x); INSERT INTO s VALUES(1);"
                    "SELECT dolt_commit('-Am','src-c1');", 0,0,0);
  sqlite3_open(b, &dst);
  sqlite3_exec(dst, "CREATE TABLE d(y); INSERT INTO d VALUES(2);"
                    "SELECT dolt_commit('-Am','dst-c1');", 0,0,0);
  if( destOnMissingBranch ){
    sqlite3_exec(dst, "SELECT dolt_branch('dstbranch');"
                      "SELECT dolt_checkout('dstbranch');", 0,0,0);
  }
  printf("pre: active=%s\n", q(dst, "SELECT active_branch()"));
  sqlite3_backup *bk = sqlite3_backup_init(dst, "main", src, "main");
  int rc = bk ? sqlite3_backup_step(bk, -1) : -1;
  int rc2 = bk ? sqlite3_backup_finish(bk) : -1;
  printf("step=%d finish=%d\n", rc, rc2);
  printf("post active=%s\n", q(dst, "SELECT active_branch()"));
  printf("post tables=%s\n", q(dst, "SELECT group_concat(name,',') FROM sqlite_master"));
  printf("post read s: %s\n", q(dst, "SELECT x FROM s"));
  printf("post read d: %s\n", q(dst, "SELECT y FROM d"));
  printf("post integrity: %s\n", q(dst, "PRAGMA integrity_check"));
  {
    char *zErr=0;
    int w = sqlite3_exec(dst, "INSERT INTO d VALUES(99)", 0,0,&zErr);
    printf("post write d rc=%d %s\n", w, zErr?zErr:"");
    sqlite3_free(zErr); zErr=0;
    w = sqlite3_exec(dst, "CREATE TABLE nt(z); INSERT INTO nt VALUES(1);", 0,0,&zErr);
    printf("post create rc=%d %s\n", w, zErr?zErr:"");
    sqlite3_free(zErr);
  }
  printf("post status: %s\n", q(dst, "SELECT group_concat(table_name||'/'||status,';') FROM dolt_status"));
  printf("recover checkout main: %s\n", q(dst, "SELECT dolt_checkout('main')"));
  printf("after checkout tables=%s s=%s\n",
      q(dst, "SELECT group_concat(name,',') FROM sqlite_master"), "");
  printf("after checkout read s: %s\n", q(dst, "SELECT x FROM s"));
  printf("after checkout integrity: %s\n", q(dst, "PRAGMA integrity_check"));
  sqlite3_close(src); sqlite3_close(dst);
  sqlite3 *r=0; sqlite3_open(b, &r);
  printf("reopen: active=%s tables=%s\n",
      q(r, "SELECT active_branch()"),
      q(r, "SELECT group_concat(name,',') FROM sqlite_master"));
  printf("reopen s=%s integrity=%s branches=%s\n",
      q(r, "SELECT x FROM s"), q(r, "PRAGMA integrity_check"),
      q(r, "SELECT group_concat(name,',') FROM dolt_branches"));
  sqlite3_close(r);
}

int main(int argc, char **argv){
  const char *dir = argc>1?argv[1]:".";
  kase(dir, 0);
  kase(dir, 1);
  return 0;
}
```

Observed output for case 1: `post read d: 2` (phantom), `post integrity: integrity check failed`, `post write d rc=5 database is locked`, `recover checkout main: (rc=1 failed to snapshot current branch state)`; reopen of the same file is healthy.

## Root cause (hypothesis)

`sqlite3_backup_step` (src/pager_shim.c, memory-dest adopt ~:1265–1272, file-dest `destCs->adoptReplacement = 1` ~:1366) replaces the dest store wholesale and calls `doltliteInvalidateBtreeWorkingState`, but nothing remaps the dest session's pinned branch when the adopted store lacks that ref (the fresh-open path falls back to the default branch). Subsequent refresh under the stale ref half-fails, leaving cached working state serving reads while writes/status/checkout error. Fix direction: on adoptReplacement, revalidate the session branch against the adopted store and fall back to its default branch, matching reopen behavior.

timsehn:
Fixes #2459.

## Problem

`sqlite3_backup_step`/`finish` into a destination connection whose session is checked out on a branch absent from the source image left the live destination connection pinned to a vanished branch. Reads could serve the pre-backup catalog, integrity checks failed, writes stayed locked, and only close/reopen recovered.

The original repair also allocated the fallback branch name after installing the replacement store. If that allocation failed, backup returned an error after leaving the destination file or memory store adopted and the live session inconsistent.

## Fix

Before changing the destination, backup now checks the locked source image and prepares any required fallback branch name. Allocation failure therefore leaves both file-backed and in-memory destinations untouched. Once adoption succeeds, installing the prepared branch is non-failing and the working state is invalidated. Sessions whose branch exists in the image and detached sessions remain unchanged.

On Windows, the destination chunk-store handle is closed and reopened around `MoveFileExW`; Windows otherwise refuses to replace the open destination file. This keeps the new file-backed regression active on Windows instead of excluding it.

## Testing

- `backup_dest_missing_branch`: 27/27, including deterministic file and memory allocation-failure cases
- `backup_safety`: 61/61
- `backup_source_write_busy`: 46/46
- Full native shell battery: 119/119 suites
- Full C regression battery: 311,164/311,164
- `make lint`

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Co-Authored-By: Codex <noreply@openai.com>


timsehn:
@itoqa Addressed both reported failures in 0c24dfd240. Backup now allocates the fallback branch name against the locked source store before either file or memory adoption, then installs it through a non-failing step only after adoption succeeds. Added fault-injection checks proving SQLITE_NOMEM preserves the old branch and content for both destination types. Local result: 27/27 targeted checks, 119/119 shell suites, and 311,164/311,164 C checks.

PR #2502, opened 2026-08-29
Refuse dbstat scans of DoltLite-format databases

### LINKED ISSUE #2500: dbstat silently returns an empty set on a populated database
## Severity

MEDIUM — silent wrong answer on a production-reachable surface. `SQLITE_ENABLE_DBSTAT_VTAB` is in the shipped build.

## Behavior

`SELECT count(*) FROM dbstat` returns **0** on a populated database with rc=SQLITE_OK — while `PRAGMA page_count` on the same connection reports a nonzero count. Stock SQLite returns real per-table page/cell/payload rows. A space-monitoring or analysis tool concludes the database is empty. Verified twice, deterministic.

## Repro

```sql
CREATE TABLE big(i INTEGER PRIMARY KEY, v TEXT);
WITH RECURSIVE c(x) AS (VALUES(1) UNION ALL SELECT x+1 FROM c WHERE x<2000)
INSERT INTO big SELECT x, printf('v%d',x) FROM c;
SELECT count(*) FROM dbstat;   -- doltlite: 0, stock: dozens of rows
PRAGMA page_count;             -- doltlite: nonzero (inconsistent with dbstat)
```

## Root cause

`src/dbstat.c` `statNext()` calls `sqlite3PagerPagecount()`, and the shim pager's `shimPagerPagecount` (`src/pager_shim.c`) hard-returns 0, so every btree walk EOFs before page 1.

Adjacent hazard worth noting in the fix: `shimPagerGet` returns `*ppPage=0` with SQLITE_OK — only the zero pagecount stands between dbstat and a null-deref, so any future pagecount change must also gate dbstat.

## Suggested fix

Make dbstat fail loud with a clear doltlite-specific error, exactly like its sibling `sqlite_dbpage` already does (`src/doltlite_dbpage.c` rejects pages >1 with "content-addressed chunk store has no page layout"-style messaging). Alternatively synthesize per-table stats from prolly chunk metadata, but the loud refusal is the beta-sized fix.

## Context

Found in the Aug 27 deep review, re-verified in the Aug 29 pre-beta review; listed as a pre-beta item ("SQL parity" dimension: the one behaving-badly surface with no issue on file).

timsehn:
`SELECT count(*) FROM dbstat` returned 0 with `SQLITE_OK` on a populated DoltLite-format database, while `PRAGMA page_count` was nonzero. Stock SQLite returns per-page b-tree rows. The shim pager reports pagecount 0, so the walk EOFed before page 1.

Fail the scan with a doltlite-specific error, the same approach as `sqlite_dbpage`, instead of synthesizing stats or pretending the database is empty:

```
doltlite: dbstat is not supported (content-addressed chunk store has no page layout)
```

Attached stock SQLite files and TEMP still walk pages. `PRAGMA table_info('dbstat')` still works.

Reclassify the `stat stat-*` testfixture gates that scanned pages from `intentional` to `unsupported issue=2500`. Keep `stat-7.2` / `stat-7.2.4` as intentional (DROP / cascade of an attached-schema dbstat vtab). Reclassify `statfault` the same way: the loud refusal shifts the no-fault (and nearby recovered) faultsim iterations, including section 2.

Fixes #2500


timsehn:
Automated fix: `sqlite-regression-core-sql` failed on four `stat` tests (`stat-0.2`, `stat-8.2`, `stat-8.3`, `stat-8.4`) that previously passed only because they expected an empty dbstat scan. They now error with the dbstat refusal, so they are gated as `unsupported issue=2500`.

timsehn:
CI failed on `sqlite-regression-fault-fuzz` after the loud dbstat refusal.

`statfault` used to fail only on section 1's no-fault page-count comparison (empty scan vs 8 pages). The scan now errors, so those fault indexes moved and section 2 fails too. Gated the 39 assertions as `unsupported issue=2500`. Local `testfixture`: `OK: statfault (359 tests, 39 known divergences)`.

PR #2531, opened 2026-09-01
Add doltlite-go: a database/sql driver with the engine vendored

timsehn:
Go bindings as a `database/sql` driver — `sql.Open("doltlite", path)` — over cgo, with the amalgamation vendored into the package so there's no system library to install.

## Why a Go binding, given embedded Dolt exists

I argued earlier in this repo's history that Go didn't need one, because [`github.com/dolthub/driver`](https://github.com/dolthub/driver) embeds Dolt in pure Go with no cgo. That reasoning holds for *new* programs wanting versioned SQL — but it doesn't cover the case that motivated this: **embedded Dolt speaks Dolt's on-disk format and cannot open a DoltLite database**, which is a single-file chunk store. Anything that has to read DoltLite files from Go — the DoltHub API, for instance — needs a real binding.

The README states this plainly and sends new Go programs to embedded Dolt instead, so the packages don't compete for the same use case.

## Shape

- Full driver set: `Driver`, `Conn`, `Stmt`, `Rows`, `Tx`, `Result` — so `database/sql` and everything layered on it works unchanged.
- Version control is plain SQL on the connection; no binding-specific API.
- Binds text and blobs with `SQLITE_TRANSIENT`, which cgo requires (C must not retain Go memory past the call), and reads text by length rather than as a C string so embedded NULs survive.
- Empty strings and empty blobs bind through a valid pointer rather than nil.

## Tests

`test-package.sh` stages the module and builds the smoke test against the **staged copy** via a `replace` directive, so a missing vendored amalgamation fails the test instead of shipping. 22 checks pass locally: every value type (UTF-8 text, blobs with embedded NULs, null, int64, real, empty string), `ErrNoRows`, column names, error paths, commit/rollback transactions, and a commit → branch → checkout → merge → `dolt_log` round trip. `gofmt` and `go vet` clean.

Build-Test runs it on ubuntu and macos. Windows is excluded and documented, same as the Rust crate: the build links the platform's zlib, which MSVC doesn't provide.

## Distribution needs a repo

Go modules resolve from a VCS path, so publishing needs a **`dolthub/doltlite-go`** repository (same pattern as `doltlite-php`) that release wiring pushes the staged module and a matching tag into. Could you create it when convenient? I'll wire `release-go` once it exists.

Until then this is usable directly with a `replace` directive pointing at `packaging/go`, which is what the DoltHub API side could do immediately if it's in a hurry.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -212,6 +212,14 @@ jobs:
         bash packaging/composer/test-package.sh build
       timeout-minutes: 5
 
+    - name: Smoke test the doltlite-go module
+      if: matrix.platform != 'windows'
+      shell: bash
+      run: |
+        chmod +x packaging/go/assemble.sh packaging/go/test-package.sh
+        bash packaging/go/test-package.sh build
+      timeout-minutes: 20
+
     - name: Smoke test the doltlite Rust crate
       if: matrix.platform != 'windows'
       shell: bash
--- a/.gitignore
+++ b/.gitignore
@@ -76,3 +76,5 @@ state.json
 .compat-cache/
 packaging/rust/vendor/
 packaging/rust/target/
+packaging/go/doltlite.c
+packaging/go/doltlite.h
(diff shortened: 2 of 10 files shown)


timsehn:
Closing: speculative. dolthub/doltlite-go already exists and covers the actual need — a pure-Go implementation of the HTTP sync protocol (prollyhash, remoteproto, litestore, remote, litehttp), deliberately without cgo, which is what hosting DoltLite repos calls for.

This PR was a different capability: executing SQL against a DoltLite file from Go through the C engine. Nothing needs that today, and the module path here collides with the existing repo's. Recoverable from the packaging/go-driver branch if a Go consumer ever needs to run queries rather than move chunks.

PR #2552, opened 2026-09-01
Add origin-backed lazy clones

eric-richardson1:
Depends on #2551 and uses its chunk-source seam as the engine-owned executor.

## Summary

- add a built-in source backed by the recorded `origin` remote, reusing the existing file and HTTP(S) remote clients
- add `dolt_clone('--lazy', url)` to install refs, origin, and tracking refs without transferring the reachable chunk graph
- enable origin-backed misses per open with `file:...?lazy_origin=1`, including bootstrap and query in one connection
- keep origin fetch and fast-forward pull refs-only while demand reads fault verified chunks into the local cache

## Phase 0 / design

- Activation is per-open configuration, not persisted state. `lazy_origin=1` is inert until an origin exists, so it activates immediately after `dolt_clone('--lazy', ...)` records origin. This makes no file-format, manifest, WAL, or version change.
- A host-registered callback takes precedence. Clearing it restores the built-in origin source for that connection.
- The source reuses the existing remote client and its 256-hash `xGetChunks` batches, credential resolution, and TLS behavior. Version 1 opens a client per callback; file remotes are opened read-only.
- Source callbacks inherit the seam's graph-lock rule, common hash verification, bounded memory cache, and verified write-through path. No network request runs under the graph lock.
- On an origin-enabled connection, `dolt_fetch('origin')` updates tracking refs without bulk transfer. Fast-forward pull remains lazy. A divergent pull fails with a materialization requirement because merge would otherwise need source reads while graph-locked.
- Without `lazy_origin=1`, cached chunks remain readable and an uncached miss reports `NOTFOUND` with the requested hash and activation remedy. With activation, unreachable-origin transport reports a hash-named chunk-source I/O error rather than corruption.

## Validation

- `./build/chunk_source_test`: 311/311
- file remote suite: 133/133
- HTTP remotesrv suite: 42/42 across 156 validated requests
- regression assertions: 311177/311177
- amalgamation enabled/disabled probes
- fresh `DOLTLITE_ENABLE_CHUNK_SOURCE=0` build
- lint and native DoltLite suites

## Deferred follow-ups

- durable lazy activation with an explicit format/version gate before an auto-enabled end-user CLI mode
- full materialization command
- remote client and HTTP connection pooling
- richer prefetch tuning and WASM host plumbing

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,8 +137,8 @@ an Emscripten-compatible socket transport or proxy; see
 [`examples/wa-sqlite-clone.mjs`](examples/wa-sqlite-clone.mjs) for a public
 clone request that exercises the client.
 
-`DOLTLITE_ENABLE_CHUNK_SOURCE=0` omits host-provided lazy chunk fetching. The
-feature is enabled by default.
+`DOLTLITE_ENABLE_CHUNK_SOURCE=0` omits host-provided and origin-backed lazy
+chunk fetching. The feature is enabled by default.
 
 ## Using as a C Library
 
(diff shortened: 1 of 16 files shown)


eric-richardson1:
Addressed in the chunk-source seam on #2551 and rebased here. Cache-writer graph-lock acquisition now honors the SQLite busy handler and uses a bounded default retry when no handler is configured; fetches still happen outside the lock, and persistence still refreshes/rechecks before committing.

A synchronized child-process regression fails on the old one-shot lock path and passes with the fix. It also reopens with the source unavailable and verifies the full scan is served locally with zero callbacks.

Focused validation: chunk_source_test 303/303, file remotes 133/133, HTTP remotes 42/42, amalgamation enabled/disabled probes, and lint.

eric-richardson1:
Ito QA concurrency failure is fixed in the callback seam dependency (`51bf2f1c69`). Source cache-writer lock acquisition now honors SQLite's busy handler and uses a bounded fallback retry when no handler is installed; it still refuses source fetches while the calling connection already holds a graph lock. Verified bytes are persisted before cache publication, so an overlapping scan waits through transient contention and the same data remains readable after reopening with the origin offline.

The deterministic fork/pipe regression fails with the old one-shot lock path and passes with the retry. After rebasing this PR: `chunk_source_test` 311/311, file remote 133/133, HTTP remotesrv 42/42, lint PASS, amalgamation enabled/disabled PASS.

PR #2558, opened 2026-09-02
Keep the Homebrew formula on the current release tarball

### LINKED ISSUE #2542: Homebrew formula pins v0.10.6 (pre-format-freeze) with no release automation
Found in the 2026-09-01 post-beta deep review at 5824327f91.

`packaging/homebrew/doltlite.rb:7-8` hardcodes:

```ruby
url "https://github.com/dolthub/doltlite/releases/download/v0.10.6/doltlite-autoconf-0.10.6.tar.gz"
sha256 "816ecedc369dd61fd06a0759985d5fbfbdf9fadcd1096cc1c35cc7f09447e7c3"
```

Latest tag is v0.50.2 — the formula is ~40 versions stale, and it predates the beta format freeze, so anyone installing from it gets a pre-beta engine writing a pre-freeze on-disk format. Every other distribution channel (deb, npm, PyPI, RubyGems, Packagist, NuGet, crates.io, split repos) is automated in `release.yml`; Homebrew is the only one with no bump job.

Fix: add a `release-homebrew` step that rewrites url/sha256 on tag (same pattern as the doltlite-python bump), or remove the formula until it can be kept current. The formula's own test block asserts `dolt_version()` matches the formula version, so a bump job is self-validating.

timsehn:
Fixes #2542.

`packaging/homebrew/doltlite.rb` pinned `v0.10.6`, which predates the beta format freeze. Homebrew-core will not accept the formula until notability, but this in-repo copy is the source of truth and was shipping a stale engine to anyone who used it.

The formula now points at the `v0.50.2` autoconf tarball. `packaging/homebrew/bump-formula.sh` rewrites `url`/`sha256`. A `release-homebrew` job runs after the GitHub release has the tarball, hashes it, and pushes the bump to `master`.

`test/homebrew_formula_test.sh` rejects a `v0.10.6` pin, checks url/tarball versions agree, and exercises the bump script.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -96,12 +96,17 @@ jobs:
         tar czf doltlite-autoconf-${VERSION}.tar.gz \
           --transform "s,^,doltlite-autoconf-${VERSION}/," \
           configure Makefile.in Makefile.linux-generic Makefile.msc main.mk \
+          doltlite.mk \
           auto.def VERSION manifest manifest.tags manifest.uuid \
           pragma.h magic.txt lempar.c \
           sqlite3.pc.in sqlite.pc.in \
           src/ ext/ tool/ autosetup/ autoconf/ \
           sqlite3.c sqlite3.h sqlite3ext.h sqlite3session.h \
           .dolt_release_version
+        prefix="doltlite-autoconf-${VERSION}"
+        for req in configure main.mk doltlite.mk auto.def sqlite3.c sqlite3.h; do
+          tar tzf doltlite-autoconf-${VERSION}.tar.gz | grep -qx "${prefix}/${req}"
+        done
         rm .dolt_release_version
 
     - name: Package wasm distribution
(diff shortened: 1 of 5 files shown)


timsehn:
Addressed the Ito QA failures in a4634d27f1:

- **Install (high):** `main.mk` includes `doltlite.mk`, but the autoconf tarball omitted it, so `make doltlite` stopped immediately. `doltlite.mk` is now in the tarball file list, and the source job greps the archive for it (plus configure/main.mk/auto.def/sqlite3.c/h) after packing. The v0.50.2 GitHub asset is still missing that file; the next tagged release is the one that will install from the formula.
- **Duplicate URL (medium):** `re.subn(..., count=1)` treated "one replacement" as success. The helper now counts matches first and refuses to write unless there is exactly one url and one sha256.
- **Malformed version (medium):** versions must be `X.Y.Z`; path-like input exits 2 and leaves the formula unchanged.
- **Concurrent bump (medium):** `release-homebrew` uses a concurrency group, skips if master already has this version or a newer one, and retries fetch/reset/push.

`test/homebrew_formula_test.sh` covers the bump rejects and the tarball file-list grep.

Issue #2570, opened 2026-09-02
WASM: sqlite3_js_db_export poisons the next dolt_commit with a spurious commit conflict


timsehn:
Found while reproducing #2563. @thrudhame worked around it with a
retry-the-commit-once rule and documented it as a "commit poison" trigger map;
it is a real defect and a regression since 0.11.53.

In the WASM build, calling `sqlite3_js_db_export()` on a handle makes that same
handle's *next* `dolt_commit` fail with

```
commit conflict: another connection committed to this branch. Please retry your transaction.
```

There is no other connection. The commit after that succeeds, so the effect is
one-shot per export, and the failed commit leaves the working set intact.

## Repro

```js
import fs from 'node:fs'; import path from 'node:path'; import { pathToFileURL } from 'node:url';
const d = process.env.JSWASM || path.resolve('ext/wasm/jswasm');
const e = path.resolve(d,'sqlite3-node.mjs'), w = path.resolve(d,'sqlite3.wasm');
{ const t = fs.readFileSync(e,'utf8'); if(t.includes("__dirname + '/'")) fs.writeFileSync(e, t.replaceAll("__dirname + '/'","new URL('.', import.meta.url).href")); }
const { default: init } = await import(pathToFileURL(e).href);
const sqlite3 = await init({ locateFile:(n)=>n==='sqlite3.wasm'?w:n, wasmBinary: fs.readFileSync(w) });
const { oo1, capi } = sqlite3;
const log = (s)=>fs.writeSync(1,s+'\n');
const commit = (db,m)=>{ try{ db.selectValue(`select dolt_commit('-Am','${m}')`); return 'ok'; }
  catch(ex){ return ex.message.includes('commit conflict') ? 'COMMIT CONFLICT' : 'ERR '+ex.message; } };

const db = new oo1.DB('/p.db','c');
db.exec(`select dolt_config('user.name','P'); select dolt_config('user.email','p@x');
         create table t(id integer primary key)`);
log(`commit 1 (no export yet):      ${commit(db,'c1')}`);
db.exec(`insert into t values(1)`);
log(`commit 2 (still no export):    ${commit(db,'c2')}`);
capi.sqlite3_js_db_export(db.pointer);
db.exec(`insert into t values(2)`);
log(`commit 3 (right after export): ${commit(db,'c3')}`);
db.exec(`insert into t values(3)`);
log(`commit 4 (no export since c3): ${commit(db,'c4')}`);
[...]

Issue #2569, opened 2026-09-02
WASM: http(s):// remotes can never connect (POSIX sockets under Emscripten)


timsehn:
Reported by @thrudhame in #2563 (finding 3, their WB1). Reproduced here against
a local `doltlite-remotesrv`.

The HTTP remote client is compiled into the WASM build
(`DOLTLITE_ENABLE_REMOTES=1` is the default in `ext/wasm/GNUmakefile`), and
`dolt_clone`/`dolt_fetch`/`dolt_push` are all present in
`pragma_function_list`. No `http(s)://` remote operation can complete, in Node
or in a browser, because the transport is raw POSIX sockets.

## Repro

Start a native relay and push a database to it:

```
build/doltlite-remotesrv -p 8399 /tmp/remotes &
build/doltlite /tmp/native.db "
  select dolt_config('user.name','N'); select dolt_config('user.email','n@x');
  create table t(id integer primary key, v text);
  insert into t values(1,'one'),(2,'two');
  select dolt_commit('-Am','c1');
  select dolt_remote('add','origin','http://127.0.0.1:8399/room.db');
  select dolt_push('origin','main');"
```

Then, from the WASM build:

```js
import fs from 'node:fs'; import path from 'node:path'; import { pathToFileURL } from 'node:url';
const d = process.env.JSWASM || path.resolve('ext/wasm/jswasm');
const e = path.resolve(d,'sqlite3-node.mjs'), w = path.resolve(d,'sqlite3.wasm');
{ const t = fs.readFileSync(e,'utf8'); if(t.includes("__dirname + '/'")) fs.writeFileSync(e, t.replaceAll("__dirname + '/'","new URL('.', import.meta.url).href")); }
const { default: init } = await import(pathToFileURL(e).href);
const sqlite3 = await init({ locateFile:(n)=>n==='sqlite3.wasm'?w:n, wasmBinary: fs.readFileSync(w) });
const { oo1 } = sqlite3;
const log = (s)=>fs.writeSync(1,s+'\n');
const url = 'http://127.0.0.1:8399/room.db';

const c = new oo1.DB('/clone.db','c');
try { log(`clone -> ${c.selectValue(`select dolt_clone(?)`, [url])}`); }
catch(ex){ log(`clone: ${ex.message}`); }

const b = new oo1.DB('/local.db','c');
b.exec(`select dolt_config('user.name','B'); select dolt_config('user.email','b@x')`);
[...]

Issue #2568, opened 2026-09-02
WASM: INSERT spins forever after ~29 export/reopen cycles


timsehn:
Reported by @thrudhame in #2563 (finding 2, their PB3/PB4). Reproduced,
narrowed to a minimal loop, and confirmed deterministic.

In the WASM build, a loop of *commit → export → reopen the exported image*
spins forever at a fixed iteration. The spin is inside an ordinary `INSERT`,
at 100% CPU, and the statement never returns, so no JS timer can fire and the
page or process is wedged. The native engine runs the same sequence cleanly.

## Repro

```js
import fs from 'node:fs'; import path from 'node:path'; import { pathToFileURL } from 'node:url';
const d = process.env.JSWASM || path.resolve('ext/wasm/jswasm');
const e = path.resolve(d,'sqlite3-node.mjs'), w = path.resolve(d,'sqlite3.wasm');
{ const t = fs.readFileSync(e,'utf8'); if(t.includes("__dirname + '/'")) fs.writeFileSync(e, t.replaceAll("__dirname + '/'","new URL('.', import.meta.url).href")); }
const { default: init } = await import(pathToFileURL(e).href);
const sqlite3 = await init({ locateFile:(n)=>n==='sqlite3.wasm'?w:n, wasmBinary: fs.readFileSync(w) });
const { oo1, capi } = sqlite3;
const log = (s)=>fs.writeSync(1,s+'\n');
const open = (p) => { const db = new oo1.DB(p,'c');
  db.exec(`select dolt_config('user.name','P'); select dolt_config('user.email','p@x')`); return db; };

let db = open('/h0.db');
db.exec(`create table t(id integer primary key)`);
db.selectValue(`select dolt_commit('-Am','genesis')`);

for (let i = 1; i <= 60; i++) {
  log(`iter ${i}: insert`);   db.exec(`insert into t values(${i})`);
  log(`iter ${i}: commit`);
  try { db.selectValue(`select dolt_commit('-Am','c${i}')`); }
  catch (ex) { db.selectValue(`select dolt_commit('-Am','c${i}')`); }   // see #2570
  log(`iter ${i}: export`);   const bytes = capi.sqlite3_js_db_export(db.pointer);
  db.close();
  capi.sqlite3_js_vfs_create_file('unix', `/h${i}.db`, bytes, bytes.byteLength);
  db = open(`/h${i}.db`);
  log(`ITER ${i}: ok rows=${db.selectValue('select count(*) from t')}`);
[...]

Issue #2563, opened 2026-09-02
[RFC] Matrix rooms in the browser on doltlite-wasm


thrudhame:
Hey @timsehn , continuing dolthub/doltgresql#2099 here, since this part is DoltLite territory now.

_(I also left an update there back in August with a Dolt homeserver prototype and a gaps list, but the issue was already closed, so I suspect nobody saw it: comment (https://github.com/dolthub/doltgresql/issues/2099#issuecomment-5390017737))_

TL;DR: rooms as databases, events as commits. And I saw you have doltlite now, so I couldn't resist to play with my long standing pipe dream: no server matrix server.

Example of two browsers chatting over WebRTC:
https://thrudhame.github.io/communico/lite/

- open in two different browsers, same room;
- try sending from both at once to get fork indicator,
- next message heals it

Code: https://github.com/thrudhame/communico/tree/spike/communico-lite
Also check [lite/RESULTS.md](https://github.com/thrudhame/communico/blob/spike/communico-lite/lite/RESULTS.md) for capability matrix and errors.

Good news first:
- the explicit-txn merge driver, deterministic convergence (equal dolt_hashof_table across different arrival orders), AS OF room state, branching at an interior commit, it all works in WASM.

Bad news: peer sync itself can't be dolt-native yet (tested in 0.11.53):

1. Remotes crash at the wasm ABI level.
2. Byte-shipping (export, vfs_create_file, fresh handle) freezes after a few adopt, commit, export generations..
3. The amalgamation build has the remote client compiled in, but a single-threaded runtime can't complete a round trip with it..

Questions:
- Are remotes in doltlite-wasm on the roadmap? Even file:// alone would unlock OPFS-backed peers.
- Is the deep-lineage hang (2) known? I can reduce it further if useful, the repro is already scripted.
- And the Doltgres-side gaps from the August comment (https://github.com/dolthub/doltgresql/issues/2099#issuecomment-5390017737) are still alive for the server-side variant, especially N-parent commits (Matrix allows up to 20 prev_events) — curious which of those are principled and which are just "nobody asked yet".

Thank you.

timsehn:
The remote surface is the least tested. I would take repros for each of the three issues you found.

> Are remotes in doltlite-wasm on the roadmap? Even file:// alone would unlock OPFS-backed peers.

Yes absolutely. I need a set of test cases and to make this work.

> Is the deep-lineage hang (2) known?

No known. A repro would be helpful

[...]

PR #2573, opened 2026-09-02
Move the store scan buffers off the stack

### LINKED ISSUE #2567: WASM: dolt_clone/dolt_fetch on a file:// remote traps at the wasm ABI level
Reported by @thrudhame in #2563 (finding 1). Reproduced and narrowed here.

In the WASM build, an operation that opens a `file://` remote traps at the
WebAssembly ABI level instead of running. The same SQL works in the native CLI.

## Repro

Save as `repro-file-remote.mjs`, run from the repo root after
`make -C ext/wasm npm` (or point `JSWASM` at an unpacked
`@dolthub/doltlite-wasm` package):

```js
import fs from 'node:fs'; import path from 'node:path'; import { pathToFileURL } from 'node:url';
const d = process.env.JSWASM || path.resolve('ext/wasm/jswasm');
const e = path.resolve(d,'sqlite3-node.mjs'), w = path.resolve(d,'sqlite3.wasm');
{ const t = fs.readFileSync(e,'utf8'); if(t.includes("__dirname + '/'")) fs.writeFileSync(e, t.replaceAll("__dirname + '/'","new URL('.', import.meta.url).href")); }
const { default: init } = await import(pathToFileURL(e).href);
const sqlite3 = await init({ locateFile:(n)=>n==='sqlite3.wasm'?w:n, wasmBinary: fs.readFileSync(w) });
const { oo1 } = sqlite3;
const log = (s)=>fs.writeSync(1,s+'\n');
const t = (label, fn) => { try { log(`${label}: ok -> ${JSON.stringify(fn())}`); } catch(ex){ log(`${label}: ${ex.message}`); } };

log(`version=${new oo1.DB(':memory:').selectValue('select dolt_version()')}`);
const a = new oo1.DB('/A.db','c');
a.exec(`select dolt_config('user.name','A'); select dolt_config('user.email','a@x');
        create table t(id integer primary key, v text); insert into t values(1,'one');`);
a.selectValue(`select dolt_commit('-Am','c1')`);
a.close();

const b = new oo1.DB('/B.db','c');
b.exec(`select dolt_config('user.name','B'); select dolt_config('user.email','b@x')`);
t('remote add',  ()=>b.selectValue(`select dolt_remote('add','peerA','file:///A.db')`));
t('fetch 1-arg', ()=>b.selectValue(`select dolt_fetch('peerA')`));
t('fetch 2-arg', ()=>b.selectValue(`select dolt_fetch('peerA','main')`));
b.close();

const c = new oo1.DB('/C.db','c');
t('clone fresh', ()=>c.selectValue(`select dolt_clone('file:///A.db')`));
c.close();
```

## Observed

| build | remote add | fetch | clone |
|---|---|---|---|
| `master` @ 6ff8549c8e, emsdk 5.0.7 | ok | ok | **null function or function signature mismatch** |
| `master` @ 6ff8549c8e, emsdk 3.1.74 (CI toolchain) | ok | ok | **null function or function signature mismatch** |
| npm `@dolthub/doltlite-wasm@0.50.2` | ok | ok | **null function or function signature mismatch** |
| npm `@dolthub/doltlite-wasm@0.11.53` | ok | **null function or function signature mismatch** / **table index is out of bounds** | ok |

The clone trap does not need a preceding fetch. Cloning alone, in a fresh
process, traps the same way on every build tested.

Both messages are Emscripten indirect-call faults (a call through the function
table with no entry or a mismatched signature), not SQL errors. SQL parsing,
URL validation and the remote-catalog write all succeed; the trap fires where
the remote store would be opened or transferred.

## Native contrast

The identical sequence works end to end in the native CLI, including under
ASan and UBSan, both with the default VFS and with `unix-none` (the VFS the
WASM build defaults to):

```
select dolt_remote('add','peerA','file:///tmp/w/a.db');  -- 0
select dolt_fetch('peerA');                              -- 0
select message from dolt_log('peerA/main');              -- c1 on A ...
select dolt_clone('file:///tmp/w/a.db');                 -- 0, rows arrive
```

## Notes

Which op traps moves between releases (0.11.53 traps on fetch and clones fine;
0.50.2 and current master clone-trap and fetch fine), so this is not one fixed
call site. No WASM test in `ext/wasm/test-doltlite-*.mjs` touches remotes, and
every remote test is a native shell test, so nothing in CI would have caught
this.

timsehn:
Opening a chunk store reserved more stack than a WebAssembly build has.

`csScanForCommittedRoot`, `csWalResolveDamage`, `csReplayWalFrom`,
`csFindLastCheckpointRoot` and `csTryLoadWalCheckpoint` each held a 64K read
window as a local. Measured with `-Wframe-larger-than` under `emcc -Oz`:

| function | frame before | after |
|---|---|---|
| `chunkStoreOpen` (scan inlined) | 65,728 | under 4,096 |
| `csTryLoadWalCheckpoint` | 68,080 | under 4,096 |
| `csReplayWalFrom` | 68,352 | 4,992 |
| `csWalResolveDamage` | 65,712 | under 4,096 |

Emscripten's stack is 65,536 bytes, so `chunkStoreOpen` alone did not fit.
Every wasm database open ran off the end of the stack into the static data
below it, and nothing reported it: the release build carries no stack check,
so the writes just landed on whatever globals the linker had placed there.

## Why it surfaced as an ABI trap

Remote operations nest a second store open under an already-deep VDBE stack,
which pushes the overflow further. The global it clobbered was a function
pointer, so `dolt_clone` and `dolt_fetch` died at an indirect call with
`null function or function signature mismatch` or `table index is out of
bounds`. Which message, and which operation, moved with the link layout:
0.11.53 trapped in fetch, 0.50.2 and master trap in clone. That drift is what
made this look like several unrelated bugs.

A development build with stack checks enabled names it directly:

```
Aborted(stack overflow (Attempt to set SP to 0x000a3d80, with stack limits
  [0x000a5560 - 0x000b5560]))
    at csTryLoadWalCheckpoint
    at chunkStoreOpen
    at doltliteFsRemoteOpen
    at openRemoteByUrl
```

## Scope

The same corruption accounts for two other reported wasm failures, both of
which clear with this change:

- **#2568** — a loop of commit, export and reopen spun forever inside a plain
  `INSERT` at a fixed iteration. Now 60/60 cycles clean.
- **#2570** — `sqlite3_js_db_export()` left the next `dolt_commit` on that
  handle failing as a spurious peer conflict. Now clean.

**#2569** (http remotes cannot connect) is a separate root cause, the POSIX
socket transport under Emscripten, and is untouched here.

## The change

The windows move to the heap at the same sizes with the same scan geometry, so
reads and the overlap arithmetic are byte-for-byte what they were. Native
behavior is unchanged. Each new allocation has a NOMEM path, and the WAL
replay buffer is allocated once per call rather than per record.

## Test

`ext/wasm/test-doltlite-remote.mjs` covers exactly the path that was broken: a
`file://` fetch with a remote-ref log check, a clone with content verified, and
a push read back through a second clone. It fails on the published 0.50.2
runtime with `null function or function signature mismatch` and passes on this
branch. Wired into the wasm CI job.

No wasm test touched remotes before this, which is why the bug reached a
release. Every remote test in the tree is a native shell test.

## Validation

Run locally on macOS arm64:

| suite | result |
|---|---|
| `test/run_doltlite_tests.sh` | 126/126 suites, 172/172 guards |
| `test/run_c_tests.sh` | 31/31 |
| recovery/corruption/WAL tcl suites | 1187 tests, 0 failures |
| OOM and fault-injection tcl suites | 771 tests, 0 failures |
| `make lint` | pass |
| amalgamation native + wasm compile | pass |
| wasm smoke, write-e2e, export-roundtrip | pass |

The wasm repros were re-run against a release `make -C ext/wasm npm` build with
`wasm-strip`, not just the development build.

Reported by @thrudhame in #2563.

Fixes #2567.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -31,6 +31,10 @@ jobs:
       run: |
         node ext/wasm/test-doltlite-export-roundtrip.mjs
 
+    - name: Test DoltLite wasm file:// remotes
+      run: |
+        node ext/wasm/test-doltlite-remote.mjs
+
     - name: Smoke test the @dolthub/doltlite-wasm npm package
       run: |
         chmod +x packaging/npm/assemble.sh packaging/npm/test-package.sh
--- a/ext/wasm/test-doltlite-remote.mjs
+++ b/ext/wasm/test-doltlite-remote.mjs
@@ -1,0 +1,120 @@
+import fs from 'node:fs';
+import path from 'node:path';
+import { pathToFileURL } from 'node:url';
+
+const nodeEntry = path.resolve('ext/wasm/jswasm/sqlite3-node.mjs');
+const wasmPath = path.resolve('ext/wasm/jswasm/sqlite3.wasm');
+let nodeEntryText = fs.readFileSync(nodeEntry, 'utf8');
+if(nodeEntryText.includes("__dirname + '/'")){
+  nodeEntryText = nodeEntryText.replaceAll(
+    "__dirname + '/'",
+    "new URL('.', import.meta.url).href"
+  );
+  fs.writeFileSync(nodeEntry, nodeEntryText);
+}
+const { default: sqlite3InitModule } = await import(pathToFileURL(nodeEntry).href);
+
+const sqlite3 = await sqlite3InitModule({
+  locateFile: (name)=> name === 'sqlite3.wasm' ? wasmPath : name,
+  wasmBinary: fs.readFileSync(wasmPath)
+});
+
+function fail(msg){
+  throw new Error(msg);
+}
+
+/* Remote ops nest a second chunk store open under an already-deep VDBE
+ * stack. That is the deepest path the engine runs in a browser, and the
+ * one that overflowed the 64K wasm stack while the release build had no
+ * check to report it. */
+function seedSource(){
+  const db = new sqlite3.oo1.DB('/remote-source.db', 'c');
+  try {
+    db.exec("select dolt_config('user.name','wasm')");
+    db.exec("select dolt_config('user.email','wasm@example.com')");
+    db.exec("create table t(id integer primary key, v text)");
+    db.exec("insert into t values(1,'alpha')");
+    db.selectValue("select dolt_commit('-Am','c1')");
+    db.exec("insert into t values(2,'beta')");
+    db.selectValue("select dolt_commit('-Am','c2')");
+  } finally {
+    db.close();
+  }
+}
+
+function checkFetch(){
+  const db = new sqlite3.oo1.DB('/remote-fetch.db', 'c');
+  try {
+    db.exec("select dolt_config('user.name','wasm')");
+    db.exec("select dolt_config('user.email','wasm@example.com')");
+    const added = db.selectValue(
+      "select dolt_remote('add','peer','file:///remote-source.db')");
+    if(added !== 0) fail(`dolt_remote add returned ${added}`);
+
+    const fetched = db.selectValue("select dolt_fetch('peer')");
+    if(fetched !== 0) fail(`dolt_fetch returned ${fetched}`);
+
+    const messages = db.selectObjects("select message from dolt_log('peer/main')")
+          .map((r)=>r.message);
+    if(!messages.includes('c2')) fail(`fetched log missing c2: ${messages.join(',')}`);
+
+    console.log(`fetch: refs=peer/main log=${messages.length} entries`);
+  } finally {
+    db.close();
+  }
+}
+
+function checkClone(){
+  const db = new sqlite3.oo1.DB('/remote-clone.db', 'c');
+  try {
+    const cloned = db.selectValue("select dolt_clone('file:///remote-source.db')");
+    if(cloned !== 0) fail(`dolt_clone returned ${cloned}`);
+
+    const count = db.selectValue("select count(*) from t");
+    const value = db.selectValue("select v from t where id = 2");
+    if(count !== 2) fail(`cloned db has ${count} rows, expected 2`);
+    if(value !== 'beta') fail(`cloned db has ${value}, expected beta`);
+
+    console.log(`clone: rows=${count}`);
+  } finally {
+    db.close();
+  }
+}
+
+function checkPush(){
+  const db = new sqlite3.oo1.DB('/remote-push.db', 'c');
+  try {
+    db.exec("select dolt_config('user.name','wasm')");
+    db.exec("select dolt_config('user.email','wasm@example.com')");
+    db.exec("create table u(id integer primary key)");
+    db.exec("insert into u values(1)");
+    db.selectValue("select dolt_commit('-Am','push me')");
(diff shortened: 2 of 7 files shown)


timsehn:
Headroom check, since the obvious follow-up question is whether 64K is enough
once the buffers are gone rather than whether it merely stopped trapping.

Run against a **development** wasm build, which keeps Emscripten's stack check
and aborts on any overflow rather than corrupting silently:

| exercise | result |
|---|---|
| clone a 20,000-row, 11-commit store over `file://` | rc=0, 20,000 rows, 119 ms |
| reopen a store with a 300-commit WAL | rc=0, 301 commits |
| clone that deep-WAL store | rc=0, 300 rows |
| `test-doltlite-remote.mjs` | pass |

No abort on any of them, so the remaining depth fits with margin. The largest
frames left on this path are `doltliteSyncChunks` at 12,832 bytes and
`csReplayWalFrom` at 4,992.

I did not raise `-sSTACK_SIZE`. It would be defense in depth, but it changes the
shipped artifact's memory profile and the measurements above say it is not
needed to fix this. Easy to add later if a deeper path turns up.

timsehn:
`Test Suite / sqlite-regression-fault-fuzz` went red on the first push. It is
not a regression from this change, and the second commit addresses it.

`attachmalloc-1` fails three transient-OOM assertions on **master** as well. I
built master's testfixture in a worktree and ran the same test against both
binaries rather than assume:

| binary | failing assertions |
|---|---|
| master @ 29b416e75b | `attachmalloc-1.transient.508`, `.936`, `.1433` |
| this branch | `attachmalloc-1.transient.510`, `.937`, `.1433` |

Same count, same assertions, same swallowed OOM behind them (`expected [1 1]
got [0 {}]`, tracked as #2492). Moving the scan buffers to the heap adds
allocations ahead of the fault point, so two of the three indices shift by one
or two. CI reported it as unexpected only because the divergence entries pinned
exact indices.

Those indices were never the content of the entries. They already needed
separate `@coverage` and `@no-coverage` spellings, because a coverage build
shifts the fault point the same way, and the comment above them records four
earlier shifts from registrations that allocate at connection open. Any engine
change that allocates differently moves them again.

So the entries are now a counted gate, the form the file documents for this and
that `statfault` already uses for its OOM assertions:

```
attachmalloc attachmalloc-1.transient.*{3} class=unsupported issue=2492
```

Three failures under that prefix stay gated; a fourth, or a fix that leaves
two, still fails the run. Verified the gate accepts master unchanged, and all
six divergence-file guards pass (`check_testfixture_exception_inventory.sh`,
`check_testfixture_crash_coverage.sh`, their `_test` counterparts,
`run_testfixture_test.sh`, `check_testfixture_inventory_issues_alive_test.sh`).

Ratchet lowered as that inventory check instructs: `gates` 4739 to 4737 and
`unsupported` 1311 to 1309, since five pinned lines, only three of which fire
in any one run, become one gate of three.

The swallowed OOM itself is untouched and stays with #2492.

PR #2575, opened 2026-09-02
Rebuild WASM library after source changes

### LINKED ISSUE #2574: WASM build relinks stale libdoltlite.a when src/ changes
Found while validating #2567: editing anything under `src/` does not rebuild the
WASM library, so a wasm build can silently link stale DoltLite objects.

`ext/wasm/GNUmakefile` builds `libdoltlite.a` with:

```make
$(dir.doltlite.build)/Makefile: $(dir.top)/Makefile
	@cp $(dir.top)/Makefile $(dir.doltlite.build)/Makefile
	@find $(dir.top) -maxdepth 1 \( -name '*.h' -o -name '*.c' \) -exec cp {} $(dir.doltlite.build) \;

$(doltlite.wasm.lib): $(dir.doltlite.build)/Makefile
	@$(MAKE) -C $(dir.doltlite.build) libdoltlite.a ...
```

The library's only prerequisite is the copied top-level `Makefile`. Nothing
under `src/` is a prerequisite, so once `libdoltlite.a` exists, `make -C
ext/wasm ...` never re-enters the sub-make no matter what changed in the engine.

## Repro

```
make -C ext/wasm jswasm/sqlite3-node.mjs     # builds libdoltlite.a
# edit any src/*.c, e.g. add a printf to a function you can reach
make -C ext/wasm jswasm/sqlite3-node.mjs     # relinks, but with the OLD objects
```

The rebuilt runtime does not contain the edit. Deleting
`ext/wasm/bld/doltlite-build/libdoltlite.a` first is the current workaround.

## Why it matters

This cost real time on #2567: two rounds of "the fix didn't work" were actually
a stale library. It is worse than a slow edit-test loop, because a green wasm
test run can attest to code that is not in the build. CI happens to be safe
today only because `make -C ext/wasm npm` forces a clean rebuild and CI starts
from a clean checkout.

A related smaller trap in the same area: the node target writes
`jswasm/node/sqlite3-node.wasm`, while `jswasm/sqlite3.wasm` comes from the
vanilla target. A harness that pairs `jswasm/sqlite3-node.mjs` with
`jswasm/sqlite3.wasm` (as the in-tree tests do) can load a binary from a
different build than the entry point it is testing.

## Fix

Make `$(doltlite.wasm.lib)` depend on the DoltLite sources, or drop the
prerequisite so the sub-make always runs and does its own up-to-date check.
The second is simpler and cheap, since the sub-make is a no-op when nothing
changed.

timsehn:
Make the outer WASM build always invoke the nested DoltLite make dependency graph, which remains a no-op when its inputs are unchanged. Also make the published Node entry point depend on its canonical `sqlite3.wasm` partner so an incremental build cannot pair outputs from different builds.

Add a CI regression that touches a DoltLite C source and verifies the object, archive, and Node entry point's WASM binary all rebuild.

Validation:
- fail-before reproduction left `prolly_xxhash.o` and `libdoltlite.a` unchanged
- incremental rebuild regression passes
- WASM smoke, write E2E, and native export roundtrip tests pass

Fixes #2574

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -829,6 +829,7 @@ jobs:
           make doltlite
           cd ..
           make -C ext/wasm fiddle npm
+          bash test/wasm_incremental_build_test.sh
           mkdir -p ci-wasm-runtime
           cp -a ext/wasm/jswasm/. ci-wasm-runtime/
           make -C ext/wasm dist
--- a/ext/wasm/GNUmakefile
+++ b/ext/wasm/GNUmakefile
@@ -1228,12 +1228,13 @@ sqlite3-wasm.c.in += $(doltlite.wasm.lib)
 endif
 
 ifeq (1,$(DOLTLITE_WASM))
-$(dir.doltlite.build)/Makefile: $(dir.top)/Makefile
+.PHONY: doltlite.wasm.sync
+doltlite.wasm.sync:
 	@mkdir -p $(dir.doltlite.build)
-	@cp $(dir.top)/Makefile $(dir.doltlite.build)/Makefile
-	@find $(dir.top) -maxdepth 1 \( -name '*.h' -o -name '*.c' \) -exec cp {} $(dir.doltlite.build) \;
+	@cp -f -p $(dir.top)/Makefile $(dir.doltlite.build)/Makefile
+	@find $(dir.top) -maxdepth 1 \( -name '*.h' -o -name '*.c' \) -exec cp -f -p {} $(dir.doltlite.build) \;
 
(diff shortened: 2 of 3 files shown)


timsehn:
Good catch. The nested build reads `src/*.h` through its absolute `TOP`, but generated top-level inputs such as `sqlite3.h` were copied only when the root Makefile changed. Updated the forced step to synchronize the root Makefile and top-level C/header inputs on every nested build while preserving timestamps, and extended the incremental regression to verify the copied-header path before checking source/object/archive/WASM rebuilding. Local incremental and Node runtime tests pass.

Issue #2637, opened 2026-09-04
Make `dolt_hashof_index()` and migrate the version control fuzzer to use it


timsehn:
We added foreign keys, indexes, partial indexes, and shadow tables to the version control fuzzer. This function will be useful for invariant testing.

timsehn:
Context from building the fuzzer coverage in #2639, plus the scope Tim set in review.

## Why this is needed: index roots are invisible today

While extending the fuzzer I went looking for a cheap invariant that would catch a corrupt index, and there isn't one. **No existing surface can see an index root.** Measured on the #2632 shape, a merge that files rows into a partial index they do not belong in:

| probe | corrupt index | correct index |
|---|---|---|
| `dolt_status` after the merge | empty | empty |
| `dolt_status` after `REINDEX` | empty | empty |
| `dolt_hashof_db()` before / after `REINDEX` | unchanged | unchanged |
| `dolt_hashof_table('t')` before / after `REINDEX` | unchanged | unchanged |
| `PRAGMA integrity_check` | `ok` | `ok` |

`REINDEX` rebuilds the index and the store is content-addressed, so a stale entry means the rebuilt root differs from the stored one. Nothing reports that. The only things that noticed #2632 at all were the foreign-key and constraint walks, and only because they iterate every index entry: a range scan skips the stale rows, since their key holds the NULL the predicate excludes.

That is why the new fuzz coverage has to lean on `foreign_key_check`, `dolt_constraint_violations`, index-versus-scan comparison and fts5's own integrity command. All indirect, all needing a constraint to exist before corruption becomes observable.

## Scope

Two parts, beyond the function itself:

1. **`dolt_hashof_index()`** — the direct instrument. The fuzzer then asserts the obvious invariant: rebuild the index, hash it, and the hash must be unchanged. That is one query per index instead of the indirect battery above, and it works whether or not the table has a constraint.

2. **Index roots must be hashed into `dolt_hashof_db()`, and index changes must show in `dolt_status` as changes to the parent table.** An index is part of the database's content, so a database whose indexes differ is a different database and must not report an identical hash. In status it belongs on the table it indexes, not as a row of its own, so a rebuilt or diverging index reads as the table being modified.

Point 2 is the part with teeth: it turns every existing test that compares `dolt_hashof_db()` or reads `dolt_status` into an index-corruption detector, including the ones already in the fuzzer.

[...]

Issue #2632, opened 2026-09-04
Merge corrupts partial indexes and reports false foreign-key violations for NULL values


Chris7:
## Summary

  A three-way merge incorrectly adds rows with a `NULL` foreign-key value to a partial index defined with `WHERE parent_item_id IS NOT NULL`.

  The merge then reports nonexistent foreign-key violations and rolls back.

  ## Reproduction

  Run the attached `repro.sql`:

  ```console
  /tmp/doltlite :memory: < repro.sql
```

  Using DoltLite v0.11.50 produces:

  Error near line 38: Committing this transaction resulted in a working set with constraint violations, transaction rolled back.

  ## Expected behavior

  The merge succeeds. All merged parent_item_id values are NULL, which is valid for the nullable foreign key and excluded from the non-NULL partial index.

  ## Actual behavior

  The merged partial index contains entries for incoming rows whose parent_item_id is NULL. Constraint verification consequently reports false self-referential foreign-key violations.

  Rebuilding the index with REINDEX removes the entries, after which dolt_verify_constraints() reports no violations.

[repro.sql](https://github.com/user-attachments/files/31845089/repro.sql)

PR #2639, opened 2026-09-04
Fuzz foreign keys, indexes and shadow tables under version control

timsehn:
The stateful version-control fuzzer modelled a single plain table, so every shape whose index maintenance runs outside the VDBE went unexercised: a foreign key, a plain index, partial indexes, and a virtual table's shadow tables. #2632 lived in exactly that gap.

> **Merge order:** this depends on #2635. Until that fix lands, CI here reports #2632, which is the point.

## What the fuzzer now builds

The seed repository gains, alongside `kv`:

- `parent`, and `child` with a **nullable foreign key** to it;
- a **plain index** on `child(grp)`;
- a **unique partial index over each arm** of the predicate, `WHERE parent_id IS NOT NULL` and `WHERE parent_id IS NULL`;
- an **fts5 virtual table**, for its shadow tables.

A new `mutate_related` operation writes those shapes: rows arrive with and without a parent, **flip between the two**, and are deleted; documents come and go from the fts5 table. Schema mutations can now create a partial index as well as a plain one, and `reindex` joins the operation set.

## Why the checks can be absolute

Every branch writes a disjoint id and `grp` range, and only ever points a foreign key at a parent seeded in the very first commit. No merge, cherry-pick, rebase or revert can therefore produce a *legitimate* constraint violation, so any violation the engine reports is a bug and the harness can fail hard rather than tolerate. Concretely, on every step:

- an index-driven read and a `NOT INDEXED` scan must return the same ids, for both arms of the partial predicate and for the plain index;
- `pragma_foreign_key_check` and `dolt_constraint_violations` must be empty;
- a duplicate must still be refused through each partial index, whatever merges and rebases did to it;

and periodically:

- `REINDEX` must not change a single answer;
- fts5's own `integrity-check` must pass, and its match count must equal its content count.

`kv` stays modelled row by row. The new tables are deliberately checked by self-consistency instead, because that is what catches an index holding rows its predicate excludes. Worth noting for future work: neither `dolt_status`, `dolt_hashof_db` nor `dolt_hashof_table` can see index roots, so a corrupt index is invisible to them. Only a walk of the index shows it, which is why the constraint checks carry the weight here. #2637 would give this a much sharper instrument.

## Evidence it works

Against a **checked build** (`DOLTLITE_PROLLY_CHECK=1`, as CI builds) of master, which lacks the #2632 fix:

| seed | result |
|---|---|
| 4, 5, 6 | FAIL at `cherry_pick` / `merge` |
| 11, 12 | FAIL at `cherry_pick`, step 298 and 363 |

each with the reported symptom:

```
Error near line 1: Committing this transaction resulted in a working set with
constraint violations, transaction rolled back.
```

With #2635 applied, the same seeds run clean: 661, 617 and 683 steps, all 34 operations exercised. Throughput is unchanged at roughly five steps per second, so the fixed-duration CI job covers the same ground it did before.

### DIFF
--- a/test/vc_stateful_fuzzer.py
+++ b/test/vc_stateful_fuzzer.py
@@ -245,6 +245,193 @@ def check_invariants(doltlite, db_path, branches, tags, model, rng):
             raise AssertionError("integrity_check on %s returned %r" % (branch, out))
     if rng.randrange(4) == 0:
         assert_refs(doltlite, db_path, branches, tags)
+    assert_related_consistent(doltlite, db_path, branch)
+    if rng.randrange(4) == 0:
+        assert_shadow_tables_consistent(doltlite, db_path, branch)
+    if rng.randrange(8) == 0:
+        assert_reindex_preserves_answers(doltlite, db_path, branch)
+
+
+def branch_base(branch):
+    return (0 if branch == "main" else int(branch[1:])) * 10000
+
+
+def mutate_related(doltlite, db_path, branch, rng, step):
+    """Write the foreign key, index and shadow-table shapes on one branch."""
+    base = branch_base(branch)
+    grp = base + rng.randrange(4)
+    key = base + 100 + rng.randrange(60)
+    action = rng.choice(
+        ("insert_null", "insert_parent", "flip_to_parent", "flip_to_null",
+         "delete_child", "duplicate", "doc_insert", "doc_delete")
+    )
+    parent = rng.randrange(1, SEED_PARENTS + 1)
+    text = "row %s %05d doc" % (branch, step)
+
+    if action == "insert_null":
+        sql = (
+            "INSERT INTO child(id, parent_id, grp, name, body) "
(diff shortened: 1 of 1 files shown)


timsehn:
CI's `Smoke / vc-fuzz` failure is the expected one: it is #2632, which #2635 fixes, not a new bug and not a harness bug.

CI ran seed `1788549176` and failed at step 567 on `cherry_pick` with

```
Error near line 1: Committing this transaction resulted in a working set with
constraint violations, transaction rolled back.
```

Replaying that exact seed locally against two checked builds (`DOLTLITE_PROLLY_CHECK=1`, as CI builds):

| build | result |
|---|---|
| master, without #2635 | **FAIL at step 567, op `cherry_pick`** — same step CI reported |
| master + #2635 | **OK, 1139 steps**, all operations exercised |

Same seed, same harness, same step number. The failure follows the engine fix, not the fuzzer.

Merge #2635 first and this goes green.

PR #2646, opened 2026-09-05
Fix staged index rootpage collisions in incremental dolt_add

### LINKED ISSUE #2644: Per-table dolt_add corrupts committed autoindex rootpages; checkout fails for 12 indexed tables
Staging newly created indexed tables individually produces a commit that fails after checkout. The same fixture passes if staged with one `dolt_add()` call.

Reproduced with the published `@dolthub/doltlite@0.50.5` package on macOS arm64, Node 24.10.0. Also reproduced against our pinned 0.11.37 dependency. Twelve tables are sufficient; the fixture has no application dependencies, foreign keys, remotes, or ignored runtime tables.

Steps:

```sh
npm init -y
npm install @dolthub/doltlite@0.50.5
node repro.cjs
STAGE_ALL=1 node repro.cjs
```

`repro.cjs`:

```js
const assert = require('node:assert/strict');
const { mkdtempSync, rmSync } = require('node:fs');
const { tmpdir } = require('node:os');
const { join } = require('node:path');
const { DatabaseSync } = require('@dolthub/doltlite');
const root = mkdtempSync(join(tmpdir(), 'doltlite-stage-index-'));
const db = new DatabaseSync(join(root, 'repro.db'));
const tables = Array.from({ length: 12 }, (_, i) => `items_${i}`);
try {
  for (const table of tables) {
    db.exec(`CREATE TABLE ${table} (id TEXT PRIMARY KEY, label TEXT UNIQUE);
      INSERT INTO ${table} VALUES ('base', 'unique-label');`);
  }
  if (process.env.STAGE_ALL === '1') db.doltAdd();
  else for (const table of tables) db.doltAdd(table);
  db.prepare("SELECT dolt_commit('-m', 'initial', '--author', 'Test <test@example.test>')").get();
  for (const table of tables) {
    assert.equal(db.prepare(`SELECT id FROM ${table} WHERE label='unique-label'`).get()?.id, 'base');
  }
  db.doltBranch('feature');
  db.doltCheckout('feature');
  for (const table of tables) {
    assert.equal(db.prepare(`SELECT id FROM ${table} WHERE label='unique-label'`).get()?.id, 'base');
  }
  console.log('PASS');
} finally {
  db.close();
  rmSync(root, { recursive: true, force: true });
}

```

Expected: both runs preserve all rows and their UNIQUE indexes after checking out a branch at the same commit.

Actual: per-table staging passes the indexed lookups before checkout, then the first lookup after checkout fails:

```text
Error: prepare: malformed database schema (sqlite_autoindex_items_4_2) - invalid rootpage
```

The `STAGE_ALL=1` run prints `PASS`.

In a larger application catalog, the same staging pattern also leads to unreadable pushed file remotes and cloned catalogs, and checkout can drop ignored runtime tables. Our earlier diagnosis blamed checkout/clone; this reduction points to construction of the staged/committed indexed schema. I have not assumed those additional symptoms have separate root causes.

timsehn:
## What happened

Rootpage numbers are local to each catalog. Each named add serializes a compact staged subset, then the next add mixes those renumbered entries with the full working catalog. Tables were reconciled by name, but unnamed index entries retained staged numbers that could collide with unrelated working entries. The live working schema remained readable; checkout exposed the corrupt committed schema.

Move the prior staged entries and their schema rootpages into a disjoint temporary range before overlaying working entries. Apply the same rule to commit -a and preserve its staged master until schema composition. Final serialization remains canonical: all-at-once, forward, and reverse staging produce identical database hashes. This prevents new corruption; it does not repair already-corrupt commits.

## Regression coverage

- 4, 12, and 24 newly created indexed tables; forward, reverse, and stage-all controls.
- Forced index lookups, checkout/reopen, unique failures, failed-add state, and file push/clone.
- Mixed commit -am preserves explicitly staged index contents and excludes unstaged tables.
- Dolt comparisons for indexed reads and duplicate/missing-table errors.

Unfixed master: 37 native assertions and four Dolt comparisons fail. Fixed: all pass. Existing index staging coverage started from an already-committed schema and missed incremental construction of new indexed catalogs.

## Validation

- Full native runner: 126 suites passed.
- Expanded index/VC matrix: 152 checks passed.
- Commit Dolt oracle: 71 cases passed.
- Rebuilt testfixture: nine SQLITE_TEST staging configurations pass checkout, reopen, indexed reads, and integrity checks.
- make -C build lint and git diff --check passed.

Separate finding during validation: hard reset while retaining untracked indexed tables can construct an invalid working schema. This remains outside the staging fix; the committed snapshot in that reproduction is readable on a fresh branch.

Fixes #2644

### DIFF
--- a/src/doltlite_add.c
+++ b/src/doltlite_add.c
@@ -718,6 +718,11 @@ int doltliteStageNamedTables(
     Pgno iTable = 0;
     int j;
     if( !zTable || strcmp(zTable, ".")==0 ) continue;
+    for(j=0; j<i; j++){
+      const char *zPrior = (const char*)sqlite3_value_text(argv[j]);
+      if( zPrior && sqlite3_stricmp(zPrior, zTable)==0 ) break;
+    }
+    if( j<i ) continue;
 
     if( !bForce ){
       int ignored = 0;
--- a/src/doltlite_add.c
+++ b/src/doltlite_add.c
@@ -694,6 +694,18 @@ int doltliteStageNamedTables(
       rc = loadSchemaFromCatalog(db, cs, doltliteGetCache(db), &stagedSrc,
                                  &aStagedSchema, &nStagedSchema);
     }
+    if( rc==SQLITE_OK ){
+      Pgno offset;
+      rc = doltliteDisjoinCatalogEntries(db, aWorking, nWorking,
+                                         aStaged, nStaged, &offset);
+      if( rc==SQLITE_OK ){
+        for(i=0; i<nStagedSchema; i++){
+          if( aStagedSchema[i].iRootpage>1 ){
+            aStagedSchema[i].iRootpage += offset;
+          }
+        }
+      }
+    }
     if( rc!=SQLITE_OK ){
       ADDNAMED_FREE_ALL();
       sqlite3_result_error(context, "failed to load schema for staging", -1);
--- a/src/doltlite_commit_cmd.c
+++ b/src/doltlite_commit_cmd.c
@@ -189,6 +189,7 @@ static int doltliteCommitStageModifiedOnly(sqlite3 *db, sqlite3_context *context
   u8 *aRemoveStaged = 0;
   const char **azTouched = 0;
   int nTouched = 0;
+  Pgno stagedOffset = 0;
 
   memset(&workingIdx, 0, sizeof(workingIdx));
   memset(&headIdx, 0, sizeof(headIdx));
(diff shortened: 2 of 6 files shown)


timsehn:
@itoqa I could not reproduce the reported abort with the ordinary, indexed, FTS5, or R-tree cases, either optimized or assertion-enabled. Dolt also accepts duplicate and case-variant table names as a successful no-op.

The code did redundantly rebuild the same staged object, so `fc05f61` now deduplicates those arguments before staging. New native and Dolt-oracle cases cover `dolt_add('t','T','t')`, virtual-table shadows, commit, checkout, reopen, forced index reads, and integrity checks. Please rerun and share the exact SQL or harness assertion if the abort remains.

PR #2664, opened 2026-09-06
Preserve ignored local tables during native merges

justintanner:
A clean branch with a committed `dolt_ignore` rule and an untracked ignored
table currently refuses `dolt_merge` with an uncommitted-changes error. Adding
an index to that table also makes `dolt_status` incorrectly report it modified.

This change compares tracked state for merge eligibility and preserves ignored
table/index data in the live working catalog. The staged and committed result
excludes those objects. Installing the complete working catalog before schema
actions and index rebuilding also preserves ignored data across allocation
failures. Fast-forward, three-way, no-commit, squash, conflict rollback, and
explicit abort use the same separation. Tracked and staged changes still block,
and incoming object-name collisions refuse to overwrite local objects.

The status index pass now honors the parent table's ignore rule. A small
schema-loader error-path cleanup frees four decoded strings when appending a
schema entry fails, covering the leak found by the new allocation-failure test.

Validation on base `37a390eb7b021962d9d287a465a2da3c9f59c3cf`, macOS arm64:

- Focused testfixture regression: 4,156 checks, zero errors or memory leaks.
- All 126 native suites and all 33 C suites pass.
- New Dolt oracle: 10/10 pass; unpatched base fails all ten.
- Existing Dolt ignore and merge oracles: 50/50 and 104/104 pass using Dolt 2.3.1.
- Native lint and clean stock-SQLite parity pass.
- Both new suites are registered in their CI buckets; oracle inventory checks pass.
- A full Videobook catalog passes merge, conflict rollback, reopen, and object
  contents checks across 56 tables, 23 ignored runtime tables, and 107 schema
  objects. Opening with published 0.50.6, merging with this patch in a separate
  process, then reopening with 0.50.6 also passes. The probe explicitly checks
  ignored `sqlite_sequence` across these transitions: after deleting job ID 2,
  allocations advance to 3 with the patch and 4 after the published binding reopens.

Ignored FTS5 virtual/shadow tables and implicit `sqlite_sequence` handling for
ignored auto-increment tables remain follow-up scope from review. Videobook has
no virtual tables and explicitly ignores its runtime-only sequence state.
Shared sequence state for tracked and ignored tables is unverified.

There is no public API or file-format version change.

Hosted CI for the submitted commit: [upstream run](https://github.com/dolthub/doltlite/actions/runs/34043486050) and [fork run](https://github.com/justintanner/doltlite/actions/runs/34043532585) each passed all 69 jobs.

Full application reproduction: https://github.com/justintanner/videobook-engine/blob/e6d74ab1e0a9f4b639cada1c764e21d03972133d/scripts/native-full-catalog-merge-probe.mjs

### DIFF
--- a/src/doltlite_internal.h
+++ b/src/doltlite_internal.h
@@ -1023,6 +1023,7 @@ int mergeAbortInPlace(sqlite3 *db);
 int mergeFastForward(
   sqlite3 *db, sqlite3_context *context, ChunkStore *cs,
   const ProllyHash *pOurHead, const ProllyHash *pTheirHead,
+  const ProllyHash *pIgnored,
   int squash
 );
 int doltliteMergeRef(
--- a/src/doltlite_merge_cmd.c
+++ b/src/doltlite_merge_cmd.c
@@ -18,13 +18,224 @@
 #include <ctype.h>
 #include <time.h>
 
+static int mergeStoreCatalog(
+  sqlite3 *db,
+  struct TableEntry *aCatalog,
+  int nCatalog,
+  SchemaEntry *aSchema,
+  int nSchema,
+  ProllyHash *pHash
+){
+  u8 *pData = 0;
+  int nData = 0;
+  int rc = doltliteSerializeCatalogEntriesForeignDomain(
+      db, aCatalog, nCatalog, aSchema, nSchema, &pData, &nData);
+  if( rc==SQLITE_OK ){
+    rc = chunkStorePut(doltliteGetChunkStore(db), pData, nData, pHash);
+  }
+  sqlite3_free(pData);
+  return rc;
+}
+
+static int mergeSplitWorkingCatalog(
+  sqlite3 *db,
+  const ProllyHash *pBase,
+  int bIgnoreListed,
+  ProllyHash *pTracked,
+  ProllyHash *pIgnored,
+  char **pzErr
+){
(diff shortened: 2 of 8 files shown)


timsehn:
Thanks for the PR. I'm reviewing this now. 

I think it rips ignore tables out of the main catalog. Not sure that's the right dimension to fix. Agree this is a bug and I'll fix the exact symptom.

timsehn:
I reproduced the original bug and agree this PR should land once CI is green. The ordinary ignored table/index path is real: the unpatched engine reports a clean `dolt_status` and then refuses `dolt_merge()` as dirty. The new native and Dolt-oracle coverage addresses that path well.

Follow-up work found during review:

- An ignored FTS5 virtual table plus its shadow tables still leaves `dolt_status` clean but causes `dolt_merge()` to fail with `uncommitted changes`. The catalog split operates on physical table entries, while the rootpage-0 virtual-table schema declaration remains in the tracked half; the restore filter also excludes rootpage-0 schema entries.
- An ignored `AUTOINCREMENT` table creates `sqlite_sequence`, which is reported as an untracked table and still blocks merge. Handling this needs care because `sqlite_sequence` may contain state for both tracked and ignored tables.

A foreign-key table with an explicit index passed an additional merge/reopen integrity probe. I consider the FTS/shadow and AUTOINCREMENT cases follow-up scope rather than blockers for this PR.

PR #2674, opened 2026-09-06
Land SeekLE/SeekGT prefix seeks on the last match in O(log n)

### LINKED ISSUE #2668: SeekLE/SeekGT prefix positioning walks every matching index entry (max()/ORDER BY DESC LIMIT 1 is O(group size))
Found in the 2026-09-06 deep review at master ca30b4a0f2.

The prolly prefix-match fix in `OP_SeekLE`/`OP_SeekGT` (`src/vdbe.c:5423`) handles "IndexMoveto landed at the first prefix match" by calling `sqlite3BtreeNext` until it leaves the matching range. That is O(number of matching entries) for what the planner costs as a point seek.

```
CREATE TABLE t(id INTEGER PRIMARY KEY, a INT, b INT);
WITH RECURSIVE s(x) AS (SELECT 1 UNION ALL SELECT x+1 FROM s WHERE x<400000) INSERT INTO t(a,b) SELECT 1, x FROM s;
CREATE INDEX iab ON t(a,b);
.timer on
SELECT max(b) FROM t WHERE a=1;                        -- stock 0.04 ms, doltlite 22 ms
SELECT b FROM t WHERE a=1 ORDER BY b DESC LIMIT 1;     -- stock 0.01 ms, doltlite 22 ms
SELECT min(b) FROM t WHERE a=1;                        -- both ~0.03 ms
```

600x at 400k rows; linear in group size. `SELECT b FROM t WHERE a<=1 ORDER BY a DESC, b DESC LIMIT 1` hits the same path.

Fix direction: implement the reverse prefix seek in the prolly `xIndexMoveto` (seek to prefix + a max-suffix sentinel, or a "last key with this prefix" descent) so the VDBE loop is not needed.

timsehn:
Fixes #2668.

`SELECT max(b) FROM t WHERE a=1` / `ORDER BY b DESC LIMIT 1` under an index `(a,b)` walked every entry of the `a=1` group: the prolly `IndexMoveto` landed on the first prefix match and the `vdbe.c` workaround stepped forward through the run. 400k-row group: 22 ms vs 0.04 ms stock.

**Change**
- `prollyBtCursorIndexMoveto` takes a new path for prefix seeks with `default_rc<0`: position at the predecessor of the prefix's successor, merged across tree and pending map (raw lower bound on both, `mergeScan`, one `Previous`). Lands on the last match with `eqSeen`, or on the last entry below the key, or reports "everything is above".
- An exact numeric prefix is also a byte prefix of the 18-byte inexact neighbours (`2^53+1`-style values) that sort right after its matches; when the landing is one of those the bound moves to `prefix+0x80`, which every neighbour starts with and no field start does.
- `blobPrefixSuccessor` moves from the count path into `sortkey.c` as `sortKeyPrefixSuccessor`; `mergeScan` is exported from `prolly_btree_cursor.c`.
- The `vdbe.c` walk is kept as the fallback for the unsupported-collation scan path (comment updated); for a correct landing it costs one step out and one back.

**Result:** `max(b)` on the 400k-row group goes from 22 ms to 0.09 ms locally.

**Tests**
- New `test/doltlite_index_reverse_prefix_seek.sh` (132 checks): indexed plan vs forced table scan for max/min/DESC LIMIT/`>`/`<=` prefix seeks over ASC and DESC second columns, TEXT/REAL/NULL/NOCASE prefixes, missing prefixes, `2^53±k` exact/inexact neighbours, composite clustered PKs, and a pending-map matrix (in-transaction inserts, deletes, moves, pending-only tables). Plus a timing check: 20 reverse seeks over a 400k-row prefix must finish under 0.1 s (0.47 s before, well under after). On master 131/132 pass and only the timing check fails; all pass with the fix.
- Local: 22 seek/index-related shell suites incl. index_prefix, txn_seek_prefix_contract, txn_seek_visibility, nocase_index, count_*, index_row_delta, edge_cases, advanced, feature_deep, savepoint, commit, merge; testfixture over the where*/index*/minmax*/orderby*/skipscan*/collate*/rowvalue*/in*/analyze*/autoindex*/descidx*/unique*/like*/seekscan1 suites (15,892 passing, 0 unexpected); `sql_differential_test.sh` seeds 1–150 vs stock clean; `make lint`; amalgamation compiles. sqllogictest runners aren't built locally, so that gate is left to CI.

### DIFF
--- a/src/prolly_btree_cursor.c
+++ b/src/prolly_btree_cursor.c
@@ -144,7 +144,7 @@ static int mergeCompare(BtCursor *pCur, ProllyMutMapEntry *e){
   }
 }
 
-static int mergeScan(BtCursor *pCur, int dir, int *pRes){
+int mergeScan(BtCursor *pCur, int dir, int *pRes){
   assert( pCur!=0 );
   assert( pCur->pMutMap!=0 );
   assert( dir==1 || dir==-1 );
--- a/src/prolly_btree_cursor_count.c
+++ b/src/prolly_btree_cursor_count.c
@@ -187,25 +187,6 @@ int sortKeyFromUnpackedForCount(
   return rc;
 }
 
-/* Exclusive bound after prefix pKey. Drop trailing 0xff or the bound is too high. */
-static int blobPrefixSuccessor(const u8 *pKey, int nKey, u8 **ppOut, int *pnOut){
-  int i;
-  u8 *pOut;
-  for(i=nKey-1; i>=0 && pKey[i]==0xff; i--){}
-  if( i<0 ){
-    *ppOut = 0;
-    *pnOut = 0;
-    return SQLITE_OK;
-  }
-  pOut = sqlite3_malloc(i+1);
-  if( !pOut ) return SQLITE_NOMEM;
-  memcpy(pOut, pKey, i+1);
-  pOut[i]++;
-  *ppOut = pOut;
-  *pnOut = i+1;
-  return SQLITE_OK;
-}
-
 /* Exclusive end after a complete first field. Exact 9-byte numerics prefix
 ** an 18-byte neighbor; a raw successor would count that neighbor in. */
 static int blobFieldSuccessor(const u8 *pKey, int nKey, u8 **ppOut, int *pnOut){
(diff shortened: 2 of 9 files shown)


timsehn:
CI red here is the pre-existing rowid OOM swallow that fts3cov pins at fixed fault indices; this PR shifts the geometry by one so the pins stop matching. Fix is #2679 (must merge first); with it cherry-picked onto this branch fts3cov is clean. Will rebase once #2679 lands.

PR #2677, opened 2026-09-06
Fix preserve-rowids dumps for clustered primary keys

### LINKED ISSUE #2669: .dump --preserve-rowids emits unrestorable INSERTs for every clustered-PK table
Found in the 2026-09-06 deep review at master ca30b4a0f2.

The shell's preserve-rowid decision (`src/shell.c.in:4051`) treats any table without an INTEGER PRIMARY KEY alias and with a `pk`-origin index as a rowid table. Clustered-PK tables read `rowid` but refuse to write it, so the dump emits `INSERT INTO t(rowid,...)` rows that fail on restore. Without `-bail` the shell continues and every row of every PK table is silently lost.

```
doltlite src.db "CREATE TABLE t(k TEXT PRIMARY KEY, v); INSERT INTO t VALUES('a',1),('b',2); CREATE TABLE i(id INT PRIMARY KEY, v); INSERT INTO i VALUES(5,'x'); CREATE TABLE n(v); INSERT INTO n VALUES('plain');"
doltlite src.db ".dump --preserve-rowids" > dump.sql
grep INSERT dump.sql
# INSERT INTO i(rowid,id,v) VALUES(5,5,'x');
# INSERT INTO n(rowid,v) VALUES(1,'plain');
# INSERT INTO t(rowid,k,v) VALUES(6964150017480843733,'a',1);
# INSERT INTO t(rowid,k,v) VALUES(6021798749535948405,'b',2);
doltlite dst.db < dump.sql
# Parse error near line 4: table i has no column named rowid
# Parse error near line 8: table t has no column named rowid
# Parse error near line 9: table t has no column named rowid
doltlite dst.db "SELECT count(*) FROM t; SELECT count(*) FROM i; SELECT count(*) FROM n;"   # 0 / 0 / 1
```

Expected: clustered-PK tables are dumped without a rowid column (same as WITHOUT ROWID tables), and the dump restores completely.

timsehn:
## Summary

- omit read-only compatibility rowids when dumping clustered-primary-key tables
- retain preserved rowids for keyless and UNIQUE-only tables
- verify `-bail` restores TEXT, INT, composite, and DESC primary-key dumps completely
- leave `DOLTLITE_PROLLY=0` shell behavior unchanged

Fixes #2669.

## Testing

- `bash test/doltlite_pk_clustered_rowid_readonly.sh` (19/19; two new checks fail before the fix)
- `bash test/sqlite_compatibility_contract_test.sh build/doltlite` (13/13)
- `bash test/oracle_dot_commands_test.sh build/doltlite build-stock/sqlite3` (61/61)
- `make -C build lint`
- clean stock `DOLTLITE_PROLLY=0` shell build and preserve-rowid probe


timsehn:
The initial `Smoke / vc-fuzz` failure was unrelated to this PR. Seed `1788723123` deterministically reaches an FTS5 integrity failure after rebase at step 454 on both this commit and the untouched parent commit `f57b44807f`.

I reran the randomized smoke job with a fresh seed; it passed the full 10-minute budget. The CI Gate and every required check are now green. No test or guard was changed.

PR #2679, opened 2026-09-06
Surface allocation failure when deriving a clustered-PK rowid

timsehn:
`doltliteSyntheticRowidFromRecord` / `doltliteSyntheticRowidFromSortKey` fell back to hashing the raw bytes on *any* codec error, folding an out-of-memory into the same path as an unencodable key. Under a malloc fault the statement succeeded with a rowid that does not belong to the row: `last_insert_rowid()` and `NEW.rowid` disagreed with `SELECT rowid`.

This is the swallow the fts3cov divergence list has been pinning as `intentional` at build-specific fault indices (`fts3cov-8.transient.573/576/722/725`, `fts3cov-10.*.46/287/292`, tagged `@coverage`/`@no-coverage`). #2674's seek change shifts the allocation geometry by one, so those pins stop matching and the same swallow shows up as five unexpected failures. Filed separately from #2674 per one-fix-per-PR; #2674 needs this merged first.

**Change**
- Both helpers return an rc with the rowid as an out parameter. The raw-bytes fallback stays for keys the codec cannot express (`SQLITE_NOTFOUND`); `SQLITE_NOMEM` propagates.
- `OP_Rowid` (NEW.rowid path) and `OP_Insert` (last_insert_rowid / update hook) abort the statement on it; `sqlite3BtreeSqlRowid` faults the cursor the same way its `currentMutMapEntry` failure already does.
- The nine fts3cov divergence entries are retired (the runner reports them stale once the swallow is gone); `intentional` and `gates` in the ratchet drop by nine.

**Tests**
- New `test/doltlite_recover_synthetic_rowid_oom.test` (fault-fuzz bucket): `oom-transient` sweep over an INSERT into a TEXT-PK table with an AFTER INSERT trigger recording `NEW.rowid`, asserting `last_insert_rowid()` and the logged `NEW.rowid` equal `SELECT rowid`, plus a cursor-rowid consistency sweep. On master iterations 195 and 196 return `1 0` with rc=0; with the fix 0 errors out of 493.
- `fts3cov` on this branch: 7859 passing, 0 known divergences, 0 unexpected, 0 stale. Cherry-picked onto #2674 it clears that PR's five unexpected failures as well.
- Local: rowid_identity, rowid_returning, pk_clustered_rowid_readonly, update_replace_trigger, edge_cases, index_prefix; testfixture fault-fuzz OOM suites (908 passing); `make lint` (exception inventory matches the lowered ratchet), crash-coverage checker, amalgamation compile.


timsehn:
fault-fuzz red was eight more `@coverage` pins (fkey_malloc, vtab_err) that this fix clears; the bucket had no unexpected failures. Retired them and lowered the ratchet by eight. Note: on macOS no-coverage builds, vtab_err/fkey_malloc already show shifted-index failures on master (14 on master, 6 with this fix) — pre-existing local geometry, not exercised by CI.

PR #2722, opened 2026-09-08
Start the CLI in DEFENSIVE mode like sqlite3

### LINKED ISSUE #2712: CLI should enable DEFENSIVE by default like sqlite3
## Summary

The doltlite CLI leaves `SQLITE_DBCONFIG_DEFENSIVE` off by default. Upstream sqlite3 turns it on for every connection unless `--unsafe-testing` is given. The override lives in `open_db()` in `src/shell.c.in`:

```
/* Upstream shell turns DEFENSIVE on unless --unsafe-testing. Under
** defensive mode, PRAGMA writable_schema=ON is a silent no-op ...
** Keep defensive off by default; --safe and `.dbconfig defensive on`
** still enable it, and --unsafe-testing still forces it off. */
```

Decision: match SQLite. Defensive on by default; `--unsafe-testing` turns it off, as upstream.

## What breaks today

Upstream `shell9-1.2.2` (runs against the doltlite CLI in CI from PR #2710 on) `.read`s a dump that writes `sqlite_master` and expects

```
Parse error near line 5 of testdump.txt: table sqlite_master may not be modified
```

The doltlite CLI applies the write. Pinned as engine-gap against this issue in `test/known_testfixture_divergences.txt`; retire the pin with the fix.

## Notes

The original motivation was that `PRAGMA writable_schema=ON` looks broken under defensive mode (it reports 0). That is upstream behaviour too, and `.dbconfig defensive off` or `--unsafe-testing` are the supported ways to get there.

timsehn:
Fixes #2712.

The shell kept `SQLITE_DBCONFIG_DEFENSIVE` off by default so `PRAGMA writable_schema=ON` would work without `.dbconfig defensive off`. Upstream turns defensive on for every connection unless `--unsafe-testing`. The doltlite branch is deleted and the upstream line is used; `--unsafe-testing`, `--safe` and `.dbconfig defensive` behave as in sqlite3.

**What changes for users.** Direct writes to `sqlite_master` and to virtual-table shadow tables (`x_model`, `f_config`, …) are rejected with `table … may not be modified` unless defensive is turned off, exactly as in sqlite3. `PRAGMA writable_schema=ON` reads back 0 under the default.

**Tests.**
- `shell9-1.2.2` (a `.read` of a dump that writes `sqlite_master`) now matches upstream; pin retired. `shell2-1.4.9` also passes on master since #2708 merged after its pin landed; retired here too (the three open PRs carry the same retirement, so expect a trivial ratchet conflict for whichever lands second). Ratchet: gates 4562 → 4560, engine-gap 3 → 1. Shell bucket 398 passing / 3 known.
- Three suites relied on the old default and now flip defensive off before their pokes, which is how the same poke is written for sqlite3: `doltlite_writable_schema.sh` (also gains a case asserting the pragma stays off under the default), `doltlite_vacuum.sh` (`vacuum_rejects_poked_schema`), `doltlite_vec1_vc.sh` (`vec1_missing_model_stays_loud` deletes a row from the `t_model` shadow table). Stock rejects both kinds of write under defensive, verified.
- Full `run_doltlite_tests.sh` battery: all suites pass with the new CLI. `sql_oracle_test.sh` vs the stock CLI 569/569. `make lint` clean.

### DIFF
--- a/src/shell.c.in
+++ b/src/shell.c.in
@@ -5381,21 +5381,7 @@ static void open_db(ShellState *p, int openFlags){
     {
       int testmode_on = ShellHasFlag(p,SHFLG_TestingMode);
       sqlite3_db_config(p->db, SQLITE_DBCONFIG_TRUSTED_SCHEMA, testmode_on,0);
-#ifdef DOLTLITE_PROLLY
-      /* Upstream shell turns DEFENSIVE on unless --unsafe-testing. Under
-      ** defensive mode, PRAGMA writable_schema=ON is a silent no-op (the
-      ** WriteSchema bit cannot be set), so the CLI always reports 0 and
-      ** looks broken next to stock sqlite3 CLIs that leave defensive off.
-      ** Keep defensive off by default; --safe and `.dbconfig defensive on`
-      ** still enable it, and --unsafe-testing still forces it off. */
-      if( testmode_on ){
-        sqlite3_db_config(p->db, SQLITE_DBCONFIG_DEFENSIVE, 0, 0);
-      }else if( p->bSafeModePersist ){
-        sqlite3_db_config(p->db, SQLITE_DBCONFIG_DEFENSIVE, 1, 0);
-      }
-#else
       sqlite3_db_config(p->db, SQLITE_DBCONFIG_DEFENSIVE, !testmode_on,0);
-#endif
     }
 
 #ifndef SQLITE_OMIT_LOAD_EXTENSION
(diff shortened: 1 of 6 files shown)


timsehn:
On the Ito additional finding (`--safe` lets `.dbconfig defensive off` through): this is upstream sqlite3 behaviour, not a doltlite divergence. Upstream's `.dbconfig` handler calls `sqlite3_db_config` directly with no `failIfSafeMode` guard (the guard covers .import/.read/.restore/.load/.clone/.cd/.archive/ATTACH/file opens), and a CLI built from the upstream shell code path reproduces the same sequence. Since the CLI policy is to match sqlite3, no doltlite issue; if it is worth raising, it belongs upstream.

PR #2725, opened 2026-09-08
Return the detector's message from dolt_verify_constraints

### LINKED ISSUE #2717: dolt_verify_constraints masks constraint detector errors
## Summary

`dolt_verify_constraints` converts constraint-detector failures to a bare SQLite error code. For expression failures such as malformed JSON in a CHECK constraint, users see only `SQL logic error` instead of the underlying detector error.

This is independent of fast-forward merge behavior from #2715: any database containing a row that makes CHECK evaluation fail reproduces it.

## Repro

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, v TEXT CHECK(json_extract(v,'$.ok')));
PRAGMA ignore_check_constraints=ON;
INSERT INTO t VALUES(1, 'not-json');
PRAGMA ignore_check_constraints=OFF;
SELECT dolt_verify_constraints('--all');
-- Error: SQL logic error
```

## Expected

Return the underlying detector message, such as `malformed JSON`, as the three-way merge path does.

## Suspected cause

`src/doltlite_verify_constraints.c` calls `sqlite3_result_error_code(context, rc)` when `doltliteDetectConstraintViolationsFiltered` fails, discarding the detailed error text.

timsehn:
Fixes #2717.

`dolt_verify_constraints` passed a null error slot to `doltliteDetectConstraintViolationsFiltered` and reported failures with `sqlite3_result_error_code()`, so an expression failure inside a CHECK came back as a bare `SQL logic error`. It now passes the slot and returns the detector's text, the way `dolt_cherry_pick` already does; the code path is unchanged otherwise.

```
CREATE TABLE t(id INTEGER PRIMARY KEY, v TEXT CHECK(json_extract(v,'$.ok')));
PRAGMA ignore_check_constraints=ON; INSERT INTO t VALUES(1,'not-json'); PRAGMA ignore_check_constraints=OFF;
SELECT dolt_verify_constraints('--all');
-- before: Error: SQL logic error
-- after:  Error: malformed JSON
```

**Tests.** Two cases in `test/doltlite_merge.sh` (`--all` and default scope) expect `malformed JSON`; against the pre-fix CLI the same statement prints `SQL logic error`. Suite 232/232, `make lint`, amalgamation CLI compiles.

**Not in this PR.** The rebase replay path (`doltlite_rebase.c`, `doltliteDetectConstraintViolationsFiltered(..., 0)`) also drops the detector text and returns only the code; it needs its own look since that function returns an int to a caller that formats its own message. Filed as #2727 with a repro.

### DIFF
--- a/src/doltlite_verify_constraints.c
+++ b/src/doltlite_verify_constraints.c
@@ -174,6 +174,7 @@ static void doltVerifyConstraintsFunc(
   int i;
   int rc;
   int nViolations = 0;
+  char *zDetectErr = 0;
   ProllyHash headCat;
   ProllyHash emptyCat;
   DoltliteCommit headCommit;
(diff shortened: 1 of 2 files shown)


timsehn:
Ito's additional finding is real and pre-existing: a verify that errors leaves the violation tables cleared (the clear runs before detection, the errored batch is discarded, and nothing rolls the clear back). Reproduced independently of this PR; filed as #2728 with a repro. Not addressed here, since this PR only changes the error message.

PR #2730, opened 2026-09-08
Write dolt_verify_constraints findings in one step

### LINKED ISSUE #2728: A failing dolt_verify_constraints erases previously recorded violations
## Summary

When `dolt_verify_constraints` fails partway (for example a CHECK whose expression errors on a row: `malformed JSON` from `json_extract`), the violations it had recorded from earlier runs are gone. The command clears the recorded findings for the tables it is about to scan before running the detectors, and an error leaves that clear in place while the new batch is never committed. The catalog ends up with fewer findings than before the failed run.

Found by Ito on PR #2725 (which only changed the error message); pre-existing.

## Repro

```sql
PRAGMA foreign_keys=OFF;
CREATE TABLE parent(id INTEGER PRIMARY KEY);
CREATE TABLE child(id INTEGER PRIMARY KEY, pid INT REFERENCES parent(id));
INSERT INTO child VALUES(1, 99);
CREATE TABLE t(id INTEGER PRIMARY KEY, v TEXT CHECK(json_extract(v,'$.ok')));
INSERT INTO t VALUES(1,'{"ok":1}');

SELECT dolt_verify_constraints('--all');              -- 1
SELECT count(*) FROM dolt_constraint_violations_child; -- 1 (foreign key, child 1 -> parent 99)

PRAGMA ignore_check_constraints=ON;
INSERT INTO t VALUES(2,'not-json');
PRAGMA ignore_check_constraints=OFF;
SELECT dolt_verify_constraints('--all');              -- Error: malformed JSON
SELECT count(*) FROM dolt_constraint_violations_child; -- 0   <-- the earlier finding is gone

DELETE FROM t WHERE id=2;
SELECT dolt_verify_constraints('--all');              -- 1
SELECT count(*) FROM dolt_constraint_violations_child; -- 1   (recreated only by the later successful run)
```

## Expected

A failed verification reports the error and leaves the recorded violations exactly as they were. Either run the clear and the detection inside one transaction that rolls back on detector error, or clear only after the detectors succeed (collect into the batch, then swap).

## Where

`src/doltlite_verify_constraints.c`: `doltliteClearConstraintViolationsForTables` / `doltliteClearAllConstraintViolations` run before `doltliteDetectConstraintViolationsFiltered`; the detector's `doltliteConstraintViolationBatchEnd(db, rc==SQLITE_OK && bPersist)` discards the new batch on error, but nothing restores what the clear removed. The merge path avoids this by holding a transaction across the pass (`doltlite_merge_cmd.c`, BEGIN / COMMIT-or-ROLLBACK around the detector call); verify has no equivalent.

timsehn:
Fixes #2728.

`dolt_verify_constraints` cleared the recorded findings for the tables it was about to scan and persisted that clear immediately; the detectors' batch was only stored on success. Two consequences: a detector error (a CHECK whose expression fails on a row, e.g. `json_extract` on `not-json`) left the catalog with fewer findings than before the call, and between the clear and the final store every other session saw the cleared set, so a concurrent `dolt_commit` could pass its violation gate (Ito's finding on the first cut of this PR).

**Fix.** The clear becomes part of the detector batch. `dolt_verify_constraints` opens the batch, drops the scanned tables from the batch's in-memory copy of the catalog (`doltliteConstraintViolationBatchDropTables`), runs the detectors (which now reuse a caller-owned batch instead of opening their own), and commits or discards the batch once. The store is written exactly once, on success: a failure leaves the recorded findings untouched, and no session can observe the cleared set without the new findings beside it. Output-only mode is unchanged. The standalone per-table clear helper lost its only caller and is removed (dead-code gate passes).

```
-- child(1) -> parent 99 recorded by a first verify; then a not-json row in t
SELECT dolt_verify_constraints('--all');              -- error
SELECT count(*) FROM dolt_constraint_violations_child; -- before: 0    after: 1
SELECT dolt_commit('-am','x');                          -- still refused: violations recorded
```

**Tests.** Three cases in `test/doltlite_merge.sh` (`--all`, default scope, fresh-session readback) assert the orphan stays recorded; before the fix the same readback printed 0. Suite 234/234, `make lint`, `test/dead_code_check.sh`.

**Note.** Based on master, so the detector call still passes a null message slot; #2725 (the message fix) touches the same lines and whichever lands second gets a small rebase.

### DIFF
--- a/src/doltlite_constraint_violations.c
+++ b/src/doltlite_constraint_violations.c
@@ -360,6 +360,7 @@ struct ConstraintViolationBatch {
   ConstraintViolationTable *aTables;
   int nTables;
   int nAppended;   /* 0 when nothing was added, matching the non-batch path */
+  int nDropped;    /* tables removed by BatchDropTables; also a write */
 };
 
 int doltliteConstraintViolationBatchBegin(sqlite3 *db){
@@ -382,7 +383,7 @@ int doltliteConstraintViolationBatchEnd(sqlite3 *db, int commit){
   int rc = SQLITE_OK;
   if( !pBatch ) return SQLITE_OK;
   doltliteSetCvBatch(db, 0);
-  if( commit && pBatch->nAppended>0 ){
+  if( commit && (pBatch->nAppended>0 || pBatch->nDropped>0) ){
     ChunkStore *cs = doltliteGetChunkStore(db);
     rc = cs ? storeUpdatedViolations(db, cs, pBatch->aTables, pBatch->nTables)
             : SQLITE_ERROR;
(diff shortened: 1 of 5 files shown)


timsehn:
Reworked after the Ito finding: the first cut restored the catalog after a detector error, which fixed the lost findings but kept the window where other sessions saw the cleared set. The clear is now part of the detector batch, so the store is written once on success and there is no intermediate state for a concurrent dolt_commit to observe. Details in the updated description.

PR #2746, opened 2026-09-09
Point concurrency contract evidence at doc/doltlite/concurrency.md

timsehn:
Master has been red since #2736: `test/concurrency_contract.tsv` carries six evidence needles of the form `README.md#<phrase>` whose phrases moved to `doc/doltlite/concurrency.md`. `concurrency_contract_test.sh` reports `evidence selector is missing` for each, which fails Build-Test (macos, windows), coverage-native-shell, and compat on every PR, including #2737.

This repoints exactly those six needles at `doc/doltlite/concurrency.md`; the `Detached revisions` needle still resolves in the README and is unchanged. All three contract tests pass locally (6/6, 13/13, 32/32).

My mistake in #2736: I grepped the contract test scripts for README and missed that the needles live in the TSV.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2747, opened 2026-09-09
Disable the Google Chrome apt source on CI runners

timsehn:
Every Linux build job on #2746 failed twice in `Install Linux build dependencies`:

```
E: Failed to fetch https://dl.google.com/linux/chrome-stable/deb/dists/stable/main/binary-amd64/Packages.gz  Hash Sum mismatch
apt-get failed after 2 attempts (exit 100)
```

The runner image (`ubuntu24/20260907.300`) ships that source; we install nothing from it. `install-apt-deps.sh` already disables the Microsoft and azure-cli sources for the same reason, so this adds `dl.google.com` to that pattern. This PR's own run is the test.


timsehn:
Superseded by #2749, which carries this fix together with the other one so CI can go green.

PR #2828, opened 2026-09-11
Let dolt_add, dolt_reset and dolt_checkout name dolt_schemas

### LINKED ISSUE #2801: dolt_schemas shows in dolt_status but table-scoped checkout/reset/add cannot address it
Found in a holistic review at 86531edccb. Verified against Dolt 2.3.1, where `CALL dolt_checkout('dolt_schemas')` reverts the trigger and clears the status row.

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, a INT);
CREATE TABLE lg(x INT);
CREATE TRIGGER tr AFTER INSERT ON t BEGIN INSERT INTO lg VALUES(NEW.id); END;
DROP TRIGGER tr;
CREATE TRIGGER tr AFTER INSERT ON t BEGIN INSERT INTO lg VALUES(NEW.id*100); END;
SELECT * FROM dolt_status;                 -- dolt_schemas|0|modified
SELECT dolt_checkout('dolt_schemas');      -- no such branch or table: dolt_schemas
SELECT dolt_reset('dolt_schemas');         -- table not found
SELECT dolt_add('dolt_schemas');           -- table not found: dolt_schemas
SELECT sql FROM sqlite_master WHERE name='tr';   -- still NEW.id*100
```

`dolt_checkout('t')` correctly leaves the trigger change behind (matches Dolt), so a view/trigger edit can only be discarded with a whole-branch `dolt_reset('--hard')`. `dolt_diff_dolt_schemas` is likewise absent.

**Root cause:** `src/doltlite_checkout.c:894-916`, `src/doltlite_add.c:805`, `src/doltlite_reset.c:644` resolve names only against `TableEntry` arrays; views/triggers live as `SchemaEntry` rows, while `src/doltlite_status.c:404-409` synthesizes the pseudo-row.

**Fix:** special-case the name `dolt_schemas` in checkout/reset/add to copy the non-table `SchemaEntry` set from HEAD/STAGED, mirroring what status reports.

timsehn:
Fixes #2801.

`dolt_status` reports a changed view or trigger under the name `dolt_schemas`, but no table-scoped command accepted that name:

```
SELECT * FROM dolt_status;          -- dolt_schemas | 0 | modified
SELECT dolt_checkout('dolt_schemas');  -- no such branch or table: dolt_schemas
SELECT dolt_reset('dolt_schemas');     -- table not found
SELECT dolt_add('dolt_schemas');       -- table not found: dolt_schemas
```

All three resolve names against catalog entries, and views and triggers are rows in the master tree with no entry of their own. A view or trigger edit could therefore only be discarded with a whole-branch `dolt_reset('--hard')`.

Dolt accepts the name for all three, so this matches it.

## How each one works

- **add** stages the view/trigger set. The staged-master composer already models those rows as "entryless" with a from-working flag, which `-A` used; naming `dolt_schemas` now sets it for a named add. Row changes to ordinary tables stay unstaged.
- **reset** unstages them, taking the entryless rows from HEAD through the same composer.
- **checkout** replaces the live set with the source's, dropping every live view and trigger and replaying the source definitions. That is how checkout already reverts a table, so it reuses the existing SQL-level approach rather than introducing a second mechanism.

## Parity

Status transitions are identical to Dolt 2.3.3 across the whole sequence:

| step | DoltLite | Dolt |
| --- | --- | --- |
| after edit | `dolt_schemas:0` | `dolt_schemas:0` |
| after `dolt_add` | `dolt_schemas:1` | `dolt_schemas:1` |
| after `dolt_reset` | `dolt_schemas:0` | `dolt_schemas:0` |
| after `dolt_checkout` | 0 rows | 0 rows |

## Evidence

New suite `test/doltlite_schemas_scoped_ops.sh`, registered in the suite manifest, 18 assertions.

| build | result |
| --- | --- |
| without the fix | 9 failures |
| with the fix | 18 passing |

Checkout is covered for all five shapes, each restoring exactly the committed set and clearing status: a view added, a view dropped, a trigger modified, a trigger dropped, and no change at all. Add is checked to stage only the schema set and leave a pending row change alone, that a following commit lands the new trigger and leaves the row change pending, and that reset puts it back while keeping the working copy. One case pins that the name is unambiguous, since the `dolt_` prefix is reserved for tables.

Also green: 133 shell suites, all 35 gated C tests, `make lint`, and the schemas, checkout, reset, add, triggers and status oracle suites against Dolt.

The checkout helper lives in `doltlite_schemas.c` rather than `doltlite_checkout.c`, which is at its 1500-line lint cap.

### DIFF
--- a/src/doltlite_add.c
+++ b/src/doltlite_add.c
@@ -633,6 +633,9 @@ int doltliteStageNamedTables(
   int nStaged = 0;
   int i;
   int updateMaster = 0;
+  /* Views and triggers are master rows with no catalog entry, which
+  ** dolt_status reports under one name. Naming it stages that whole set. */
+  int stageSchemas = 0;
   int rc;
 
   SchemaEntry *aWorkSchema = 0;
@@ -717,6 +720,13 @@ int doltliteStageNamedTables(
       if( zPrior && sqlite3_stricmp(zPrior, zTable)==0 ) break;
     }
     if( j<i ) continue;
+
+    if( sqlite3_stricmp(zTable, "dolt_schemas")==0
+     && !sqlite3FindTable(db, zTable, "main") ){
+      stageSchemas = 1;
+      updateMaster = 1;
+      continue;
+    }
 
     pLive = sqlite3FindTable(db, zTable, "main");
     if( pLive ) zTable = pLive->zName;
@@ -933,7 +943,7 @@ int doltliteStageNamedTables(
               pStagedMaster ? &pStagedMaster->root : 0,
               pStagedMaster ? pStagedMaster->flags : 0,
               (const char**)azTouched, nTouched,
-              aStaged, nStaged, 0,
+              aStaged, nStaged, stageSchemas,
               &composedRoot);
       if( rc!=SQLITE_OK ){
         ADDNAMED_FREE_ALL();
--- a/src/doltlite_checkout.c
+++ b/src/doltlite_checkout.c
@@ -17,9 +17,20 @@ struct CheckoutSchemaInfo {
   int hasCurrent;
   int hasSource;
   int rebuilt;
+  int isSchemas;
   char *zCurrentSql;
   char *zSourceSql;
 };
+
+#define CHECKOUT_SCHEMAS_NAME "dolt_schemas"
+
+/* Views and triggers have no catalog entry of their own; they are rows in the
+** master tree that dolt_status reports under one name. Checking that name out
+** replaces the live set with the source's. */
+static int checkoutIsSchemasName(const char *zName){
+  return zName && sqlite3_stricmp(zName, CHECKOUT_SCHEMAS_NAME)==0;
+}
+
 
 static void checkoutSchemaInfoClear(CheckoutSchemaInfo *aInfo, int nInfo){
   int i;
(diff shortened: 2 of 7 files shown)


timsehn:
@itoqa Both failures were real and are fixed in a follow-up commit. Thanks.

The schema pass drops and recreates the live objects of each name in turn, and an unknown name was only refused later, by the catalog pass. So a failed checkout left every name ahead of it already rewritten:

```sql
CREATE VIEW v2 AS SELECT 2;                         -- uncommitted
SELECT dolt_checkout('dolt_schemas','nosuchtable'); -- fails
-- v2 was gone anyway
```

Names are now checked as their schema is loaded, before anything is dropped.

One correction to the report: this is **not** specific to this PR's code path. The same hole was already there for ordinary tables, and the fix covers both. On master, before any of my changes:

```sql
ALTER TABLE t ADD COLUMN b INT;
SELECT dolt_checkout('t','nosuchtable');   -- fails
-- column b was dropped
```

That case now survives too, and there is a regression test for each.

The error message was also misleading: it named the first argument, so it claimed `dolt_schemas` did not exist when `nosuchtable` was the missing one. It now names the one that was actually missing.

Four cases added to `test/doltlite_schemas_scoped_ops.sh`, bringing it to 22. Each fails without the fix.

PR #2875, opened 2026-09-12
Fail a test suite that stops before reporting its results

### LINKED ISSUE #2865: [P1] Shell suites report success when they die mid-run; oracle_generated_columns_vc_test.sh records failures and exits 0
Found in a holistic review at 333f5af477.

A shell suite that dies partway through reports success. 133 of the 134 suites under `test/` that use `set -u` have no `set -e` and no EXIT/ERR trap, so the only nonzero exit is the final `if [ $fail -gt 0 ]` at the bottom of the file. Anything that aborts the script before that line yields exit 0, and the runner counts it as a pass.

This is live, not theoretical. `test/oracle_generated_columns_vc_test.sh` records real failures and still exits 0:

```sh
bash test/oracle_generated_columns_vc_test.sh <engine-that-errors> dolt; echo $?
#   FAIL: d_merge_virtual_count; got BROKEN ENGINE OUTPUT
#   FAIL: e_conflict_detected; got BROKEN ENGINE OUTPUT
# test/oracle_generated_columns_vc_test.sh: line 144: OUTPUT: unbound variable
# 0
```

Line 144 is `[ "$YVAL" = "$((XVAL*2))" ]`. Under `set -u`, arithmetic expansion over non-numeric engine output ends the script, and bash exits 0. Minimal demonstration:

```sh
set -u
fail=0
X="BROKEN ENGINE OUTPUT"
[ "1" = "$((X*2))" ] && echo ok || { fail=1; echo "FAIL recorded"; }
if [ $fail -gt 0 ]; then exit 1; fi
# GOT: "OUTPUT: unbound variable", rc=0
# with X="" instead: "FAIL recorded", rc=1
```

The suite also stops at test E, so sections F through J (cherry-pick, revert, multiple generated columns) never run and their absence is invisible. The trigger is reachable from a plain regression: any change that makes `dolt_merge` or that `SELECT` error turns this suite green and silent.

Two more oracle suites pass against an engine that does nothing but print to stderr and exit 1, because every assertion they make is satisfied by any failure:

| Suite | Result against a broken engine |
|---|---|
| `oracle_generated_columns_vc_test.sh` | records failures, exits 0 |
| `vc_oracle_hashof_errors_test.sh` | 4 passed, 0 failed |
| `vc_oracle_null_args_test.sh` | 13 passed, 0 failed |

Those two assert that invalid refs and NULL arguments error. They cannot distinguish erroring for the right reason from erroring for any reason.

Method: ran all 94 `vc_oracle_*_test.sh` / `oracle_*_test.sh` with a wrapper that prints garbage and exits 1. 3 of 94 passed. The same sweep over the 135 native suites via `DOLTLITE_BUILD_DIR` pointed at the wrapper gives 129 failed, and the 6 that pass are legitimately engine-independent (build guards, packaging layout, a negative test).

Same family as #2780 and #2791, which fixed suites whose engine argument was ignored. This is a different mechanism: the engine is used, the failures are detected, and the exit status is still lost.

Distinct from #2857, which is about `run_test` ignoring `$1` and the engine rc.

**Fix:**
- Report the tally from an `EXIT` trap so a mid-script death cannot suppress it, and have the trap exit nonzero when the script did not reach its own end marker.
- Never feed raw engine output to `$(( ))`; compare as strings, or validate numeric first.
- Give the error-parity suites a positive control so that erroring on everything fails them.

timsehn:
Fixes #2865.

Suites use `set -u` without `set -e`, and the only non-zero exit is the final tally check at the bottom of the file. Anything that ends the script before that line leaves bash's status at 0, and `run_doltlite_tests.sh` counts it as a pass. 133 of the 134 suites that set `set -u` are shaped that way.

`test/oracle_generated_columns_vc_test.sh` is the live case. `$((XVAL*2))` over non-numeric engine output ends it at test E, so it exits 0 with two failures already recorded and sections F through J never run.

```
  FAIL: d_merge_virtual_count; got BROKEN ENGINE OUTPUT
  FAIL: e_conflict_detected; got BROKEN ENGINE OUTPUT
test/oracle_generated_columns_vc_test.sh: line 144: OUTPUT: unbound variable
rc=0
```

**Three changes.**

`test/run_guarded_suite.sh` runs a suite, tees its output, and fails it when no `N passed, M failed` tally was printed. The native runner and the four CI oracle loops (`test.yml` twice, `macos-test.yml`, `sanitizers.yml`) invoke suites through it. I chose a wrapper over an `EXIT` trap in the shared libs because 200 of 312 suites already install their own `trap ... EXIT`, and a second `trap` replaces the first, so a lib-installed guard would be silently dropped by exactly the suites that use traps.

The arithmetic in the generated-columns suite now validates that the value is numeric before doing arithmetic on it. The suite runs to the end and reports all 24 cases instead of dying at 2.

`vc_oracle_require_working_engines` in `test/lib/vc_oracle_common.sh` gives `vc_oracle_hashof_errors_test.sh` and `vc_oracle_null_args_test.sh` a positive control. Every assertion in those two is "this must error", which an engine that errors at everything satisfies; they now prove both sides can run a known-good script first.

Four single-assertion suites ended on a bare `PASS` line with no tally, so they would have tripped the guard. They now print one on every exit path, including the nofollow suite's skip. I did not add an exemption list: a suite the guard skips is the hole this fixes.

**Fail-before / pass-after**

| | before | after |
|---|---|---|
| `oracle_generated_columns_vc_test.sh` against a broken engine | rc=0, dies at test E | rc=1, 0 passed / 24 failed |
| `vc_oracle_hashof_errors_test.sh` against a broken engine | 4 passed, 0 failed | rc=1, engine_probe fails |
| `vc_oracle_null_args_test.sh` against a broken engine | 13 passed, 0 failed | rc=1, engine_probe fails |
| all 93 oracle suites against a broken engine | 3 pass | 0 pass |

`test/run_guarded_suite_test.sh` is a self-test for the guard, wired into the `no-dead-code` job next to the existing orphaned-suite self-test. It builds the #2865 shape as a fixture and asserts the guard catches it, that a completed pass stays 0 and a completed failure stays non-zero, and that the fixture still reproduces the bug unguarded.

**No false failures.** I checked the marker against every suite on a healthy engine before adopting it:

| Gate | Result |
|---|---|
| `test/run_doltlite_tests.sh` (guarded) | 135 passed, 0 failed, zero GUARD FAIL |
| all 93 oracle suites through the guard, real engine | all rc=0, zero GUARD FAIL |
| `test/lint_orphaned_suites.sh` | OK, 0 quarantined |
| `test/run_guarded_suite_test.sh` | 5 passed, 0 failed |
| native runner against a broken engine | 130 of 135 fail; the 5 that pass are the build-guard and packaging suites that never touch the engine |

An earlier guarded run showed four failures in `doltlite_commit.sh`, `doltlite_cherry_pick.sh`, `doltlite_rebase.sh` and `doltlite_pk_clustered_rowid_readonly.sh`. That was a stale `build/doltlite` predating the `--amend` refusal that landed in #2861; they are the `amend_during_merge_*` cases and they pass after a rebuild. The table above is the rebuilt run.


timsehn:
Pushed c1bc52540b after `no-dead-code` went red on Ubuntu.

The self-test asserted that the unguarded dying fixture exits 0. That is true on bash 3.2 and false on bash 4.4+, so the assertion was platform-dependent and the job failed on Ubuntu. Verified:

```
/bin/bash 3.2.57 (macOS)      -> rc=0
bash 5.3.15 (Linux container) -> rc=1
```

The invariant that holds on both is that the abort never reaches the tally, which is exactly what the guard keys on, so the self-test now asserts that instead. It passes under both bash 3.2 and bash 5.3.

This also narrows the original claim in #2865, which I have corrected there: the swallowed exit status only reads green in the macOS oracle loop. What the guard fixes on every platform is the silent truncation (the suite stops at test E, so sections F through J never run), an early `exit 0` on any return path, and the two error-parity suites passing an engine that only errors.

timsehn:
Rebased onto master and pushed two more commits after CI found two cases my local run could not.

`engine_floor_test.sh` landed in #2871 after this branch was cut. It ends on a bare `PASS` line, so the guard rejected it; both its exit paths now print a tally.

`doltlite_open_nofollow.sh` failed in `coverage-native-shell` because that job has no `libdoltlite.a`, so the suite takes an early `SKIP` path with no tally. Rather than chase every skip path in every suite, the guard now accepts a suite that ends on a skip. The skip has to be the last thing the suite said, so an intermediate skip followed by an abort is still a failure, and the self-test pins both halves of that.

Re-verified after the change:

| Gate | Result |
|---|---|
| `test/run_doltlite_tests.sh` (guarded, 136 suites incl. the new one) | 136 passed, 0 failed |
| all 93 oracle suites against a broken engine | 0 pass |
| `test/run_guarded_suite_test.sh` on bash 3.2 and bash 5.3 | 7 passed, 0 failed |
| `doltlite_open_nofollow.sh` with no libdoltlite.a | skips, rc=0 |

CI is green: 67 checks passed, 1 skipped, 0 failures.

timsehn:
Good catch, and all three reports are the same real hole. Fixed in 857cca7bd5.

The guard accepted a tally anywhere in the output as proof of completion. That is unsound in this repo, not just against a crafted fixture: suites print sub-tallies as they go, and `doltlite_regression_test_c.sh` prints one from the C binary before the shell prints its own. A run that stopped after any of those looked finished.

Reproduced your case before changing anything:

```
$ bash test/run_guarded_suite.sh early_tally_then_dies.sh
Results: 1 passed, 0 failed
phase 2 starting
early_tally_then_dies.sh: line 6: NUMBER: unbound variable
rc=0
```

Completion is now stated rather than inferred. A suite ends by emitting `__SUITE_COMPLETE__`, and the guard requires it from any suite claiming success. A non-zero status already fails the run, so only the success path has to prove it got there, and a suite that never ran still says so with a trailing `SKIP:`.

Rather than add the line to 230 suites by hand, it comes from the shared finish helpers: `dltest_finish` and `stock_oracle_finish` cover 95, and a new `vc_oracle_finish` replaces the tally block that 72 oracle suites had each copied. That is why the diff deletes more than it adds.

All three of your fixtures now fail, and two are pinned in the self-test:

| fixture | before | after |
|---|---|---|
| early tally, then dies | rc=0 | rc=1 |
| incidental tally in setup, then `exit 0` | rc=0 | rc=1 |
| tally, then truncation text | rc=0 | rc=1 |

Re-verified against every real suite, since a stricter rule risks inventing failures:

| Gate | Result |
|---|---|
| `test/run_doltlite_tests.sh` (136 suites) | 136 passed, 0 failed, zero GUARD FAIL |
| all 93 oracle suites, real engine | all rc=0, zero GUARD FAIL |
| all 93 oracle suites, broken engine | 0 pass |
| `test/run_guarded_suite_test.sh` on bash 3.2 and 5.3 | 9 passed, 0 failed |
| `test/lint_orphaned_suites.sh` | OK, 0 quarantined |

timsehn:
Green: 67 checks passed, 1 skipped, 0 failures.

One more fix after the previous push. `vc_oracle_version_test.sh` and `vc_oracle_correct_schema_merge_matrix_test.sh` kept their own copy of the tally block without sourcing `lib/vc_oracle_common.sh`, so replacing that block left them calling a function they had never loaded:

```
../test/vc_oracle_version_test.sh: line 169: vc_oracle_finish: command not found
```

The oracle bucket caught it. My local sweep had not, and the reason is worth recording: the sweep appended every suite's status to one shared results file, and a `rm -f` of that file raced with a still-running background job, so a stale passing row from an earlier run masked the failure. It reported 93 clean while the failure was sitting in the same file.

The sweep now writes one file per suite, so a stale row cannot mask a real one, and it carries a deliberately broken suite as a control. Re-run after the fix:

| Gate | Result |
|---|---|
| 93 oracle suites + 1 planted broken control | only the planted control fails |
| `test/run_doltlite_tests.sh` (136 suites) | 136 passed, 0 failed, zero GUARD FAIL |
| `test/run_guarded_suite_test.sh` on bash 3.2 and 5.3 | 9 passed, 0 failed |
| `test/lint_orphaned_suites.sh` | OK, 0 quarantined |

PR #2948, opened 2026-09-15
Bound open time by chunk count as well as WAL bytes

### LINKED ISSUE #2909: Database open time is still linear in commits since GC; the 64 MiB checkpoint never fires for small commits
Found in a full-repo review at `0ba280f06f`.

#2266 reported database open time growing with incremental commits and was fixed for the shape it reported. The growth is still there for ordinary small commits, because the automatic WAL checkpoint only fires at 64 MiB and a stream of small commits never reaches it.

Measured, 5000-row table, one-row-update commits, median of 5 opens (`SELECT 1;`):

| commits since GC | 0 | 250 | 500 | 1000 | 2000 | 4000 |
|---|---|---|---|---|---|---|
| open | 4.6 ms | 10.5 | 15.3 | 23.6 | 41.2 | **78.5 ms** |

Linear, ~18.5 µs per commit, unbounded. `dolt_gc()` resets it to 4.8 ms **with the file still 24 MB** — the data is unchanged, so this is WAL replay, not data volume.

The same effect on an ordinary freshly built database: a 500k-row table with two indexes opens in **39.7 ms**; after `dolt_gc()` the same database opens in **4.4 ms**. Stock SQLite opens the equivalent data in 4.1 ms. So DoltLite is at open-parity with stock once checkpointed, and ~10x worse before.

Every reader pays this, and the deployment shape that suffers most — a connection per request against a database that commits often — is a common one.

## Cause

`csWalCheckpointThreshold()` returns 64 MiB (`src/chunk_wal.c:134`):

```c
static i64 csWalCheckpointThreshold(void){
  i64 n = 64*1024*1024;
#if defined(SQLITE_TEST) || defined(DOLTLITE_MECH_REPRO)
  const char *z = getenv("DOLTLITE_WAL_CHECKPOINT_THRESHOLD");
  ...
#endif
  return n;
}
```

`csWalCheckpointDue` (`:145`) compares `iFileSize - iCheckpointReplay` against it; open replays from `iCheckpointReplay` (`csReplayWalFrom`, `src/chunk_wal.c:1290` / `:1376`). Checkpoints are written only when due, on commit (`src/chunk_store_commit.c:518`) or on close (`src/chunk_store.c:639`) — so even a clean close does not checkpoint below the threshold.

Traced with an instrumented build, 3000 small commits on a 5000-row table:

| threshold | checkpoints written | open time |
|---|---|---|
| 64 MiB (shipped) | **0** | 0.06 s |
| 8 MiB | 2 | 0.07 s |
| 4 MiB | 5 | 0.01 s |

At the shipped threshold the safety valve never engages. Write cost is unchanged from 4 to 64 MiB (0.6 s for 3000 commits either way); it only triples at 1-2 MiB. I could not explain why 8 and 16 MiB checkpoint but do not improve open time — that deserves its own look and I am not drawing a conclusion from it.

The threshold override is compiled out of release builds, so there is no supported way to bound this. Verified: setting `DOLTLITE_WAL_CHECKPOINT_THRESHOLD=1048576` has no effect on `build/doltlite` (44.9 ms vs 41.2 ms unset, i.e. unchanged).

## The regression test cannot see it

`test/doltlite_open_perf.sh` measures exactly the right quantity and is calibrated so it cannot fail. It commits `zeroblob(2097152)` — 2 MiB per commit — which is the payload from #2266's own repro script. At that size 64 commits is 128 MiB, so **every stage is already past a checkpoint** and the suite measures the post-checkpoint steady state:

```
as shipped (2 MiB payload):
  64 commits: 969ms   128 commits: 880ms   256 commits: 717ms
  growth: -252ms; maximum allowed: 750ms
  PASS
```

Open time *falls* across stages. Change only the payload constant to 512 bytes and the same suite fails on master today:

```
same suite, payload 512 bytes:
  1000 commits: 487ms   2000 commits: 873ms   4000 commits: 1669ms
  growth: 1182ms; maximum allowed: 750ms
  FAIL: open time grew by 1182ms
```

The regression test inherited the reporter's payload size, and that payload is the one value that hides the remaining problem.

## Fix

Make the checkpoint actually bound open cost for small-commit workloads, and make the threshold a supported setting available in release builds rather than a `SQLITE_TEST`-only env var. Then re-calibrate `doltlite_open_perf.sh` to a small payload and more stages so it proves the bound — it already detects the regression, it just is not pointed at it.

timsehn:
Open replays the chunk WAL from the last checkpoint, and `csWalCheckpointDue` only considered the byte span — 64 MiB. A stream of small commits never reaches that, so open time grew with every commit and only `dolt_gc()` ever reset it.

Bytes are the wrong unit on their own: replay costs a read per record, so thousands of small commits cost thousands of reads while the byte span stays tiny, and a handful of 2 MiB commits blow the byte budget after a few records. The fix checks both a byte limit and a chunk limit and checkpoints on whichever comes first.

## Measured

Open time (median of 5, `sqlite3_open` + `SELECT 1`, 5000-row table, one-row-update commits):

| commits since gc | 1000 | 2000 | 4000 | 8000 |
|---|---|---|---|---|
| master | 9.4 ms | 18.5 | 37.1 | (growing) |
| this PR | 22.2 | 23.2 | 22.8 | 18.4 |

Flat instead of linear. The plateau is the chunk limit: at 1024 it is ~11 ms, at 4096 ~48 ms.

A freshly built database — 500k rows, two indexes, the issue's second scenario — gets the bigger win, for nothing:

| | build | open | size |
|---|---|---|---|
| master | 0.7 s | 55.3 ms | 43 MB |
| this PR | 0.7 s | **0.2 ms** | 44 MB |

The cost is file growth between collections on the small-commit workload: at 8000 commits, 33 MB at a 4096 chunk limit against 37 MB at 2048 and 44 MB at 1024. `dolt_gc()` reclaims it. Commit throughput is unchanged at every setting measured (1.0–1.3 s for 4000 commits).

## Defaults

8 MiB / 2048 chunks. 2048 halves open time against 4096 for 19% more file before collection, and leaves 1024 available to anyone who wants to spend more disk on it.

Both limits are now read outside `SQLITE_TEST` builds — `DOLTLITE_WAL_CHECKPOINT_THRESHOLD` and `DOLTLITE_WAL_CHECKPOINT_CHUNKS`, documented in `doc/doltlite/cli.md` alongside `DOLTLITE_HTTP_TIMEOUT_MS`. Bounding open time trades file growth, and which side of that a deployment wants is not the engine's call.

## Tests

`test/doltlite_open_perf.sh` measured the right quantity but committed `zeroblob(2097152)`, so every stage was already past a 64 MiB checkpoint and it measured the post-checkpoint steady state. The payload is now 512 bytes (a named `PAYLOAD_BYTES`) over 1000/2000/4000 commits. Same 750 ms budget:

```
=== FAIL-BEFORE (master-equivalent thresholds) ===
  1000 commits: 911ms    2000: 1657ms    4000: 3109ms
  growth: 2198ms; maximum allowed: 750ms
FAIL: open time grew by 2198ms

=== PASS-AFTER (shipped defaults) ===
  1000 commits: 251ms    2000: 668ms     4000: 430ms
  growth: 179ms; maximum allowed: 750ms
PASS
```

Reproducible across three runs: 179 / 199 / 194 ms. The recalibrated suite is also cheaper than the old one — 7000 small commits instead of ~900 MB of zeroblobs.

`test/doltlite_checkpoint_threshold_env.sh` existed to assert the env var is ignored in production builds. It now asserts the opposite, for both limits, plus that a single small commit still checkpoints nothing under the defaults.

## Validation

`make lint`, `run_c_tests.sh . coverage`, `run_doltlite_tests.sh`, the `storage`, `core-sql`, `version-control`, `ported`, `fault-fuzz` and `ddl-schema` testfixture buckets, `doltlite_checkpoint_perf.sh`, `doltlite_perf.sh`, `doltlite_structural.sh` (the structural-sharing gate, the one most exposed to more frequent checkpoints), and all five Docs-validation scripts.

Fixes #2909


timsehn:
CI caught `index5 index5-1.3` as a stale divergence entry: the change makes it pass.

index5-1.3 asserts that building an index writes mostly forward — `nForward > 2*(nBackward + nNoncont)` over the raw VFS write offsets. A 100k-row `CREATE INDEX` now crosses the chunk limit several times, and each checkpoint writes a long forward run of tree pages, which carries the ratio.

Confirmed it genuinely passes rather than erroring out, with the same binary and only the limits differing:

```
=== master-equivalent (64 MiB, no chunk limit) ===
index5-1.3...
1 errors out of 4 tests
!Failures on these tests: index5-1.3

=== shipped defaults (8 MiB / 2048) ===
index5-1.3... Ok
0 errors out of 4 tests
```

`index5-1.3... Ok` on three consecutive runs. Entry removed from `known_testfixture_divergences.txt`; ratchet lowered `gates 4698 -> 4697`, `intentional 3379 -> 3378`. `make lint` passes and the `ddl-schema` bucket is now green locally (14994 passing).

The `ported` bucket — which carries `doltlite_recover_corruption.test` and its `DOLTLITE_WAL_CHECKPOINT_THRESHOLD` usage — is also green (10855 passing). The `storage` and `core-sql` buckets show only the macOS-only artifacts (`avfs`, `loadext`, `shell1`, `recover`, `bigfile`, `bigfile2`, `bigmmap`); those lists are byte-identical with and without the change, and both buckets are green in this PR's own CI.

Issue #2917, opened 2026-09-14
A NOCASE index costs a full index scan at plan time on every prepare while edits are pending


timsehn:
Found in a full-repo review at `0ba280f06f`.

While any edit is pending on a table, every `prepare` of a statement that considers a NOCASE index pays a full scan of that index. The cost is linear in table size and is pure planning overhead — both engines pick the same plan.

## Measured

Inside `BEGIN`, one `UPDATE` to the indexed column, then 100 index-equality `SELECT`s, whole-process wall time (the ~30 ms floor is process startup, visible in the two control columns):

| rows | DoltLite, NOCASE index | DoltLite, BINARY index | stock, NOCASE index |
|---|---|---|---|
| 50,000 | 130.0 ms | 32.6 ms | 31.2 ms |
| 200,000 | 428.6 ms | 44.6 ms | 28.8 ms |

Net of startup that is roughly **1.0 ms per statement at 50k rows and 4.0 ms at 200k** — 4x the rows, 4x the cost — while the BINARY index and stock stay flat. Only NOCASE indexes are affected; there is an early-out for every other collation.

## Cause

`src/where.c:4146-4163` runs inside the per-candidate-index loop of `whereLoopAddBtree`, so it fires on every index on every planning pass:

```c
if( !pProbe->bNocaseNul
 && iDb>=0 && iDb<db->nDb && db->aDb[iDb].pBt
 && !sqlite3BtreeUsesOrig(db->aDb[iDb].pBt) ){
  rc = sqlite3BtreeProllyIndexHasNocaseNul(
      db->aDb[iDb].pBt, pProbe->tnum, pProbe->nKeyCol,
      pProbe->azColl, &hasNocaseNul);
```

`sqlite3BtreeProllyIndexHasNocaseNul` (`src/prolly_btree.c:62`) answers by scanning the index, and its memo is disabled whenever the index has any uncommitted edit (`src/prolly_btree.c:89-90`):

```c
pMap = (ProllyMutMap*)pTE->pPending;
cacheable = !pMap || prollyMutMapIsEmpty(pMap);
if( cacheable && pTE->nocaseNulState
 && prollyHashCompare(&pTE->nocaseNulRoot, &pTE->root)==0 ){
```

So a transaction that writes once and then reads many times — an extremely ordinary shape — rescans the whole index per prepared statement.

## Fix

[...]

PR #2961, opened 2026-09-16
Carry the NOCASE-NUL answer across a flush

### LINKED ISSUE #2956: NOCASE index planning rescans the index once per commit; the memo is keyed on the committed root
Follow-up to #2917, split out of PR #2955 rather than bundled into it.

#2917 fixed the case where the memo was thrown away because edits were *pending*. The memo is still keyed on the committed root, so every commit invalidates it and the next prepare scans the whole index again. A statement-per-commit workload — a connection per request, or plain autocommit — therefore still pays a full index scan per commit, linear in table size.

## Measured

50 `UPDATE` + `SELECT` pairs in autocommit, whole-process wall time (the ~50 ms floor is process startup, visible in the stock column):

| rows | master | PR #2955 | stock |
|---|---|---|---|
| 50,000 | 106.8 ms | 102.7 ms | 50.0 ms |
| 200,000 | 271.7 ms | 280.5 ms | 49.4 ms |

4x the rows, ~2.6x the cost, and #2955 makes no difference here — as expected, since each commit moves `pTE->root`.

For contrast, the shape #2955 does fix (one write then many reads inside `BEGIN`, 200k rows, 800 statements) went from 3004 ms to 65 ms.

## Cause

`sqlite3BtreeProllyIndexHasNocaseNul` (`src/prolly_btree.c`) memoises on `pTE->nocaseNulRoot == pTE->root`. That is correct but conservative: the new tree is the old tree plus the flushed edits, so a memo saying "no NOCASE NUL" could carry forward whenever none of the flushed inserts brings one in. Deletes can only remove such a key, so ignoring them errs toward a rescan.

## Fix

Two candidates, both from #2917's own Fix section:

1. **Carry the memo across a flush.** `applyMutMapToTableRoot` (`src/prolly_btree_mutation.c:20`) is the single choke point for DML flushes and holds both `pTE` and the map being flushed. It does not have the index's collation list, which `recordHasNocaseNul` needs — so the memo would have to cache the `azColl` it was populated for, alongside the answer, and the TableEntry copy/free paths would need to carry and release it. Merge flushes go through `prollyMutateFlush` directly and would simply fall back to a rescan.

2. **A sticky per-index flag set at write time**, cleared only on rebuild. Simpler to reason about, but it never goes back to "no NUL" after the offending row is deleted, which costs plan quality permanently. Note `bNocaseNul` is already sticky on the in-memory `Index` object once set, so this may be closer to current behaviour than it looks.

A collation-free proxy ("does any flushed record contain a 0x00 byte") is not viable: serial-type headers and integer payloads contain NUL bytes routinely, so it would clear the memo on nearly every flush.

Low severity — it needs a NOCASE index and a commit-per-statement workload — but it is the half of #2917 that the merged fix does not reach.

timsehn:
#2917 stopped the rescan while edits were *pending*. The memo is still keyed on the committed root, so every commit invalidated it and the next prepare rescanned the whole index — the half #2917 did not reach.

A flush makes the new tree the old one plus the edits being written. So an answer of "this index holds no NOCASE key with a NUL byte" still holds unless one of the flushed inserts brings one in, which is cheap to check where rescanning is not. `applyMutMapToTableRoot` is the single choke point for DML flushes and has both the table entry and the map.

The obstacle noted in the issue was that the flush has no access to the index's collation list. It does not need it: the scanner only ever asks "is key column *i* NOCASE", so a **bitmask** captures it exactly, outlives the `Index` the names came from, and costs the table entry no ownership — no strings to copy, no teardown sites to touch. Indexes past 32 key columns simply do not carry.

Only the negative answer carries. A tree that *had* such a key may have just lost it, and claiming otherwise without looking would cost plan quality, so that case still rescans.

## Measured

Autocommit write/read pairs at 200k rows, NOCASE index against an otherwise identical BINARY one:

| pairs | master | this PR | BINARY control |
|---|---:|---:|---:|
| 25 | 157.7 ms | 72.1 ms | ~35 ms |
| 100 | 455.0 ms | 89.3 ms | ~53 ms |
| 400 | 1657.5 ms | 166.2 ms | ~130 ms |

The NOCASE penalty over BINARY goes from growing linearly with commits (124 -> 399 -> 1527 ms) to a **constant ~36 ms** — one initial scan, then carried. The issue's own repro (50 pairs) goes 246.4 ms -> 75.4 ms at 200k rows.

## Tests

`test/doltlite_perf.sh` gains `nocase_plan_25_to_400_autocommit`, the axis that exposes this — per-statement cost must not grow with how many commits came before:

```
=== FAIL-BEFORE (master) ===
  25 autocommit pairs: 149ms
  400 autocommit pairs: 1478ms
  FAIL: nocase_plan_25_to_400_autocommit — 149ms → 1478ms (991%/400%)

=== PASS-AFTER ===
  25 autocommit pairs: 67ms
  400 autocommit pairs: 157ms
  PASS: nocase_plan_25_to_400_autocommit — 67ms → 157ms (234%/400%)
```

The risk here is a stale memo, so `test/doltlite_nocase_index.sh` gains three cases asserting the answer through its visible form — whether the planner trusts the index order — with commits interleaved: absent-and-stays-absent across commits, introduced-after-a-clean-probe, and a two-column NOCASE index. **All three pass on master too, by design**: they pin behaviour, not speed. 12/12 on both engines.

Separately, plans are byte-identical between master and this PR across a matrix covering clean probes, updates, inserts, a NUL insert, a delete of it, and a reopen.

## Validation

`make lint`, `run_c_tests.sh . coverage`, `run_doltlite_tests.sh` (all suites), `doltlite_perf.sh`, and the `core-sql`, `ddl-schema` and `storage` testfixture buckets — `ddl-schema` clean, the other two showing only the macOS-only set (`avfs`, `loadext`, `shell1`, `recover`, `bigmmap`, `bigfile`, `bigfile2`).

Fixes #2956


timsehn:
Thanks — both Ito findings reproduce, and neither is introduced by this PR. Detail and evidence below.

## The failing gate: connection-per-request, not a regression

The 15.797 ratio is the **connection-per-pair** shape. This PR's memo lives in the in-memory `TableEntry`, so it dies with the connection and cannot help there. Measured at 200k rows, each `UPDATE`+`SELECT` pair in its own process:

| pairs | master | this PR |
|---|---:|---:|
| 25 | 1.04 s | 1.07 s |
| 100 | 4.67 s | 4.81 s |

Linear in pairs and **identical on both engines**; extrapolating to 400 gives ~16x, which is the reported number. So this is a pre-existing cost that this PR neither causes nor fixes.

Within one connection — the axis this PR is about — the carry holds, and it holds at scale:

| pairs | master 200k | this PR 200k | master 1M | this PR 1M |
|---|---:|---:|---:|---:|
| 25 | 163.9 ms | 66.4 ms | 655.3 ms | 247.2 ms |
| 100 | 440.3 ms | 84.2 ms | 2042.8 ms | 302.8 ms |
| 400 | 1663.3 ms | 157.5 ms | 7520.9 ms | 365.3 ms |

Master grows 10-11x across that range; this PR grows 1.5-2.4x, most of which is the write work itself (an equivalent BINARY index moves 35 -> 140 ms over the same range).

The gate as written uses one connection, which is why it passes here. I have filed the per-connection scan separately rather than widen this PR: fixing it means the answer has to outlive the connection, which is a storage-format question, not a memo question.

## The sticky-flag finding is correct

`pProbe->bNocaseNul` is set once on the in-memory `Index` and never cleared, and `src/where.c` skips the probe entirely once it is set — so after the NUL row is deleted the planner stays on the temp-B-tree plan even though the table memo was dropped. My own plan matrix in this PR shows the same thing, and shows master doing it too, so it is pre-existing and not a merge blocker. Filed with the above.

Both are now tracked in the follow-up issue.

timsehn:
Pushed two fixes for the CI failures. Neither was in the engine change.

## The perf gate was measuring the disk, not the planner

`nocase_plan_25_to_400_autocommit` timed autocommit pairs directly. A commit costs a sync, and on the macOS runner that sync is ~100ms — so 2491ms for 25 pairs and 38302ms for 400 is just 16x the commits, and the planning cost the gate exists to isolate is invisible beside it. The gate would have failed there on any engine, including master. That is also the same number Ito reported, so both were this.

It now times an identical BINARY-collated table alongside and asserts on the difference, which is what my manual measurements did and what makes the quantity meaningful:

```
=== master ===
  25 pairs: nocase 158ms, binary 34ms, nocase costs 124ms extra
  400 pairs: nocase 1533ms, binary 122ms, nocase costs 1411ms extra
  FAIL: nocase_plan_overhead_25_to_400_autocommit — 124ms → 1411ms (1137%/400%)

=== this PR ===
  25 pairs: nocase 69ms, binary 33ms, nocase costs 36ms extra
  400 pairs: nocase 159ms, binary 129ms, nocase costs 30ms extra
  PASS: nocase_plan_overhead_25_to_400_autocommit — 36ms (floor 50ms) → 30ms (60%/400%)
```

Worth saying plainly: CI is the first real test of the reformulated gate, since the condition that broke the old one does not exist on my machine. If it is still unhappy there I would rather drop the gate than keep tuning a timing assertion.

## Dead code

`doltliteRecordMaskHasNocaseNul` has no caller outside `prolly_btree.c` — the carry function lives there too — so it never needed exporting. Now static, header declaration removed. `test/dead_code_check.sh` passes.

Re-validated locally: `make lint`, `dead_code_check.sh`, `run_c_tests.sh . coverage`, `doltlite_perf.sh`, and `doltlite_nocase_index.sh` 12/12.

timsehn:
Gate dropped. `test/doltlite_perf.sh` is back to what it had before this PR, plus nothing.

Isolating planning cost from commit cost requires a subtraction between two timings, and that is not an assertion worth maintaining against a runner whose sync cost moves around. What remains guarding this change is the three plan-shape cases in `doltlite_nocase_index.sh` — absent-and-stays-absent across commits, introduced-after-a-clean-probe, and a two-column NOCASE index. Those assert behaviour rather than speed, pass on master as well as here, and are the part that can silently break.

The measurements stay in the PR description as the record of what the change buys.

`doltlite_perf.sh` and `doltlite_nocase_index.sh` (12/12) both pass locally after the removal.

PR #2974, opened 2026-09-16
Add small-table bulk updates to performance hotspots

timsehn:
Bulk updates of non-key columns on text-primary-keyed tables are much slower than stock SQLite even when the entire database fits in cache. Add `bulk_update_text_pk` to a separate **Text PK Bulk Updates** table in the existing CI comment and regression gate.

The workload updates 5,000 balances with one `UPDATE` inside an explicit transaction. Each trial uses a fresh fixture copy, warms it, and times BEGIN, UPDATE, and COMMIT. The cache is 64 MiB; fixture sizes are 546,799 bytes for DoltLite and 577,536 bytes for stock. Stock uses WAL, both engines use synchronous=FULL, and the harness verifies aggregates and integrity before and after reopening the result.

Five-trial local medians from the full six-workload suite:

| Workload | PR base | Candidate | Stock SQLite | Candidate / stock |
|---|---:|---:|---:|---:|
| bulk_update_text_pk | 26.787 ms | 26.864 ms | 2.312 ms | **11.62×** |

Both DoltLite arms use a build whose engine sources match master: this PR changes the benchmark, not the engine. It clears the requested >3× stock threshold. A broader exploratory sweep tried 38 workloads; the three qualifying variants were non-key bulk updates on text/composite primary keys. One representative metric captures that family. Joins, grouping, sorting, triggers, upserts, batched inserts/deletes, and integer-keyed updates did not qualify.

Profiling the 10,000-row text-key variant attributes about 95% of active samples to cursor-position ranking through `prollyMutMapOrderIndexFromEntry`. Its unsorted-map path scans pending edits for each updated row. Isolated confirmation at 1,000 / 5,000 / 10,000 rows took approximately 1.8 / 27.8 / 102.3 ms on master, consistent with quadratic work. The gap also reproduces with WITHOUT ROWID and is present before the index-build optimization in #2973.

Validation:

- 50 hotspot/report/retry harness tests passed.
- Full default benchmark: five trials, all six workloads, all query checks and integrity checks passed; relative performance gate passed.
- Documentation link check and `git diff --check` passed. Updated the existing performance-guide overview to reflect the additional report table.


timsehn:
Superseded by #2975, which removes the quadratic cursor-ranking cost. The optimized workload measures 1.27× stock SQLite in the full local benchmark, below the 3× threshold for a new hotspot report entry.

PR #2981, opened 2026-09-17
Measure reads over a large pending index edit in the hotspot benchmark

timsehn:
Adds a **Large Index Edits** section to `test/performance_hotspots.py`, following the pattern #2976 set.

Half of a table's indexed values move inside one transaction, and the index is then read before the commit. The edits arrive in primary-key order, which says nothing about index-key order, so every read step merges the tree with a large pending map. Profiling put ~29% of such a read in `prollyCursorApplyMergeStep → mergeCompare`, and the cost turns out to be **per step once any map is active, not per pending entry** — 20k and 400k pending edits cost the same. Nothing else benchmarks this shape; sysbench's read-after-write mixes carry pending sets of a few rows.

| 262,144 rows | PR base ms | Candidate ms | Candidate/base | Stock ms | Candidate/stock |
|---|---:|---:|---:|---:|---:|
| index_edit_update | 250.092 | 241.821 | 0.97× | 171.279 | 1.41× |
| index_edit_walk | 21.855 | 21.948 | 1.00× | 9.646 | 2.28× |
| index_edit_range | 12.595 | 12.728 | 1.01× | 4.810 | 2.65× |

Same binary on both arms; walk samples 21.2–22.3 ms, range 12.4–12.9 ms across five trials.

Ratios grow with scale — the ordered walk reached **4.64×** at 1M rows with 500k pending edits (5.16 → 23.95 ms) against 2.14× with none pending — most likely because a 131k-entry pending map stays cache-resident where a 500k one does not. The section uses the harness's `--rows` for consistency with the other fixtures; the point is to have the number tracked.

## Mechanics

- Each trial runs against a fresh copy of the fixture and ends with `ROLLBACK`, so the fixture is never consumed. Stale copies and `.db-lock` sidecars are cleared first.
- The UPDATE is timed alone; its `SELECT changes()` check runs after `.timer off`.
- The two reads are verified against answers **computed from the generating formula in Python**, not taken from either engine — so a wrong result on either side fails the run rather than being recorded as a timing. A unit test checks that formula against real SQL through the `sqlite3` module.
- Both end-to-end runs (32k plumbing, 262k default) passed with real binaries, which is the live confirmation those expectations match.

The OFFSET-pagination shape from the exploration was dropped: at the default size it lands at ~3.5 ms, under the gate's 10 ms minimum regression, so it could never fire. The full ordered walk measures the same mechanism at 22 ms.

`python3 test/performance_hotspots_test.py`: 18 tests. Three new — section rendering and TSV keys (and that `index_scan` stays in the queries section despite the shared prefix), fixture expectations vs real SQL, and `measure_index_edits` statement order, timing isolation, fixture copying and bad-result rejection. The `main` test checks the measurement runs once per arm per trial on a work copy with the fixture's expectations threaded through.


timsehn:
Pushed a second commit: the Large Index Edits section now sizes itself at 1,048,576 rows under a 128 MiB cache. Measured locally (median of 3, stock vs master): update 1.47× → 2.26× → 2.69× and walk 2.31× → 2.42× → 3.05× at 262k → 524k → 1M rows, with the 57 MiB table still resident. The earlier tsan red was an unrelated stock race, fixed in #2988.

PR #2997, opened 2026-09-17
Fail the sqllogictest gate when the stock run does not complete

### LINKED ISSUE #2987: sqllogictest_gate.py ignores stock-runner crashes and timeouts
Found in the 2026-09-17 review at e851e5c70d.

AGENTS.md calls SQLLogicTest a full-pass / 100% parity gate. `test/sqllogictest_gate.py` fails doltlite crashes/timeouts, then runs stock and **discards stock's return code**. Divergences are `doltlite_tokens - stock_tokens`, so a stock timeout that happened to print `!DIVERGE 1` removes doltlite's real miss on line 1.

The gate's own unit test requires that to be a pass:

```
# test/sqllogictest_gate_test.py
("stock-timeout", "", ("!DIVERGE 1\n1 errors out of 2 tests", 1), ("!DIVERGE 1", 124), 0, []),
```

## Repro

```bash
python3 test/sqllogictest_gate_test.py
```

Observed: case `stock-timeout` expects gate rc `0`.

Expected: stock rc `124` (or any non-success) must be a gate failure, the same way doltlite `timeout` is failed in that file.

## Cause

```
# test/sqllogictest_gate.py
output, rc = run(doltlite, path, timeout)
if rc == 124 or b"errors out of" not in output:
    ...  # fail doltlite
    continue
stock_output, _ = run(stock, path, timeout)
divergences = tokens(output) - tokens(stock_output)
```

## Fix

Fail the gate if stock's rc is non-zero / timeout. Do not subtract stock `!DIVERGE` tokens unless stock completed successfully.

Related skip-as-pass (same class, different files): `test/engine_floor_test.sh` prints `SKIP: no stock sqlite3...` then `__SUITE_COMPLETE__` and `exit 0`, so `run_guarded_suite.sh` counts a missing stock binary as a full pass. `doltlite_sqlite_file_oracle.sh` / `doltlite_attach_sqlite.sh` / `doltlite_open_sqlite_file.sh` likewise exit 0 with the sentinel when `sqlite3` is missing.

timsehn:
Fixes #2987.

`sqllogictest_gate.py` subtracts stock's `!DIVERGE` tokens from doltlite's but discarded stock's return code, so a stock timeout or crash that had already printed a divergence line erased doltlite's real miss on that line. The unit test's `stock-timeout` case required exit 0 for exactly that.

**Change**
- A stock run has to complete the same way the doltlite run does: not rc 124 and carrying the `errors out of` summary. Otherwise the file is reported as `CRASH/TIMEOUT: <file> (stock rc=N)`, counted with the crashes, and the gate fails.
- Unit test: `stock-timeout` now expects failure; new `stock-crash` and `stock-missing-summary` cases; every stock stub prints the completion summary the real runner always prints.

**Proof:** the extended test fails against master's gate at `stock-timeout` (gate exit 0, expected 1) and passes 17/17 with this change.

The related skip-as-pass suites named in the issue (`engine_floor_test.sh` and the three sqlite-file suites exiting 0 with `__SUITE_COMPLETE__` when `sqlite3` is missing) are a different mechanism in different files; filed separately.


timsehn:
Ito's crash-after-summary case is real: the completion check accepted any exit code but 124. Fixed in https://github.com/dolthub/doltlite/pull/3000 by requiring the exit code to match the summary's error count, which is the runner's actual protocol.

PR #3046, opened 2026-09-18
Fix SeekLE below all keys with a pending smaller insert

### LINKED ISSUE #3013: SeekLE below every key returns a non-matching row while a smaller uncommitted insert is pending
Found in a full-repo review at `c7c774b535`; found by the randomized SQL differential sweep (seed 1294) and reduced.

Inside a transaction that has inserted a value smaller than every committed key, a `<=` seek below all keys returns the smallest committed row instead of nothing. Stock SQLite returns no rows. The same statement is correct after COMMIT or ROLLBACK, and `SeekLT` and forward scans are fine.

```
                                                          doltlite   stock
BEGIN; INSERT (v=-2) over committed v=-1;
  SELECT v FROM t WHERE v <= -5 ORDER BY v DESC             -1        (none)
  ... WHERE v IS NULL AND k1 < 9e18 ORDER BY k1             row       (none)
  ... WHERE k1='zzz' AND k2>0 ORDER BY k2 DESC              row       (none)
  ... BETWEEN ... ORDER BY DESC, ... LIMIT 1                row       (none)
after COMMIT                                                (none)    (none)
```

Repro:

```sql
CREATE TABLE t(k INTEGER PRIMARY KEY, v);
CREATE INDEX tv ON t(v);
INSERT INTO t VALUES(1,-1);
BEGIN;
INSERT INTO t VALUES(2,-2);
SELECT v FROM t WHERE v <= -5 ORDER BY v DESC;   -- returns -1, expected no rows
COMMIT;
SELECT v FROM t WHERE v <= -5 ORDER BY v DESC;   -- no rows
```

## Cause

Likely `indexMovetoPrefixLast` / `positionBeforeBound` in `src/prolly_btree_cursor_seek.c`: the backward step from a pending (mutmap) landing yields the committed tree's first, higher entry instead of reporting DONE, and the re-armed cached seek key makes `OP_IdxLE` trust the landing without re-comparing.

## Fix

When the merged seek lands before every key and the step back leaves the pending side, the cursor must report end-of-range rather than fall through to the tree's first entry; clear the cached seek key on that path so `IdxLE`/`IdxGE` re-compare. Fail-before/pass-after test: the SQL above run through both engines inside and after the transaction.

timsehn:
Inside a transaction that inserts a value smaller than every committed key, a `<=` seek below the range returned the smallest committed row. Stock SQLite returns no rows.

Previous from the pending first key left the tree cursor on that greater committed entry (`First()` did not set `mergeStepDir`), and `indexMovetoPrefixLast` re-armed the cached seek key so `IdxLE` trusted the landing.

The first commit set `mergeStepDir` after `First()`/`Last()`, re-sought the tree below a MUT landing, and stopped caching the seek key when the prefix range is empty.

A follow-up restores the `mergeStepDir > 0` guard so consecutive backward steps do not pay an extra blob tree seek. `First()` already sets `mergeStepDir=1`, so Previous from the first pending row still re-seeks. The unguarded re-seek was a confirmed 1.333× on `compositepk / ac_writes / oltp_update_index_ac`.

Fail-before: `pending_insert_below_all_seekle` expected `none`, got `-1`. After: reverse-prefix suite 134/134; parity 120/120.

Fixes #3013.


timsehn:
@itoqa thanks for re-running on `8df9d9d`.

That commit restores the `mergeStepDir > 0` guard so consecutive backward MUT steps do not re-seek the tree (the extra blob seek was the composite PK autocommit regression). `First()` still sets `mergeStepDir=1`, so Previous from the first pending row still re-seeks.

The six cases you reran are the ones that matter for that change: reverse-cursor predecessor, SeekLE below every key, and the 134 reverse-prefix checks. Good to see they still match the table scan / stock SQLite.

PR #3051, opened 2026-09-18
Keep read-only connections read-only after a refused reset

### LINKED ISSUE #3017: A refused dolt_reset on a read-only connection makes every later VC writer succeed and persist

On a connection opened with `SQLITE_OPEN_READONLY` (CLI `-readonly`) against a database that has something staged, `dolt_reset('t')` or `dolt_reset()` is correctly refused with `attempt to write a readonly database`. From then on the same connection can run `dolt_branch`, `dolt_tag`, `dolt_tag -d`, `dolt_branch -D`, `dolt_checkout`, `dolt_clean` and `dolt_commit`; they return 0 or a hash and a fresh read-write connection sees the new branch, the deleted branch and tag, and (in the C-API variant) a new HEAD commit. A plain `INSERT` stays refused throughout because the btree layer has its own gate, so only the VC write path is bypassed.

```
readonly connection                      before reset     after refused reset
dolt_branch('x')                         READONLY (8)     0, persisted
dolt_tag / dolt_tag -d / dolt_branch -D  READONLY (8)     0, persisted
dolt_commit(...)                         READONLY (8)     hash, persisted
INSERT                                   READONLY (8)     READONLY (8)
PRAGMA query_only=1 instead              refuses all      refuses all
```

Repro (shell):

```sh
doltlite x.db "CREATE TABLE t(a INT PRIMARY KEY, b TEXT); INSERT INTO t VALUES(1,'x');
  SELECT dolt_commit('-A','-m','c1'); SELECT dolt_tag('v1'); SELECT dolt_branch('feat');
  INSERT INTO t VALUES(2,'y'); SELECT dolt_add('t');"
doltlite -readonly x.db "SELECT dolt_branch('before_reset');   -- refused, correct
  SELECT dolt_reset('t');                                    -- refused, correct
  SELECT dolt_branch('after_reset');                         -- succeeds
  SELECT dolt_tag('ro_tag'); SELECT dolt_tag('-d','v1'); SELECT dolt_branch('-D','feat');
  SELECT dolt_commit('-m','ro commit');"
doltlite x.db "SELECT name FROM dolt_branches; SELECT tag_name FROM dolt_tags; SELECT message FROM dolt_log LIMIT 1;"
-- after_reset, main / ro_tag / (commit if anything was staged)   expected feat, main / v1 / c1
```

## Cause

The refused reset drives the store through `csReloadFromDisk` (`src/chunk_store_lock.c:737`), which reopens the file with `SQLITE_OPEN_READWRITE | SQLITE_OPEN_MAIN_DB` regardless of the connection's mode, and `csAdoptOpenedStoreState` (`src/chunk_wal.c:464`) then does `pDst->readOnly = pSrc->readOnly`, clearing the connection's read-only flag. Only the opposite direction is guarded (`lock.c:757`: "Reload must not replace a writable store with a fallback read-only one"). Every ref writer gates on `cs->readOnly` via `chunkStoreWriteRefused`, so once the flag is gone they all pass.

Separately, `resetStageNamedPaths` (`src/doltlite_reset.c:686-700`) mutates the session's staged catalog in memory before `doltlitePersistWorkingSet` fails, and `reset_cleanup` never restores it; afterwards the session's working tree even reads as clean ("nothing to commit").

## Fix

Open the reload with the connection's own flags (or never adopt `readOnly` from a reload; assert `tmp.readOnly==cs->readOnly`), and check writability up front in every VC writer before any in-memory mutation, routing the reset failure through the same restore-on-failure helper commit uses. Fail-before/pass-after: the shell above; every statement after the reset must fail with rc 8 and the rw connection must see `feat,main / v1 / c1`.

timsehn:
A refused `dolt_reset` on `SQLITE_OPEN_READONLY` was reopening the store `READWRITE` and adopting that `readOnly` bit, so later `dolt_branch` / `dolt_tag` / `dolt_commit` on the same connection persisted.

Reload with the connection's own flags, keep `cs->readOnly` across adopt, reject VC writers before they mutate, and restore reset's session if persist fails.

Fail-before: `after_reset` / `ro_tag` persisted; `feat` and `v1` gone. After: `doltlite_readonly_vc.sh` 5/5; `readonly_reader_writer_test` 33/33.

Fixes #3017.


timsehn:
@itoqa thanks for re-running on `6f01e0f`.

`dolt_checkout` was going through `doltCheckoutParsedFunc`, which never called `doltliteCmdRejectReadOnly` (`dolt_connect_branch` already did). A refused mutate was then formatted as `checkout failed`. The follow-up rejects at the start of `doltCheckoutParsedFunc`, so checkout now returns the same read-only error as the other writers and does not switch branch.

Good to see that case flip ❌→✅ and the rest of the read-only writer set still pass.

PR #3055, opened 2026-09-18
Project workspace staged rows onto the staged schema

### LINKED ISSUE #3019: dolt_workspace row staging into a table whose staged schema is narrower writes the wider working record; the hidden field resurrects later

Stage an untracked table, add a column in the working set, insert a row that uses it, and stage that row through `dolt_workspace_<t>`. The staging succeeds and the commit records schema `(a, b)` whose row 2 carries a third field. On a branch at that commit, `ALTER TABLE t ADD COLUMN c INT DEFAULT 9` yields `2,y,2`: the hidden value comes back instead of the default. `dolt_diff_t` for that commit reports only row 1 as modified.

```
                                             doltlite               Dolt 2.3.1
UPDATE dolt_workspace_t SET staged=1 (to_a=2) succeeds, no residual   succeeds; residual unstaged
                                                                     row 2,false,modified,2,y,2
committed schema                             (a,b) + 3-field record  (a,b), 2-field records
later ADD COLUMN c INT DEFAULT 9             1,x,9 / 2,y,2           1,x,9 / 2,y,9
```

Repro:

```sql
CREATE TABLE u(a INT PRIMARY KEY);
SELECT dolt_commit('-A','-m','c0');
CREATE TABLE t(a INT PRIMARY KEY, b TEXT);
INSERT INTO t VALUES(1,'x');
SELECT dolt_add('t');
ALTER TABLE t ADD COLUMN c INT;
INSERT INTO t VALUES(2,'y',2);
UPDATE dolt_workspace_t SET staged=1 WHERE to_a=2;
SELECT id, staged, diff_type, to_a FROM dolt_workspace_t;
SELECT dolt_commit('-m','c1');
SELECT to_create_statement FROM dolt_schema_diff('HEAD~1','HEAD');   -- CREATE TABLE t(a INT PRIMARY KEY, b TEXT)
SELECT dolt_checkout('-b','x','HEAD');
ALTER TABLE t ADD COLUMN c INT DEFAULT 9;
SELECT a, b, c FROM t ORDER BY a;                                    -- 2,y,2  expected 2,y,9
```

## Cause

`wsApplyRowToStaged` in `src/doltlite_workspace.c` copies `r->pCurVal` (the working record) verbatim into the staged tree. The guard added for #2983, `wsTableSchemaChanged`, compares HEAD vs WORKING only, so a table that is absent from HEAD but present in STAGED with a different schema is never refused.

## Fix

Compare STAGED vs WORKING schema hashes in `wsTableSchemaChanged` and refuse like Dolt ("not modifiable due to schema change"), or project the record onto the staged schema before writing. Fail-before/pass-after: the SQL above.

timsehn:
`dolt_workspace` row staging copied the working record into the staged tree. After `dolt_add` of a new table, `ALTER TABLE … ADD COLUMN`, and `UPDATE dolt_workspace_t SET staged=1`, the commit stored schema `(a, b)` with a 3-field record. `ALTER TABLE t ADD COLUMN c INT DEFAULT 9` then showed `2,y,2` instead of the default.

Dolt 2.3.1 stages the narrower record and leaves an unstaged residual modify. This projects extra working fields off before the staged insert.

Fail-before: residual was `1|added|1` / `1|added|2` (no unstaged modify); after hard reset + ADD COLUMN, `2|y|2`. After: residual includes `0|modified|2`, and `1|x|9` / `2|y|9`. Workspace suite 53/53.

Fixes #3019.


timsehn:
@itoqa thanks for the 11/11 on `64f0d22`.

The HEAD-absent vs already-committed distinction you called out is the one that matters: `wsTableSchemaChanged` still refuses staging when HEAD already has the table and working altered it. This PR only projects when the table is already in STAGED (from `dolt_add`) with a narrower schema than WORKING, which is the path that was writing extra fields under `(a, b)`.

Good to see the composite-key / default-9 case and the residual unstaged modify stay consistent with that.

PR #3058, opened 2026-09-18
Allow BEGIN IMMEDIATE single-file commit with an attached database

### LINKED ISSUE #3023: BEGIN IMMEDIATE or EXCLUSIVE with any second file-backed database attached makes a single-file COMMIT fail and drops the write

With a second file-backed database attached, `BEGIN IMMEDIATE` (or `EXCLUSIVE`) followed by a write to only one of the two files fails at COMMIT with `atomic commit across multiple file-backed databases is not supported`, and the write is gone. Plain `BEGIN` (deferred) commits. Stock SQLite 3.54.0 commits every variant. #2490 covers genuine two-file write transactions; this is a one-file write.

```
                                                doltlite                 stock
attach aux; BEGIN IMMEDIATE; INSERT main; COMMIT   error, row lost        ok
attach aux; BEGIN EXCLUSIVE; INSERT main; COMMIT   error, row lost        ok
attach aux; BEGIN IMMEDIATE; INSERT aux;  COMMIT   error, row lost        ok
attach aux; BEGIN;           INSERT main; COMMIT   ok                     ok
```

`doc/doltlite/bindings.md:93` says the Go driver issues `BEGIN IMMEDIATE` for every transaction, so a Go application that attaches any second file cannot commit anything.

Repro:

```sql
CREATE TABLE t(a);
ATTACH 'aux.db' AS a1;
BEGIN IMMEDIATE;
INSERT INTO t VALUES(1);
COMMIT;                          -- Error: atomic commit across multiple file-backed databases is not supported
SELECT count(*) FROM t;          -- 0
```

## Cause

`vdbeCommit` (`src/vdbeaux.c:2984-3012`) counts every Btree in `TRANS_WRITE` (`nFileWrite++`) and refuses at more than one. `BEGIN IMMEDIATE`/`EXCLUSIVE` opens a write transaction on every attached btree, so the count is 2 even though the aux btree has nothing to write. Stock reaches the same point but a clean pager's `CommitPhaseOne` is a no-op.

## Fix

Count only btrees with pending changes: for prolly, dirty working state, schema or mutmap; for orig-format, dirty pager pages. Fail-before/pass-after: the SQL above through both engines.

timsehn:
`BEGIN IMMEDIATE` or `EXCLUSIVE` with a second file attached opened a write transaction on every btree. `vdbeCommit` counted those as file writes, so an INSERT into only one file failed with `atomic commit across multiple file-backed databases is not supported` and the row was gone. Deferred `BEGIN` committed. Stock SQLite commits every variant. The Go driver issues `BEGIN IMMEDIATE` for every transaction.

Count only btrees with pending changes (prolly: dirty working/schema/mutmap/staging; orig-format: dirty pager pages). Two-file writes still refuse and roll back.

Fail-before: `multifile_immediate_insert_main` got the atomic-commit error. After: sqlite compatibility 16/16; attach write matrix 129/129.

Fixes #3023.


timsehn:
@itoqa thanks for re-running on `b453319`.

`temptrigger-8.3.2` dropping off the unsupported list is expected: `8.1.2` still refuses the two-file INSERT, so `t2`/`aux.t1` stay empty and `8.3.2`'s DELETE is a no-op that matches the empty expected result. Genuine two-file DELETE is still refused (`multifile_temp_trigger_delete_rejected`).

Good to see IMMEDIATE/EXCLUSIVE one-file commits and the two-file reject/rollback cases all still match.

timsehn:
@itoqa thanks for restarting after the rebase on `3c5f752`.

The 12 passing cases match the intended split: IMMEDIATE/EXCLUSIVE one-file commits (including attached orig-format and in-memory) go through, and two-file writes plus TEMP-trigger side effects still refuse and roll back.

PR #3070, opened 2026-09-18
Emit dolt_diff_summary for index-only schema changes

### LINKED ISSUE #3032: dolt_diff_summary emits no row for an index-only schema change

After `CREATE INDEX` (or `DROP INDEX`) with no data change, `dolt_status` says `t modified`, `dolt_diff` says `t,0,1`, `dolt_schema_diff` lists the index, but `dolt_diff_summary` returns nothing for the working set and for the committed range. Dolt returns `t,t,modified,false,true`.

Repro:

```sql
CREATE TABLE t(a INT PRIMARY KEY, b TEXT);
INSERT INTO t VALUES(1,'x');
CREATE INDEX i1 ON t(b);
SELECT table_name, status FROM dolt_status;                                       -- t, modified
SELECT table_name, data_change, schema_change FROM dolt_diff WHERE commit_hash='WORKING';   -- t,0,1
SELECT * FROM dolt_diff_summary('HEAD','WORKING');                                -- (nothing)
SELECT dolt_commit('-A','-m','c2');
SELECT * FROM dolt_diff_summary('HEAD~1','HEAD');                                 -- (nothing)
```

## Cause

`dssAppendTableChange` (`src/doltlite_diff_stat.c:1305-1316`) decides "modified" from the table's root hash and `schemaHash` only; an index is a separate catalog entry and moves neither.

## Fix

Include index entries (and any other per-table catalog rows) in the schema-change decision, the way `dolt_diff` and `dolt_status` already do. Fail-before/pass-after: the SQL above vs Dolt for CREATE and DROP INDEX.

timsehn:
After \`CREATE INDEX\` or \`DROP INDEX\` with no data change, \`dolt_status\` and \`dolt_diff\` report the table as modified with \`schema_change\`. \`dolt_diff_summary\` returned no row because it only compared the table root and \`schemaHash\`. Indexes are separate catalog entries. Dolt returns \`t,t,modified,false,true\`.

Load \`sqlite_master\` rows once per scan and treat index diffs as a schema change on the parent table, matching \`dolt_diff\`.

Fail-before: three summary tests empty vs \`t|t|modified|0|1\`. After: \`doltlite_diff.sh\` 62/62.

Fixes #3032.


timsehn:
@itoqa thanks for the 9/9 on `1c3d753`.

Those are the cases this PR is for: CREATE/DROP/replace index with no row change is `modified, data_change=0, schema_change=1`, an unchanged sibling table stays off the summary, and a data-only change stays `schema_change=0`.

timsehn:
@itoqa thanks for restarting after the rebase on `bf0a3bb`. 14/14 is the same split as before plus the `dolt_schemas` tests that landed on master: index-only CREATE/DROP/replace still `modified|0|1`, sibling tables stay quiet, and a data-only change stays `schema_change=0`.

PR #3073, opened 2026-09-18
Return no such table for dolt_workspace_<missing>

### LINKED ISSUE #3036: SELECT from dolt_workspace_<missing table> fails with "out of memory" instead of "no such table"

```
SELECT * FROM dolt_diff_nope;        -- Parse error: no such table: dolt_diff_nope      (rc 1)
SELECT * FROM dolt_workspace_nope;   -- Parse error: out of memory (7)
```

`dolt_history_<x>` and `dolt_at_<x>` also say `no such table`. A NOMEM result for a name lookup is misleading and trips callers that treat SQLITE_NOMEM as fatal.

Repro:

```sql
CREATE TABLE t(a INT PRIMARY KEY);
SELECT * FROM dolt_workspace_nope;
```

## Cause

`doltliteHistoricalModuleRegister` now also covers `dolt_workspace_` (`src/doltlite_at.c:917`) and its not-a-user-table path surfaces as NOMEM.

## Fix

Return SQLITE_ERROR with `no such table: dolt_workspace_nope` from the module's xConnect when the base table is not a user table, as the sibling modules do. Fail-before/pass-after: the SQL above asserting rc 1 and the message.

timsehn:
\`SELECT * FROM dolt_workspace_nope\` returned \`out of memory (7)\` while \`dolt_diff_nope\` / \`dolt_history_*\` / \`dolt_at_*\` already said \`no such table\`. Callers treat SQLITE_NOMEM as fatal.

\`PRAGMA table_xinfo\` on a missing table has zero columns. \`sqlite3_malloc(0)\` is NULL, so column load returned NOMEM. Treat that as an empty table and map the connect failure to \`no such table: dolt_workspace_nope\`.

Fail-before: \`out of memory (7)\`. After: workspace suite 55/55, including the missing-table message.

Fixes #3036.


timsehn:
@itoqa thanks for the 8/8 on `9492b29`.

The missing-workspace case is the one this PR changes: `dolt_workspace_nope` is `no such table: dolt_workspace_nope`, not `out of memory`. Good to see the sibling `dolt_diff_` / `dolt_history_` / `dolt_at_` names stay in the error, two clients keep their own names, and a later create/insert on the same connection still works.

PR #3076, opened 2026-09-19
Reject --allow-empty together with --skip-empty

### LINKED ISSUE #3037: dolt_commit accepts --allow-empty together with --skip-empty and creates an empty commit

Dolt refuses with `cannot use both --allow-empty and --skip-empty`. DoltLite accepts the pair and creates the empty commit.

Repro:

```sql
CREATE TABLE t(a INT PRIMARY KEY);
SELECT count(*) FROM dolt_log;                                     -- 2
SELECT dolt_commit('--allow-empty','--skip-empty','-m','both');    -- succeeds, expected error
SELECT count(*) FROM dolt_log;                                     -- 3, expected 2
```

## Fix

Reject the combination in the commit option parser with Dolt's message. Fail-before/pass-after: the SQL above vs Dolt.

timsehn:
Dolt refuses \`dolt_commit('--allow-empty','--skip-empty',...)\` with \`cannot use both --allow-empty and --skip-empty\`. DoltLite accepted the pair and created an empty commit.

Reject the combination in the commit option parser with Dolt's message. HEAD does not move.

Fail-before: the dual-flag commit succeeded and \`dolt_log\` grew from 2 to 3. After: commit suite 124/124.

Fixes #3037.


timsehn:
@itoqa thanks for the 6/6 on `e782ea9`.

The flag-conflict error is the whole change: `--allow-empty` plus `--skip-empty` is `cannot use both --allow-empty and --skip-empty`, HEAD does not move, and a positional extra argument still wins as its own error. Good to see that hold across retry, two clients, a staged change, and reopen.

PR #3080, opened 2026-09-19
Record DROP COLUMN index-error order as an adapted SQLite contract

### LINKED ISSUE #3045: ALTER TABLE DROP COLUMN with two indexes on the column names a different index in its error than stock

Both engines refuse the DROP, but the message names a different (also valid) index because DoltLite iterates the table's indexes in a different order. Low severity; it only matters to tests and tooling that match the message.

```
doltlite:  error in index b_g after drop column: no such column: g
stock:     error in index b_rg after drop column: no such column: g
```

Repro:

```sql
CREATE TABLE b(r, g);
CREATE INDEX b_rg ON b(r, g);
CREATE INDEX b_g ON b(g);
ALTER TABLE b DROP COLUMN g;
```

## Fix

Iterate indexes in stock's `pTab->pIndex` order when reporting the first failing index, or accept the divergence and note it in the SQLite compatibility contract.

timsehn:
Both engines refuse \`ALTER TABLE b DROP COLUMN g\` when \`b_rg\` and \`b_g\` still name \`g\`. Stock reports \`error in index b_rg\` because \`sqlite_master\` is scanned in rowid/insertion order. DoltLite reports \`error in index b_g\` because \`sqlite_master\` row order follows the catalog (already \`sql.sqlite_master_canonical\`).

\`pTab->pIndex\` is rebuilt from that catalog on reload, so it is not creation order. Matching stock would mean changing \`sqlite_master\` row assignment. This records the message as adapted and locks the DoltLite spelling.

Fail-before: no contract row. After: compatibility contract 18/18.

Fixes #3045.


timsehn:
@itoqa thanks for the 8/8 on `2b09de1`.

The locked spelling is catalog order: two indexes on the dropped column, error in index `b_g` (not stock's first-created `b_rg`). Good to see that hold after commit/reopen, that a successful drop with no dependent index still works, and that the rejected DROP leaves rows and both indexes in place.

PR #3083, opened 2026-09-19
Seal the SQL transaction after a successful dolt_reset

### LINKED ISSUE #3044: BEGIN; INSERT; dolt_reset('--soft'); ROLLBACK discards the row; Dolt keeps it

Dolt ends the SQL transaction on `dolt_reset` (like `dolt_commit`), so a row inserted before the reset survives a following ROLLBACK. DoltLite rolls it back. `doc/doltlite/transactions.md` currently codifies the DoltLite behaviour ("Undone"), so this is a documented divergence to decide on rather than a plain bug; Dolt is the authority on VC semantics.

```
rows after ROLLBACK: doltlite=3  dolt=4
```

The same matrix shows Dolt also ends the transaction on `dolt_checkout -b` and `dolt_branch`. Plain `dolt_checkout(existing)` inside BEGIN is the already documented divergence (DoltLite keeps the row with the branch; Dolt discards it) and is not part of this issue.

Repro:

```sql
CREATE TABLE t(a INTEGER PRIMARY KEY, b TEXT);
INSERT INTO t VALUES(1,'x'),(2,'y'),(3,'z');
SELECT dolt_commit('-Am','c1');
BEGIN;
INSERT INTO t VALUES(8,'q');
SELECT dolt_reset('--soft');
ROLLBACK;
SELECT count(*) FROM t;     -- 3; Dolt 4
```

## Fix

Either match Dolt (soft reset, `dolt_branch`, `dolt_checkout -b` implicitly commit the open SQL transaction) and update transactions.md, or keep the current behaviour and record the divergence explicitly in the doc and the compatibility contract.

timsehn:
Dolt ends the SQL transaction on \`dolt_reset\` (like \`dolt_commit\`), so \`BEGIN; INSERT; dolt_reset('--soft'); ROLLBACK\` keeps the row. DoltLite only released savepoints, so ROLLBACK dropped it (\`count=3\`, Dolt \`4\`).

On success, \`COMMIT\` then \`BEGIN\` as \`dolt_branch\` already does. A following \`ROLLBACK\` does not undo the reset or prior row writes. Failed resets stay rollbackable.

Fail-before: issue repro returned 3. After: reset suite 90/90, savepoint suite 133/133.

Fixes #3044.


timsehn:
@itoqa thanks for the 6/6 on `ebb2d70`.

The case this PR changes is the one you have under Reset: `BEGIN; INSERT; dolt_reset('--soft'); ROLLBACK` keeps the row. Good to see a failed reset still lets ROLLBACK drop the new row, that hard reset stays at HEAD after ROLLBACK/reopen, and that writes after the sealed txn still commit normally.

